PR #78, opened 2026-04-01
Implement gas refund tracking and streamline Ethereum test execution

ale-linux:
Implement gas refund tracking and streamline Ethereum test execution
This commit implements proper gas refund accounting in SnapshotDB, replacing
the previous stub implementations with full tracking of gas refunds from
operations like SSTORE and SELFDESTRUCT. The refund counter now correctly
integrates with the snapshot and revert mechanism to maintain consistency
during transaction rollbacks. The integration test infrastructure has been
enhanced with a dual-state testing mode that validates state changes against
both Ethereum and Fabric state databases simultaneously for improved debugging.

Additionally, the Ethereum test runner has been simplified to execute only the
hash scheme with trie configuration per subtest, reducing redundant test
permutations while maintaining coverage, and the short mode now randomly skips
tests rather than running a random subset of configurations.


integration/state_test_util.go
```diff
@@ -586,6 +590,80 @@ func makePreState(db ethdb.Database, accounts types.GenesisAlloc, snapshotter bo
...
+			statedb.SetState(addr, k, v)
+		}
+	}
+
+	// Commit and re-open to start with a clean state
+	root, _ := statedb.(*endorser.DualStateDB).EthStateDB().Commit(0, false, false)
+
+	// Commit the fabric state to the database
+	rws := sim.Result()
+	err := fabricDB.UpdateWorldState(context.TODO(), blocks.Block{Number: 0, Transactions: []blocks.Transaction{
```
arner: Loving this!

If I understand correctly we don't need to load the full testharness anymore! We're purely testing the endorser (always have been). Let's fully lean into it and make this the EVMEngine + StateDB integration tests, instead of calling gateways[0].ExecuteTransaction. Create an endorser or EVMEngine, prime it through this function, and use it directly.

The change I'm working on, to test end to end with the `fabrictest` mock backend, is less performant because of a potential race for the priming to be committed on the endorser side after being submitted on the gateway/orderer side (we have to wait for about 10ms, which adds up over a thousand tests). I'll continue with the TrieDB wrapper that operates on committed blocks. Then we can use the StatePrimer again for full ledger tests.

arner (approved):
Alright, it's bringing us a lot of progress as it is, we can see what we want to keep later!

PR #80, opened 2026-04-02
Fix EVM transaction execution error handling and state snapshot manag…

ale-linux:
…ement

Some ethereum tests were failing for certain forks (EIP158, Byzantium, Constantinople, ConstantinopleFix). The root cause was twofold:

First, the executor was not taking a snapshot before executing transactions via ApplyMessage. When transactions failed during execution (like code size exceeded errors), the EVM would revert to snapshot 0 instead of the correct pre-transaction state. This was fixed by adding a snapshot call before ApplyMessage in the executor's execute method, matching the pattern used in go-ethereum's state test runner.

Second, EVM execution errors (such as max code size exceeded) were being propagated as function errors from ProcessProposal, causing the entire proposal to fail. However, these are not fatal errors - they represent valid transactions that executed but failed. The fix wraps such errors in a ProposalResponse instead of returning them as function errors, allowing the transaction to be processed normally with the error encoded in the response. This matches the pattern already used for Call and State operations.

More tests could be enabled after fixing these issues.

### DIFF
--- a/endorser/dual_statedb.go
+++ b/endorser/dual_statedb.go
@@ -209,7 +209,7 @@ func (d *DualStateDB) GetStateAndCommittedState(addr common.Address, hash common
 	d.logger.Debugf("GetStateAndCommittedState: ethCurrent=%s, ethCommitted=%s", ethCurrent.Hex(), ethCommitted.Hex())
 	d.logger.Debugf("GetStateAndCommittedState: snapCurrent=%s, snapCommitted=%s", snapCurrent.Hex(), snapCommitted.Hex())
 	if ethCurrent != snapCurrent || ethCommitted != snapCommitted {
-		d.logger.Debugf("GetStateAndCommittedState: MISMATCH DETECTED!")
+		d.logger.Warn("GetStateAndCommittedState: MISMATCH DETECTED!")
 	}
 	return snapCurrent, snapCommitted
 }


endorser/dual_statedb.go
```diff
@@ -203,9 +203,15 @@ func (d *DualStateDB) GetRefund() uint64 {
...
-	current, committed := d.snapshotDB.GetStateAndCommittedState(addr, hash)
-	d.logger.Debugf("GetStateAndCommittedState: output current=%s, committed=%s", current.Hex(), committed.Hex())
-	return current, committed
+	// Call both to verify they return the same data
+	ethCurrent, ethCommitted := d.ethStateDB.GetStateAndCommittedState(addr, hash)
+	snapCurrent, snapCommitted := d.snapshotDB.GetStateAndCommittedState(addr, hash)
+	d.logger.Debugf("GetStateAndCommittedState: ethCurrent=%s, ethCommitted=%s", ethCurrent.Hex(), ethCommitted.Hex())
+	d.logger.Debugf("GetStateAndCommittedState: snapCurrent=%s, snapCommitted=%s", snapCurrent.Hex(), snapCommitted.Hex())
+	if ethCurrent != snapCurrent || ethCommitted != snapCommitted {
+		d.logger.Debugf("GetStateAndCommittedState: MISMATCH DETECTED!")
```
arner: Maybe this is worth a warn or error?

utils/logger/logger.go
```diff
@@ -0,0 +1,222 @@
...
+func (l *ConfigurableLogger) String() string {
+	l.mu.RLock()
+	defer l.mu.RUnlock()
+	return fmt.Sprintf("ConfigurableLogger{component=%s, level=%s}", l.component, l.level)
+}
+
+// Ensure ConfigurableLogger implements Logger
+var _ Logger = (*ConfigurableLogger)(nil)
+
+// Made with Bob
```
arner: Maybe you can tell Bob not to add that ;)

ale-linux: Anyway the file is gone now :innocent:

PR #81, opened 2026-04-02
Calculate block header StateRoots

arner:
This change adds a Chain struct for the gateway which wraps the existing relational db of the gateway as well as a new pebble TrieStore. When a new block comes in, we replay the writes of the block into the trie store (via the familiar StateDB pattern). We then commit the tries and retrieve the resulting StateRoot hash. The hash is added to the block header information.

A new boolean constant is added to ethereum_tests.go; VERIFY_TRIE_ROOT. Changing it to true allows the developer to test the resulting state root against the ethereum tests. It is currently set to false because of known limitations in our codebase, see COMPATIBILITY.md. For example, several tests fail because we don't revert writes on a partial revert within a transaction.

The backend for local tests is now a fake fabric `fabrictest` from the SDK, so we can test the end to end flow (including submitters and synchronizers). For ethereum tests we keep bypassing the ledger for performance.

We add a sinkLogger for the DualStateDB, allowing the user to pass their own logger. We use it to pass the test logger, which only prints on failed tests. The TestHarness construction is unified. By keeping the StatePrimer on the TestHarness instead of constructing it on the fly, most of the fields on the TestHarness struct could be removed.

### DIFF
--- a/endorser/endorser.go
+++ b/endorser/endorser.go
@@ -50,8 +50,6 @@ func New(engine *EVMEngine, builder endorsement.Builder) (*Endorser, error) {
 }
 
 // SetEthStateDB sets the ethStateDB on the underlying EVMEngine.
-// An optional logSink receives debug log lines from DualStateDB during execution;
-// pass t.Logf in tests to capture state-op traces on failure.
 func (f *Endorser) SetEthStateDB(ethStateDB *ethstate.StateDB) {
 	f.engine.SetEthStateDB(ethStateDB)
 }
--- a/endorser/executor.go
+++ b/endorser/executor.go
@@ -59,8 +59,6 @@ func NewEVMEngine(namespace string, db state.ReadStore, evmConfig *EVMConfig, mo
 }
 
 // SetEthStateDB sets the ethStateDB for the EVMEngine.
-// An optional logSink receives debug log lines from DualStateDB during execution;
-// pass t.Logf in tests to capture state-op traces on failure.
 func (e *EVMEngine) SetEthStateDB(ethStateDB *ethstate.StateDB) {
 	e.ethStateDB = ethStateDB
 }
@@ -227,6 +225,7 @@ func newExecutor(sim *state.SimulationStore, blockInfo *utils.BlockInfo, evmConf
 	if ethStateDB == nil {
 		stateDB = NewSnapshotDB(sim)
 	} else {
+		// NOTE: this is only meant to be used in testing
 		stateDB = NewSnapshotDBWithDualState(sim, ethStateDB)
 	}
 


integration/test_helpers.go
```diff
@@ -111,340 +142,234 @@ func (th *TestHarness) PrimeStateFromJSON(ctx context.Context, jsonFilePath stri
...
+	var ethChainConfig *params.ChainConfig
+	if evmConfig != nil {
+		ethChainConfig = evmConfig.ChainConfig
 	}
 
-	// no certificates for local tests
-	cfg.Endorsers[0].PeerTLS = ""
-	cfg.Endorsers[0].MspDir = ""
+	// Build all endorsers.
+	dbs := make([]*state.VersionedDB, len(cfg.Endorsers))
```
ale-linux: Very nice! So we have a single place where all DBs reside so that we can close them all cleanly! 2 question though: 

1. where do we close them? I couldn't spot it (but maybe because this patch was IMMENSE! 😆 ok, sorry for the dig, but maybe in the future we can break these up into smaller pieces...? 😇 )
2. we also have the `ReadDB` and `WriteDB` abstractions, do we also need to keep track of those to close them, or closing the underlying DB is enough?

endorser/dual_statedb.go
```diff
@@ -37,14 +35,23 @@ type ExtendedStateDB interface {
 type DualStateDB struct {
 	ethStateDB *ethstate.StateDB
 	snapshotDB *SnapshotDB
-	logger     sdk.Logger
```
ale-linux: I vote for keeping my logger pls?

endorser/endorser.go
```diff
@@ -49,9 +49,11 @@ func New(engine *EVMEngine, builder endorsement.Builder) (*Endorser, error) {
...
 }
 
-// SetEthStateDB sets the ethStateDB on the underlying EVMEngine. This allows reusing a primed ethStateDB.
-func (f *Endorser) SetEthStateDB(ethStateDB *ethstate.StateDB) {
-	f.engine.SetEthStateDB(ethStateDB)
+// SetEthStateDB sets the ethStateDB on the underlying EVMEngine.
+// An optional logSink receives debug log lines from DualStateDB during execution;
+// pass t.Logf in tests to capture state-op traces on failure.
+func (f *Endorser) SetEthStateDB(ethStateDB *ethstate.StateDB, logSink ...func(string, ...any)) {
+	f.engine.SetEthStateDB(ethStateDB, logSink...)
```
ale-linux: hmm, should we just keep my logging instead, or do we need this function ptr?

ale-linux (changes requested):
Thanks Arne, this is fantastic work!

gateway/storage/trie/store.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/hyperledger/fabric-x-sdk/blocks"
+)
+
+// Store wraps the geth TrieDB to keep track of a full MPT of the world state.
+// When dbPath is non-empty it is backed by PebbleDB and survives restarts.
+// When dbPath is empty it uses an in-memory database (suitable for tests).
+// Blocks must be committed in order, starting from the block after initialRoot.
+type Store struct {
+	mu         sync.Mutex
+	diskdb     ethdb.Database // underlying disk store; closed on Close()
```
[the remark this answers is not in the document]
arner: We do... the stateRoot is the root of _all_ state, including that was committed earlier. So we can't calculate it on a new block if we don't have the current state.

ale-linux:
Would we want to also update https://github.com/hyperledger/fabric-x-evm/blob/main/docs/COMPATIBILITY.md ?

PR #84, opened 2026-04-03
Use local SimulationStore with native snapshot and revert support

ale-linux:
This commit moves SimulationStore from the external fabric-x-sdk package into the local endorser package, making it the authoritative source for tracking state operations and managing snapshots. The key architectural change is that SimulationStore now natively handles snapshot creation and revert operations using a journal-based mechanism, ensuring that only non-reverted operations appear in the final read-write set. SnapshotDB has been simplified to act as a passthrough for snapshot and revert calls, delegating all snapshot management to the underlying SimulationStore. This design ensures that when a snapshot is reverted, all reads and writes performed after that snapshot are properly removed from the transaction's read-write set, fixing correctness issues with EVM execution that relies on snapshot/revert semantics. The refactoring also eliminates dependencies on external state types throughout the codebase, updates DualStateDB to compare state implementations for debugging, adds comprehensive tests verifying snapshot revert behavior affects the read-write set correctly, and enables 13 previously blacklisted ethereum tests that now pass with the corrected implementation.


arner (approved):
Fantastic!

PR #89, opened 2026-04-08
Run slow and legacy tests only optionally

ale-linux:


arner (approved):
approved modulo one comment

PR #92, opened 2026-04-13
Implement missing RPC methods

kushnireyal:
Needed as part of the HardHat integration task

#### Issue 1: eth_accounts
**Problem**: Method not implemented  
**Solution**: Added test account management system  
**Files**: `gateway/config/config.go`, `gateway/api/eth.go`, `testdata/test-accounts.json`

#### Issue 2: eth_sendTransaction
**Problem**: Method not implemented (Hardhat uses server-side signing)  
**Solution**: Implemented full ECDSA transaction signing with secp256k1  
**Files**: `gateway/api/eth.go`, `gateway/config/config.go`


ale-linux:
Thanks @kushnireyal, looks good. Iiuc `eth_accounts` is meant to be a way to let the node manage wallets on behalf of users. Now this is obviously a much bigger story that has to do with authentication, authorisation, custody etc. Is it correct that for now we don't want to have this as part of the gateway? Because if not, I would propose to just restructure the PR to move all the new code inside the test package, creating a new test gateway object that wraps our gateway and exposes the missing functions (and passes through for all others). Otherwise we are already pre-burdening our "production" gateway with functions that we are not prepared to expose anyway. Wdyt? (and also @arner , your thoughts?)

arner:
yes, both eth_accounts and eth_sendTransaction are for transactions managed by the node. I agree that doing that in production is a big thing. However, it's part of the official API: https://ethereum.github.io/execution-apis/api/methods/eth_sendTransaction so I think we should support it on the Gateway. We also want to use it sometimes when we don't run tests from go (like with the external hardhat tests). Now we could consider adding a flag to run it with these APIs exposed, but IMO that's not necessary. The configuration in Eyal's PR is clear from the naming: an operator wouldn't by accident load private keys in the `testAccounts` in a production instance, so I don't see a risk in exposing the endpoints by default.

ale-linux:
yeah but then - given that it's clearly a feature we use in test, and one that is 100% unusable in production without basically rewriting it, then why not just making it clear by moving that code in a test package?

arner:
Agree for real production, but there's a gradient.

Let's say you're a smart contract developer for a Fabric-X network. You want to make sure that your solidity code works on Fabric-X. In that case you might not want to (only) use a local Hardhat or Anvil node to test your contracts, but an actual Fabric-X test network. Then it's really useful to be able to run the gateway binary with this API to use as a backend for integration tests.

You wouldn't run the gateway or endorser integration tests from this repo. You probably don't even have the source code, because the gateway, endorsers, peers and orderers are running in docker containers pulled from docker. Same images as production, but explicitly configured with a json file with test addresses.

All major nodes have the endpoint (Nethermind, geth, Besu, Erigon), even though it's maybe more historical and for development than something that's really used in production... We could change the TestAddressesPath from a line in the yaml config to a command line flag when you run it, e.g. `--unsafe-test-addresses` or something to make it more explicit?

ale-linux:
Right, I'm not saying we shouldn't do it, I'm just saying we should decouple production code from test code and keep those two separate. Two different files, two different packages - then nothing prevents us from spinning up a main `App` that when run with `--unsafe-support-sign` creates the gateway that supports this. 

I'm sorry to be a pain, but we're still a "young" project and should try to remain free of non-production clutter for as long as we can imho..

arner:
Oh, maybe I misunderstood, I thought you wanted to make the `eth_accounts` and `eth_sendTransaction` code part of a `_test` package that is not compiled into the binary. Putting it in a separate file or package makes sense to keep the distinction clear 👍

ale-linux:
so yes, originally I thought that parking the impl. of those functions into a _test.go file was my proposal. Your argument made sense to me however, that someone might want to test against the binary and not just in UT. So yes, separating the impl. of those new functions in a different package that makes it clear this is code that is not production-ready, but we can instiatiate it both from UT and from app.go (under any necessary flag).

gateway/testimpl/server.go
```diff
@@ -0,0 +1,55 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+
+WARNING: This package contains test-only/unsafe RPC implementations.
+DO NOT use in production environments.
+*/
+
```
arner, line 9: The canonical way would be to have it like this above the package definition: 
```
// package testimpl contains test-only/unsafe RPC implementations.
// DO NOT use in production environments.
```

Then it will be visible in godoc in the editor and on pkg.go.dev.

gateway/testimpl/test_eth_api.go
```diff
@@ -0,0 +1,141 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+
+WARNING: This package contains test-only/unsafe RPC implementations.
+DO NOT use in production environments. These methods perform server-side
+transaction signing which is inherently insecure and should only be used
+for development and testing purposes.
```
arner, line 9: See above, imo we wouldn't need it here then. Just in one file will be enough to surface the comment.

arner (approved):
Well that makes it quite clear ;). Nice addition!

Looks good to me, minor nitpick about the package comments.

ale-linux (commented):
Fantastic, thanks for the refactor!

ale-linux (approved):

PR #96, opened 2026-04-15
Remove signed proposals

ale-linux:
Given that for now endorser and gateway are colocated within the same process, there is no reason to sign the proposal. If and when we decouple them into separate processes, we can consider whether signing the proposal is the best way to authenticate the request, or whether other types of authentication are preferrable.


gateway/core/endorse.go
```diff
@@ -105,25 +108,66 @@ func (e *EndorsementClient) CallContract(ctx context.Context, args ethereum.Call
...
-	res, err := e.endorsers[0].ProcessProposal(ctx, prop, nil)
+	hdr, err := protoutil.UnmarshalHeader(proposal.Header)
 	if err != nil {
-		return nil, fmt.Errorf("process proposal: %w", err)
+		return endorsement.Invocation{}, fmt.Errorf("failed to deserialise header: %w", err)
 	}
-	if res.Response.Status < 200 || res.Response.Status >= 400 {
-		return nil, fmt.Errorf("query response was not successful, error code %d, msg %s", res.Response.Status, res.Response.Message)
+
+	// Compute proposal hash (hash of empty byte array as we don't have a real proposal)
```
arner: Didn't we fill the proposal.Payload on line 128?

arner (approved):
That simplifies it! The way we had the ProcessProposal API, like the fabric peer but not quite, was a bit awkward. And the same for the signatures we don't validate. 

Maybe we can improve it further. The code in the`api` package is not the GRPC API it pretended to be. When we add an actual API, it can live in this package, but right now I'm wondering if the separation between `api.EndorserAPI` and `endorser.Endorser` still serve their purpose. They both mainly forward calls, without actually translating DTOs, and the layered architecture only really makes sense if the `api` layer is an actual layer...

And as we discussed offline, let's remove the `deserializer.go`; we can bring it back if we want to start verifying signatures, but right now it's unused code.

ale-linux:
Thanks Arne, I think I addressed all your comments!

endorser/endorser.go
```diff
@@ -59,41 +69,145 @@ func (f *Endorser) GetEthStateDB() *ethstate.StateDB {
 	return f.engine.GetEthStateDB()
 }
 
-// ExecuteTransaction processes a transaction and returns a signed proposal response.
-func (f *Endorser) ExecuteTransaction(ctx context.Context, inv endorsement.Invocation, ethTx *types.Transaction, blockInfo *utils.BlockInfo) (*peer.ProposalResponse, error) {
+// ProcessEVMTransaction processes an Ethereum transaction and returns a signed proposal response
+func (f *Endorser) ProcessEVMTransaction(ctx context.Context, inv endorsement.Invocation, ethTx *types.Transaction, blockInfo *utils.BlockInfo) (*peer.ProposalResponse, error) {
+	// Validate the ethereum transaction signature
+	ethChainConfig := f.ethChainConfig
+	if ethChainConfig == nil {
```
arner: we could also do this in the constructor

arner (approved):
Nice!

endorser/endorser.go
ale-linux: yep good call - done

PR #97, opened 2026-04-15
Mempool

ale-linux:


integration/test_helpers.go
```diff
@@ -566,6 +652,21 @@ func callSmartContract(t *testing.T, client *EthClient, addr ethcommon.Address,
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	for pending := true; pending; {
```
arner: If we have this, doesn't that mean we can get rid of the `submitWaitTime` config? It's a hack that we wanted to remove anyway, a heuristic for making sure that the transaction was "probably" committed by then. But now that we return immediately, there's no reason for the worker to wait after submitting, is there?

integration/test_helpers.go
```diff
@@ -307,9 +325,46 @@ func buildTestHarness(t *testing.T, logger sdk.Logger, cfg config.Config, evmCon
 	return th, syncs, nil
 }
 
+// applyConfigOverrides applies overrides from a map to a config struct using reflection.
+// Keys use dot notation like "Gateway.WorkerCount" to specify nested fields.
+func applyConfigOverrides(cfg *config.Config, overrides map[string]any) error {
```
arner, line 298: I couldn't find where this was used (we seem to pass `nil` everywhere?). What's it for?

ale-linux: good point - I use this stuff in a follow-up commit where I vary from testing the number of workers programmatically but yeah, it's not used in this commit yet. Okay to keep it or do you want it moved to that commit?

integration/test_helpers.go
ale-linux: Done! Great call!

integration/test_helpers.go
```diff
@@ -693,9 +678,101 @@ func querySmartContractExpect(t *testing.T, client *EthClient, addr ethcommon.Ad
...
+}
+
+func waitForCommitT(t *testing.T, ec *ethclient.Client, tx *types.Transaction) {
+	err := waitForCommit(t.Context(), ec, tx)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func waitForCommit(ctx context.Context, ec *ethclient.Client, tx *types.Transaction) error {
```
arner, line 742: maybe this this is a bit more flexible in use if we don't require the full transaction as parameter but just the hash? (not sure if we need it but could be?)

gateway/core/api.go
```diff
@@ -57,62 +62,104 @@ type Store interface {
...
+	for {
+		tx, ok := g.txQueue.Dequeue()
+		if !ok {
+			// Queue is closed and empty
+			return
+		}
+
+		// Process the transaction (old SendTransaction logic)
+		if err := g.processTx(ctx, tx); err != nil {
+			// TODO: Add proper error handling/logging
```
arner, line 101: Oh, that's a big one. We have different error types: pre-execution, execution, internal server errors, Fabric errors.

Our current behaviour is: wait until it's "probably" committed, then return success or error (regardless of the error type). We currently don't submit failed or reverted transactions to the ledger, but we do return every error message.

In our endorser docs:

```go
// isPreExecutionError checks if an error is a pre-execution validation error
// that should reject the transaction, as opposed to an execution error that
// should be included in the transaction result.
//
// According to go-ethereum's ApplyMessage documentation:
// "An error always indicates a core error meaning that the message would always
// fail for that particular state and would never be accepted within a block."
//
// Pre-execution errors include:
// - Nonce errors (too low, too high)
// - Insufficient funds
// - Gas limit errors
// - Init code size exceeded (EIP-3860)
//
// Execution errors (NOT pre-execution) include:
// - Out of gas during execution
// - Execution reverted
// - Invalid opcode
```

Ethereum clients expect the following error modes (AFAIK):
- Not accepted: pre-execution and internal server errors return immediately to the caller (nonce, insufficient funds, etc).
- Accepted and mined with FAILED status (revert, out of gas, etc) -> user can query by txHash, from any node.
- Accepted but never mined (congestion, gas price too low, nonce gap, replaced tx) -> looks like PENDING for a while if you query by hash on this specific node, and eventually disappears.

With this PR, all errors are swallowed and lead to timeouts for the client who will be polling for the tx hash, and no way to find out what the issue was.... It will likely cause issues with external tooling and the upcoming Hardhat integration tests.

Of course we don't have to solve the full error handling now, but do you think we can do something to at least surface them to the client? On polling GetTransactionByHash might be enough for now, if we keep errors in memory with a TTL?

integration/test_helpers.go
arner: Sure let's keep it :)

gateway/core/api.go
ale-linux: Good point, we need to behave like geth from this perspective, which requires understanding what checks are done.. I think this is a big one so I'd defer it to a follow-up PR? Wdyt?

integration/test_helpers.go
ale-linux: we could, but anyway we need to construct a tx object to get its hash, and this is just test code so I wouldn't mind leaving it as is..

gateway/core/api.go
arner: Ok, as discussed offline I can live with it if it's just for a short while :)

arner (approved):
LGTM

PR #98, opened 2026-04-16
add option to zero out gas fees

arner:
Before this, the following command:

```
cast send --rpc-url http://localhost:8545 --chain-id 31337 --private-key $PRIVATE_KEY \
  --create "$(cat bin/GLDToken/GLDToken.bin)$(cast abi-encode 'constructor(uint256)' 1000000000000000000000  | sed 's/^0x//')"
```

Failed with:

```
Error: server returned an error response: error code -32000: process proposal: insufficient funds for gas * price + value: address 0xaFE0ef8c1c986b2dE97da30c1F4DA0E09733Cf5C have 0 want 5000000
```

### DIFF
--- a/endorser/app/factory.go
+++ b/endorser/app/factory.go
@@ -48,6 +48,7 @@ func NewEndorser(
 
 	evmConfig := &endorser.EVMConfig{
 		ChainConfig: common.BuildChainConfig(network.ChainID),
+		FreeGas:     true,
 	}
 
 	var builder endorsement.Builder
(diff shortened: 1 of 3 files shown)


ale-linux:
LGTM - maybe can you see how hard is it to tell ethereum clients to set gas price to zero? If it's easy we might not need this one, if it's complicated let's go with this patch

arner:
I did the research, and it runs out that both cast and metamask send:

```
  "gas": "0x0",
  "maxFeePerGas": "0x1",
  "maxPriorityFeePerGas": "0x1",
```

With cast you can do `--gas-price 0 --priority-gas-price 0`, but with metamask it's not so easy. So I think this workaround is justified!

ale-linux (approved):

PR #101, opened 2026-04-18
add blockhash to logs, return block 0, include from in tx

arner:
This PR includes several small changes that improve the API compatibility. Tested with Blockscout.


ale-linux (approved):
LGTM - thanks @arner. Let's just wait for the tests

PR #109, opened 2026-04-22
feat: add Fablo integration test network

### LINKED ISSUE #100: Use Fablo instead of Fabric Samples network as classic Fabric backend
Fablo starts faster (14 seconds instead of a minute), does not need the "init" step, is more configurable, and uses BFT by default so the transactions are committed much faster (we don't have to wait for the default 2 second block commit time in the samples network).

A lot of the implementation can be copied from https://github.com/hyperledger/fabric-x-sdk/pull/11 .

Daniel-1600:
closes #100

@arner I did this with the help of AI I have indicated on the commit


arner (changes requested):
Hi, I tried to run the integration tests script but they don't work:

```
➜ ./scripts/run_integration_test.sh 
Starting Fabric network with Fablo...
✓ Fablo binary and config verified
Generating network artifacts with Fablo...

┌──────      .─.       ┌─────.    ╷           .────.
│           /   \      │      │   │         ╱        ╲ 
├─────     /     \     ├─────:    │        │          │
│         /───────\    │      │   │         ╲        ╱ 
╵        /         \   └─────'    └──────     '────'                      v2.6.0
┌┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┐
│ https://fablo.io | created at SoftwareMill | backed by Hyperledger Foundation│
└┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┘
Generating network config
    FABLO_VERSION:      2.6.0
    FABLO_CONFIG:       /Users/arne/dev/fabric/fabric-x-evm/fablo-config.json
    FABLO_NETWORK_ROOT: /Users/arne/dev/fabric/fabric-x-evm/fablo-target
File /Users/arne/dev/fabric/fabric-x-evm/fablo-config.json does not exist
Failed to generate Fablo artifacts
```

Can you also change the targets in the makefile? I propose to use the same commands as in the SDK:

```
.PHONY: start-fablo
start-fablo:
	cd testdata/fablo && ./fablo up

.PHONY: stop-fablo
stop-fablo:
	cd testdata/fablo && ./fablo prune

.PHONY: test-fablo
test-fablo:
	@go test -timeout 60s -run ^TestFablo$$ ./integration

.PHONY: clean-fablo
clean-fablo:
	cd testdata/fablo && ./fablo prune || true
	rm -rf testdata/fablo/snapshot.fablo.tar.gz
```

And to remove the `init-3 start-3` etc for fabric samples. Finally it'd be great if you can then also update the README.md with these commands. Thanks!

Daniel-1600:
The checks are now passing and also updated the readme too!
image

Makefile
```diff
@@ -53,29 +53,23 @@ test-x:
...
-	@go test -timeout 360s -v -run ^TestFabric$$ ./integration
-
-.PHONY: stop-3
-stop-3:
-	@./testdata/fabric-samples/test-network/network.sh down
-	@docker rm -f $$(docker ps -a -q --filter "ancestor=basic_ccaas_image") || true
-
+.PHONY: start-fablo
+start-fablo:
+	  echo "127.0.0.1 peer0.org1.example.com peer0.org2.example.com orderer0.group1.orderer.example.com orderer1.group1.orderer.example.com" | sudo tee -a /etc/hosts
```
arner: Although it might make things easier, we shouldn't be touching people's /etc/hosts. The other suggestion by copilot, to change the URLs in the FabloConfig to localhost (or explicitly 127.0.0.1) is better for now. We'll soon add configuration files instead of hardcoded config, that will resolve some of the friction here.

arner (commented):
Great that you got it working! One more thing about the URLs. Looking at the Fabric-X-SDK, most likely the following will fix it:

```go
		Orderers: []network.OrdererConf{{
			Address: "127.0.0.1:7030",
			TLS: network.TLSConfig{
				Mode:        network.TLSModeTLS,
				ServerName:  "orderer0.group1.orderer.example.com",
				CACertPaths: []string{path.Join(orderer0, "tls", "ca.crt")},
			},
		}},
		Peer: network.PeerConf{
			Address: "127.0.0.1:7041",
			TLS: network.TLSConfig{
				Mode:        network.TLSModeTLS,
				ServerName:  "peer0.org1.example.com",
				CACertPaths: []string{path.Join(peer0, "tls", "ca.crt")},
			},
		},
```

This makes it connect to localhost, while still accepting TLS certificates from another hostname.

Daniel-1600:
@arner The tether_token_replay test is occasionally flaky when run back-to-back. It passes consistently in isolation (make test-fablo) but sometimes hangs or fails with nonce too high in consecutive runs via run_integration_test.sh. This appears to be a pre-existing network timing issue unrelated to the Fablo config changes in this PR.

Here is gist https://gist.github.com/Daniel-1600/b75d8fd575292ddcf289d53aa61b117c

arner (approved):
Very nice, thanks!

arner:
I found the flakiness issue and will make another PR to resolve that. Our tests wait until the gateway has processed the block, and then immediately send the query. But the endorsers process the blocks a tiny bit slower, so they haven't caught up yet and give a stale response. We didn't see this issue with the fabric samples network, because one transaction takes 2 seconds for it and the backoff in our test by then is long enough that it the endorsers are always caught up.

PR #113, opened 2026-04-26
integration: replace time.Sleep with WaitGroup in tether_token_parallel

### LINKED ISSUE #77: Change integration tests to submit and then poll
Our integration tests currently submit a transaction and then wait till it commits. We should instead better align with ethereum to asynchronously submit the tx and then start polling to see its status

Storm1289:
Closes #77

## What's the problem?
`testTetherTokenParallel` used a `time.Sleep(200ms)` at the end to wait
for two fire-and-forget `go submit(...)` goroutines to finish. This is a
wall-clock guess  it can flake on slow machines or pass on fast ones
even if something is wrong.

## What's the fix?
Replaced the `go submit / submit / time.Sleep` block with a
`sync.WaitGroup`. Both submits now run as goroutines and the test blocks
on `wg.Wait()` until both are done. Since `submit` already polls
`TransactionByHash` via `waitForCommit`, the test ends deterministically
when both transactions are confirmed committed.

This removes the last remaining wall-clock wait in the integration tests,
finishing the cleanup started in PR #97 (which removed `SubmitWaitTime`).

## What changed?
- Dropped `time.Sleep(200 * time.Millisecond)` from `testTetherTokenParallel`
- Wrapped both `submit` calls in a `sync.WaitGroup`
- Imports: removed `"time"`, added `"sync"`
- No production code touched

## Test plan
- [x] `make checks`
- [x] `make unit-tests`
- [x] `make test-local`  all 7 subtests pass, including `tether_token_parallel`

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -11,8 +11,8 @@ import (
 	"io"
 	"math/big"
 	"os"
+	"sync"
 	"testing"
-	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/tests"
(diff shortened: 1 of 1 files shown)


ale-linux (approved):
Thanks for your contribution @Storm1289 !

Issue #62, opened 2026-03-26
Support Transaction Replacement


ale-linux:
Implement support for Ethereum-style transaction replacement where a client can submit a new transaction with the same nonce as a pending transaction but with higher gas price to replace it. Detect replacement transactions by comparing nonces of incoming transactions with queued transactions from the same sender. Implement the replacement logic that removes the old transaction from the mempool and queues the new one, ensuring atomic replacement to avoid race conditions. While fabric-x-evm does not use gas prices for transaction ordering, support the replacement mechanism for Ethereum client compatibility, potentially using other criteria like submission timestamp. Document the replacement policy clearly and ensure clients receive appropriate feedback when replacements succeed or fail. Test replacement scenarios thoroughly to ensure correct behavior.

---

**Quarter:** Q3

PR #115, opened 2026-04-27
gateway: add geth-style pre-flight tx validation to SendTransaction

### LINKED ISSUE #99: Preserve geth failure model for `eth_sendRawTransaction`
As discussed [here](https://github.com/hyperledger/fabric-x-evm/pull/97#discussion_r3099768186), we need to deliver the same type of errors as geth does when `eth_sendRawTransaction` is called. Currently we never err out there

Storm1289:
## Summary

Add geth-style pre-flight transaction validation to the gateway `SendTransaction` path so `eth_sendRawTransaction` returns the same class of errors an Ethereum client expects before the transaction enters the async endorse/submit pipeline.

---

### New: `gateway/core/validate.go`

Pre-flight validation now calls geth’s exported helpers directly instead of re-implementing checks, ensuring behavior stays aligned with upstream:

- `core/txpool.ValidateTransaction` for stateless checks:
  - signer recovery
  - intrinsic gas
  - init-code size
  - fee cap / tip cap bounds
  - negative value
  - nonce overflow
  - tx-type validation

- `core/txpool.ValidateTransactionWithState` for stateful checks:
  - nonce-too-low
  - insufficient funds  
  Uses an ephemeral `*state.StateDB` seeded with sender nonce + balance from the endorser

- Replay protection (`!tx.Protected()`) is handled explicitly at the RPC layer (matches geth behavior)

Validation configuration:

- Accepted tx types: `LegacyTxType | AccessListTxType | DynamicFeeTxType`
- Blob (EIP-4844) and set-code (EIP-7702) transactions are rejected
- `MaxBlobCount = 0`
- `MaxSize = 128 KiB`
- `MinTip = 0`

Execution context:

- Synthetic block `head`:
  - `GasLimit = 30_000_000`
  - `Difficulty = 0` (post-merge)
- Fork rules unchanged (all forks active at genesis)

Replacement transaction logic is intentionally stubbed:

- `ExistingCost → nil`
- `ExistingExpenditure → 0`
- `FirstNonceGap → nil`

Tracked separately in: #62

All errors are preserved as geth sentinel errors and wrapped with `%w` so `errors.Is` works correctly on RPC clients.

---

### Modified: `gateway/core/api.go`

- Added `chainConfig` and `signer` to `Gateway` (derived from `chainID`)
- `SendTransaction` now calls validation before enqueueing
- Existing `NonceAt` and `BalanceAt` satisfy required state access (no extra plumbing)

---

### New: `gateway/core/validate_test.go`

Added unit tests covering:

- Unprotected transaction  
- Chain ID mismatch  
- Tip > fee cap  
- Intrinsic gas failure  
- Nonce too low  
- Insufficient funds  
- Init code too large  
- Negative value  
- State lookup error propagation  
- Valid transaction (happy path)

---

### Modified: `integration/ethclient.go`

Validator surfaced two issues in test helpers:

- Fixed `GetCtxForSigner()`:
  - Previously returned `(nil, 0)` → caused fallback to `FrontierSigner`
  - Now returns `(big.NewInt(0), 0)` to ensure EIP-155 signing

- Fixed gas handling:
  - Legacy transactions previously used `Gas: 0` with fallback
  - Added `defaultGas = 5_000_000` explicitly in:
    - `txForDeploy`
    - `txForCall`

---

### Docs

Updated `docs/COMPATIBILITY.md`:

- Added section: **Pre-flight transaction validation**
- Updated:
  - Nonce management
  - Gas and fees  
  (now enforced synchronously instead of deferred)

---

## Test Plan

- [x] `make checks`  
- [x] `make unit-tests`  
- [x] `make test-local-x` (all subtests passing)  

---

Closes #99

### DIFF
--- a/docs/COMPATIBILITY.md
+++ b/docs/COMPATIBILITY.md
@@ -126,24 +126,25 @@ Here, because of execute-order-commit:
 
 ## Pre-flight transaction validation
 
-`eth_sendRawTransaction` runs the same stateless and stateful checks geth performs before a
-transaction enters its mempool, so callers see geth-style errors immediately rather than after
-endorsement. The gateway calls geth's exported `core/txpool.ValidateTransaction` and
-`core/txpool.ValidateTransactionWithState` directly (see `gateway/core/validate.go`) — when
-go-ethereum is upgraded, the validation set tracks upstream automatically.
+`eth_sendRawTransaction` runs geth's stateless validation before a transaction is enqueued for
+endorsement, so callers see geth-style errors immediately. The gateway calls
+`core/txpool.ValidateTransaction` directly (see `gateway/core/validate.go`); upgrading
+go-ethereum picks up upstream changes automatically.
 
-The stateful check builds an ephemeral `*state.StateDB` seeded with only the sender's nonce and
-balance fetched from the endorser; we don't share the endorser's persistent state DB.
+The only stateful check is **nonce-too-low**: we look up the sender's committed nonce from the
+endorser and reject the transaction if it is lower than expected. We do not call
+`txpool.ValidateTransactionWithState` — building a per-tx `*state.StateDB` for one nonce/balance
+lookup is too expensive — so the balance/cost check is skipped (gas is not metered anyway).
 
 **Deliberate deviations from geth's failure model:**
 
-- **Replacement transactions are not supported**: `ExistingCost` always returns `nil` and
-  `ExistingExpenditure` always returns `0`. This means the validator never bumps fee on a
-  replacement tx and never enforces nonce-account slot limits. The gateway has no mempool, so
-  there is nothing to replace; tracking this gap is open work.
-- **Nonce gaps are accepted**: `FirstNonceGap` is left nil, so a transaction with a nonce
-  arbitrarily higher than the account's current nonce is admitted. The endorser is responsible
-  for the eventual ordering check.
+- **Sender balance is not validated**: `txpool.ValidateTransactionWithState` would reject a tx
+  whose `gas × gasPrice + value` exceeds the sender balance. We skip this since gas is not
+  metered and balances are unfunded by default (see Gas and fees).
+- **Replacement transactions are not supported**: tracked in #62. The gateway has no mempool
+  today, so there is nothing to replace.
+- **Nonce gaps are accepted**: a transaction with a nonce higher than the account's current
+  nonce is admitted; the endorser enforces ordering at execution time.
 - **Blob transactions (EIP-4844, type 3) are rejected at submission**: the `Accept` bitmap
   excludes them and `MaxBlobCount` is `0`. They were previously accepted and executed without
   KZG proof validation; that path is no longer reachable through the JSON-RPC gateway.
@@ -261,15 +262,15 @@ read `block.number` or `block.timestamp` inside a view function will see inconsi
 
 Gas mechanics are intentionally not implemented.
 
-| Aspect                      | Fabric.                                      | Ethereum                            |
-| --------------------------- | -------------------------------------------- | ----------------------------------- |
-| `GASPRICE` opcode           | `0`                                          | actual tx gas price                 |
-| Sender balance check        | enforced at submission (`gas×price + value`) | must cover `gas × gasPrice + value` |
-| Intrinsic gas deduction     | enforced at submission, not deducted         | ~21 000 deducted before execution   |
-| Gas refund counter          | always `0`                                   | tracks SSTORE/SELFDESTRUCT refunds  |
-| Default gas per call/deploy | `5 000 000` if not specified                 | whatever the tx sets                |
-| Block gas limit             | `10 000 000`                                 | network-set limit                   |
-| Access list warmup          | not done                                     | sender, to, precompiles pre-warmed  |
+| Aspect                      | Fabric.                              | Ethereum                            |
+| --------------------------- | ------------------------------------ | ----------------------------------- |
+| `GASPRICE` opcode           | `0`                                  | actual tx gas price                 |
+| Sender balance check        | not performed                        | must cover `gas × gasPrice + value` |
+| Intrinsic gas deduction     | enforced at submission, not deducted | ~21 000 deducted before execution   |
+| Gas refund counter          | always `0`                           | tracks SSTORE/SELFDESTRUCT refunds  |
+| Default gas per call/deploy | `5 000 000` if not specified         | whatever the tx sets                |
+| Block gas limit             | `10 000 000`                         | network-set limit                   |
+| Access list warmup          | not done                             | sender, to, precompiles pre-warmed  |
 
 **JSON-RPC fee stubs**: `eth_estimateGas` always returns `0`; `eth_gasPrice` and
 `eth_maxPriorityFeePerGas` always return `1 wei`; `eth_feeHistory` returns all-zero arrays. Clients
--- a/docs/COMPATIBILITY.md
+++ b/docs/COMPATIBILITY.md
@@ -124,10 +124,50 @@ Here, because of execute-order-commit:
 
 ---
 
+## Pre-flight transaction validation
+
+`eth_sendRawTransaction` runs the same stateless and stateful checks geth performs before a
+transaction enters its mempool, so callers see geth-style errors immediately rather than after
+endorsement. The gateway calls geth's exported `core/txpool.ValidateTransaction` and
+`core/txpool.ValidateTransactionWithState` directly (see `gateway/core/validate.go`) — when
+go-ethereum is upgraded, the validation set tracks upstream automatically.
+
+The stateful check builds an ephemeral `*state.StateDB` seeded with only the sender's nonce and
+balance fetched from the endorser; we don't share the endorser's persistent state DB.
+
+**Deliberate deviations from geth's failure model:**
+
+- **Replacement transactions are not supported**: `ExistingCost` always returns `nil` and
+  `ExistingExpenditure` always returns `0`. This means the validator never bumps fee on a
+  replacement tx and never enforces nonce-account slot limits. The gateway has no mempool, so
+  there is nothing to replace; tracking this gap is open work.
+- **Nonce gaps are accepted**: `FirstNonceGap` is left nil, so a transaction with a nonce
+  arbitrarily higher than the account's current nonce is admitted. The endorser is responsible
+  for the eventual ordering check.
+- **Blob transactions (EIP-4844, type 3) are rejected at submission**: the `Accept` bitmap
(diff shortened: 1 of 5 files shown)


ale-linux (changes requested):
Thanks for your contribution @Storm1289.

Here's a list of checks I could find by manual inspection

- check tx fee: https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L1610
- protected check: https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L1613
- supported tx check: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L64-L66
- blob count: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L67-L69
- max size check: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L72-L74
- enabled tx types check: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L75-L88
- init code size: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L90-L94
- gas too high: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L95-L97
- negative value: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L98-L102
- gas block limit: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L103-L106
- large numnbers: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L107-L113
- fee cap/tip cap check: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L114-L117
- verify signer: 
    - https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L1620-L1626
    - https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L118-L121
- nonce wraparound: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L122-L125
- intrinsic gas: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L125-L134
- floor data gas: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L135-L144
- gas price/min tip for pool check: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L145-L148
- authorisation: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L152-L156
- nonce check: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L261-L264
- nonce gaps: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L265-L271
- balance/cost: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L272-L279
- replacement/expansion: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L280-L302
- delegation: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/blobpool/blobpool.go#L1321
- replacement txes: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/blobpool/blobpool.go#L1326-L1361

Something that stood out after the manual inspection is that [ValidateTransaction](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L62) is exported and so in theory we could just call it. We need to find the proper way of constructing the `opts` argument. Same for [ValidateTransactionWithState](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L254).

We'll need to make some assumptions for now (and document them in https://github.com/hyperledger/fabric-x-evm/blob/main/docs/COMPATIBILITY.md). For example for now 

- we could ignore blob transactions
- we have to ignore replacement since we don't support that yet (we'd need to create a github issue to track that)

For the rest, I would really try to directly call those geth functions so that we're as compatible as possible.

Storm1289:
Hi @ale-linux , thanks for the review,
I've pushed 3292325 which refactors the validator to call geth's exported helpers directly:

- Stateless checks now go through `core/txpool.ValidateTransaction` with `Accept = Legacy | AccessList | DynamicFee`, `MaxBlobCount = 0`, `MaxSize = 128 KiB`, `MinTip = 0`, and a synthetic head (`GasLimit = 30M`, `Difficulty = 0` for post-merge). Fork rules are unchanged since all our forks activate at genesis.
- Stateful checks go through `core/txpool.ValidateTransactionWithState`, passed an ephemeral `*state.StateDB` seeded only with the sender's nonce and balance fetched from the endorser. This avoids coupling the gateway to the endorser's persistent store.
- Replay protection (`!tx.Protected()`) stays as an explicit check above the txpool call — geth does it at the RPC layer in `internal/ethapi.SubmitTransaction`, not inside the txpool.

Per your guidance I've stubbed out what we don't support yet:
- **Blob (EIP-4844) and set-code (EIP-7702) txs** — rejected at submission via the `Accept` bitmap.
- **Replacement transactions** — `ExistingCost → nil`, `ExistingExpenditure → 0`. This is already tracked by #62, so I've referenced it in COMPATIBILITY.md rather than opening a new issue.
- **Nonce gaps** — `FirstNonceGap` left nil; the endorser still enforces ordering at execution time.
- **RPC tx-fee cap** — we don't expose an operator knob for this, so submission is accepted regardless of total fee (subject to geth's 256-bit sanity).

`docs/COMPATIBILITY.md` has a new "Pre-flight transaction validation" section enumerating each deviation, and the existing "Nonce management" / "Gas and fees" entries are updated since nonce-too-low, sender-balance, and intrinsic-gas are now enforced synchronously.

Let me know if anything else needs adjusting and also updated the pr too.

ale-linux:
Thanks for your updated patch. I must apologise, I thought we could just provide a `StateDB` interface and not the specific `*state.StateDB` implementation. Instantiating a real db per transaction is a bit too expensive imho... I tried this patch

```
diff --git a/gateway/core/validate.go b/gateway/core/validate.go
index 1540484..4683c2d 100644
--- a/gateway/core/validate.go
+++ b/gateway/core/validate.go
@@ -13,9 +13,7 @@ import (
        "math/big"
 
        "github.com/ethereum/go-ethereum/common"
-       "github.com/ethereum/go-ethereum/core/rawdb"
        "github.com/ethereum/go-ethereum/core/state"
-       "github.com/ethereum/go-ethereum/core/tracing"
        "github.com/ethereum/go-ethereum/core/txpool"
        "github.com/ethereum/go-ethereum/core/types"
        "github.com/ethereum/go-ethereum/params"
@@ -96,15 +94,43 @@ func validateTx(
        return txpool.ValidateTransactionWithState(tx, signer, stateOpts)
 }
 
-// newEphemeralStateDB lets us hand a real *state.StateDB to
-// txpool.ValidateTransactionWithState without sharing the endorser's store.
-func newEphemeralStateDB(addr common.Address, nonce uint64, balance *big.Int) (*state.StateDB, error) {
-       tdb := triedb.NewDatabase(rawdb.NewMemoryDatabase(), nil)
-       sdb, err := state.New(types.EmptyRootHash, state.NewDatabase(tdb, nil))
-       if err != nil {
-               return nil, err
+// accountReader is a minimal state.Database that only returns account data
+// for a single address. All other queries return empty/zero values.
+type accountReader struct {
+       *state.CachingDB
+       addr    common.Address
+       nonce   uint64
+       balance *uint256.Int
+}
+
+func (r *accountReader) TrieDB() *triedb.Database {
+       // needed to satisfy the query inside `state.New`
+       return triedb.NewDatabase(nil, &triedb.Config{
+               IsVerkle: false,
+       })
+}
+
+func (r *accountReader) Reader(root common.Hash) (state.Reader, error) {
+       return r, nil
+}
+
+func (r *accountReader) Account(addr common.Address) (*types.StateAccount, error) {
+       if addr == r.addr {
+               return &types.StateAccount{
+                       Nonce:    r.nonce,
+                       Balance:  r.balance,
+                       Root:     types.EmptyRootHash,
+                       CodeHash: types.EmptyCodeHash.Bytes(),
+               }, nil
        }
-       sdb.SetNonce(addr, nonce, tracing.NonceChangeUnspecified)
+
+       panic(fmt.Sprintf("expected account %s, got %s", r.addr.String(), addr.String()))
+}
+
+// newEphemeralStateDB creates a minimal StateDB using a fake reader that only
+// answers queries for the specified address. This hack avoids constructing a
+// full trie database just to validate a single transaction.
+func newEphemeralStateDB(addr common.Address, nonce uint64, balance *big.Int) (*state.StateDB, error) {
        if balance == nil {
                balance = new(big.Int)
        }
@@ -112,6 +138,28 @@ func newEphemeralStateDB(addr common.Address, nonce uint64, balance *big.Int) (*
        if overflow {
                return nil, fmt.Errorf("balance %s exceeds 256 bits", balance)
        }
-       sdb.SetBalance(addr, bal, tracing.BalanceChangeUnspecified)
-       return sdb, nil
+
+       return state.New(types.EmptyRootHash, &accountReader{
+               addr:    addr,
+               nonce:   nonce,
+               balance: bal,
+       })
+}
+
+// unused methods
+
+func (r *accountReader) Storage(common.Address, common.Hash) (common.Hash, error) {
+       panic("not implemented")
+}
+
+func (r *accountReader) Code(common.Address, common.Hash) ([]byte, error) {
+       panic("not implemented")
+}
+
+func (r *accountReader) CodeSize(common.Address, common.Hash) (int, error) {
+       panic("not implemented")
+}
+
+func (r *accountReader) Copy() state.Reader {
+       panic("not implemented")
 }
```

which works and gets the job done with a smaller footprint, but it's a bit ugly... I can't think of any cleaner and low-footprint ways.

I had a chat with @arner and we decided that the best course of action for now would be to call the stateless validation (`txpool.ValidateTransaction`) as you do now, and then just inline the nonce check from `txpool.ValidateTransactionWithState` so that we no longer have to call that function, and so we can dispense with the expensive stateDB construction.

Would it be okay to go this way? Many thanks in advance!

Storm1289:
@ale-linux ,Thank you for the review.

I’ve made the requested changes:
- Dropped the ephemeral StateDB and removed the ValidateTransactionWithState call
- Kept stateless validation via txpool.ValidateTransaction
- Inlined only the nonce-too-low check
- Removed balance/cost validation (consistent with non-metered gas)

Updated docs/COMPATIBILITY.md:
- Reverted "Sender balance check" to not performed
- Updated the Pre-flight validation section to explicitly list balance validation as a deliberate skip (alongside replacement transactions and nonce gaps)

Please take another look when you have time.

ale-linux (approved):

arner (approved):

ale-linux (approved):

ale-linux:
Thanks a lot @Storm1289 !

PR #118, opened 2026-04-28
add sleep to ensure endorsers are caught up

arner:
As spotted by @Daniel-1600 , there is a flakiness issue in the tests. This PR gives the endorsers enough time to catch up.

https://github.com/hyperledger/fabric-x-evm/pull/109#issuecomment-4320625527

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,12 +63,13 @@ stop-fablo:
 
 .PHONY: test-fablo
 test-fablo:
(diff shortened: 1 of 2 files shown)


arner:
No longer needed now that we use a single synchronizer for the integration tests https://github.com/hyperledger/fabric-x-evm/pull/121

PR #119, opened 2026-04-29
Revert "feat: add Fablo integration test network (#109)"

ale-linux:
This reverts commit cd4080f271b38f5b44819d8629874c994111d678.

Integration tests fail: https://github.com/hyperledger/fabric-x-evm/actions/runs/25074892261/job/73465438636

Let's revert for now so that we can root cause and fix it while having a stable main.


arner (approved):
Agreed, looking into it!

PR #120, opened 2026-04-29
Verify pre-execution errors when running ethereum tests

ale-linux:
This PR asserts that pre-execution errors returned by the gateway are a subset of pre-execution errors returned by the endorser, for the types of transactions that the gateway validation logic accepts.

Whenever we run the ethereum tests, we first call the pre-execution validation in the gateway, and then submit the tx to the endorser. Both these steps can err out, but here we assert that

- if the gateway errs out, so will the endorser
   - there is an exception to this: the gateway currently only supports protected transactions and doesn't support blob transactions, and so in those cases it will err out, whereas the endorser will succeed
- if the endorser errs out, the gateway doesn't necessarily have to since it performs more checks (I verified in geth that `ApplyMessage` performs checks that the pre-mempool-enqueue doesn't


arner (approved):
Clever!

ale-linux:
FYI @Storm1289 , we are now calling the pre-execution validation function you created also for the ethereum compatibility tests

Issue #116, opened 2026-04-28
`TransactionByHash` always returns `false`


ale-linux:
`TransactionByHash` should comply with the contract of geth's homonymous API and not always return `false` as second return argument

Storm1289:
Hi @ale-linux , while working on #41 I noticed `TransactionByHash` always returns `isPending = false` (tracked here). I’ve added a TODO in the tests, and I’m planning to take up #116 next with a follow-up PR.

Storm1289:
Hi @ale-linux, I saw the breakdown of #116 into sub-tasks. I’d be interested in picking up one of the later phases (testing/docs/robustness) once the initial work (e.g., #131) lands. In the meantime, I’ll start working on #46.

PR #122, opened 2026-04-29
integration: add query_validation subtest

### LINKED ISSUE #41: Expand Tests with Query Validation
Enhance existing integration tests to include comprehensive validation of Ethereum JSON-RPC query endpoints in addition to transaction submission. After submitting a transaction and waiting for it to commit, add test assertions that query the transaction by hash using eth_getTransactionByHash and verify all fields match expected values including sender, recipient, value, input data, and status. Query the transaction receipt using eth_getTransactionReceipt and validate the receipt fields including block number, gas used, logs, and status. Query the block using eth_getBlockByNumber and verify the transaction is included in the block with correct metadata. This expanded test coverage ensures the Gateway's query APIs return accurate and complete data consistent with Ethereum client expectations.

---

**Quarter:** Q2

Storm1289:
## Summary
Add systematic test coverage for the gateway's JSON-RPC query surface via a dedicated integration subtest using the native go-ethereum `ethclient.Client`.

## Background
Issue #41 highlights missing direct coverage for read-side APIs. Current tests focus on submit-side behavior, while APIs like:
- `eth_getTransactionByHash`
- `eth_getTransactionReceipt`
- `eth_getTransactionByBlockHashAndIndex`
are only exercised indirectly and not explicitly validated.

### integration/integration_test.go
- Added `testQueryValidation(t, th)` registered as `query_validation`
- Runs under `TestLocal`, `TestLocalX`, `TestFabric`, `TestFabricX`
- Uses native `*ethclient.Client` (`ec`) instead of internal APIs

## Test Coverage

Deploy + call (`Counter.increment`) validated after commit:

### TransactionByHash
- Returns tx with `isPending == false`
- Full equality via `MarshalJSON()` comparison

### TransactionReceipt
- `receipt.Status == types.ReceiptStatusSuccessful`
- Deploy: `receipt.ContractAddress == crypto.CreateAddress(from, nonce)`

### TransactionInBlock
- Matches original tx hash via `(blockHash, index)`

### Negative case
- Unknown hash → `(nil, false, ethereum.NotFound)`

## Note on #116

The behavior around `isPending` is intentionally unchanged in this PR.

A follow-up PR (#127) was opened to address #116, but it was closed based on maintainer guidance since it represents a later stage in a broader workflow requiring earlier phases to land first. The implementation will be revisited once those dependencies are in place.

This PR focuses only on adding test coverage and documents the current behavior (`isPending == false`) without modifying it.

## Known Limitations

- `ec.BlockByHash` / `ec.BlockByNumber` not asserted due to invalid hex encoding in gateway (`LogsBloom`, `ExtraData` not `0x`-prefixed). Tracked for a separate fix.
- Block validation is indirectly covered via receipt + `TransactionInBlock`.

## Test Plan
- [x] `make checks`
- [x] `make unit-tests`
- [x] `go test -run '^TestLocalX$/^query_validation$' ./integration -v`
- [x] `go test -run '^TestLocalX$' ./integration`

All passing.

Closes #41

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: LGPL-3.0-or-later
 package integration
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"math/big"
@@ -15,6 +16,8 @@ import (
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/tests"
 	"github.com/hyperledger/fabric-x-evm/endorser"
 	"github.com/hyperledger/fabric-x-evm/integration/contracts"
@@ -72,6 +75,11 @@ var cases = []testCase{
 	{
 		name:  "nonce_validation",
 		fn:    testNonceValidation,
+		nodes: 2,
+	},
+	{
+		name:  "query_validation",
+		fn:    testQueryValidation,
 		nodes: 2,
 	},
 }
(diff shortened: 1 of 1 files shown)


Storm1289:
Hi @ale-linux ,The quick-tests job was cancelled due to a timeout, not test failures.

`make unit-tests` (`go test ./... -short`) still runs integration tests (`TestLocal`, `TestLocalX`, `TestEthereumTests`, `TestSingleAdd11`) since they don’t check `testing.Short()`. These take ~6–8 minutes on CI and exceed the 10-minute limit.

This is a pre-existing issue and unrelated to this PR.

Possible fixes:
- Add `testing.Short()` guards to those tests
- Or increase `timeout-minutes` from 10 to 20 in `.github/workflows/tests.yml`

integration/integration_test.go
```diff
@@ -769,3 +777,165 @@ func testUniswapFactory(t *testing.T, th *TestHarness) {
...
+		tx     *types.Transaction
+		isCall bool
+	}{
+		{name: "deploy", tx: deployTx, isCall: false},
+		{name: "call", tx: callTx, isCall: true},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			domTx, isPending, err := node.TransactionByHash(t.Context(), c.tx.Hash())
```
ale-linux: I think here we don't want to use the internal `node` API but we want to call the methods of the native ethereum client instance `ec` we got on line 784 and invoke those methods instead

ale-linux: Same for all other calls to `node.` functions below: please replace with the equivalent call on the ethereum client instance.

integration/integration_test.go
```diff
@@ -769,3 +777,165 @@ func testUniswapFactory(t *testing.T, th *TestHarness) {
...
+			}
+			if recovered.Nonce() != c.tx.Nonce() {
+				t.Errorf("nonce mismatch: got %d, want %d", recovered.Nonce(), c.tx.Nonce())
+			}
+			if recovered.Gas() != c.tx.Gas() {
+				t.Errorf("gas mismatch: got %d, want %d", recovered.Gas(), c.tx.Gas())
+			}
+			if recovered.ChainId().Cmp(c.tx.ChainId()) != 0 {
+				t.Errorf("chainId mismatch: got %s, want %s", recovered.ChainId(), c.tx.ChainId())
+			}
```
ale-linux: I think a more compact way to compare all of these fields would be to call the `MarshalJSON` method over the returned transaction and assert equality. It _looks_ like all of the fields are covered (see [here](https://github.com/ethereum/go-ethereum/blob/master/core/types/transaction_marshalling.go#L31-L59)) but it's worth double-checking.

integration/integration_test.go
```diff
@@ -769,3 +777,165 @@ func testUniswapFactory(t *testing.T, th *TestHarness) {
...
+				}
+			}
+
+			byHashIdx, err := node.GetTransactionByBlockHashAndIndex(t.Context(), common.BytesToHash(domTx.BlockHash), domTx.TxIndex)
+			if err != nil {
+				t.Fatalf("GetTransactionByBlockHashAndIndex: %v", err)
+			}
+			if byHashIdx == nil || !bytes.Equal(byHashIdx.TxHash, domTx.TxHash) {
+				t.Errorf("by-(blockHash,index) lookup did not return the same tx")
+			}
```
ale-linux: Same as above - I'd marshal-json and check if equal - same for below

integration/integration_test.go
```diff
@@ -769,3 +777,165 @@ func testUniswapFactory(t *testing.T, th *TestHarness) {
...
+
+	callTx, err := counter.TxForCall(t.Context(), node, &contractAddr, "increment", nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := ec.SendTransaction(t.Context(), callTx); err != nil {
+		t.Fatal(err)
+	}
+	waitForCommitT(t, ec, callTx)
+
```
ale-linux, line 812: We are still missing `TransactionReceipt`, which afaik is the only one that looks at the `Status` flag. We should query it and expect to have a valid status for these txes. 

Btw - an interesting follow-up test for another PR, if you're interested, would be to take a smart contract whose invocations can fail (e.g. the ERC-20 we have), execute a valid and an invalid transaction (e.g. `transfer` from an empty account) and assert that `TransactionReceipt` gives us an invalid status for the latter.

ale-linux (changes requested):

Storm1289:
Hi @ale-linux ,Thanks for the review, I’ve addressed all three points in the latest push:
- Switched to `ethclient` for all lookups:
  - `TransactionByHash`, `TransactionInBlock`, and negative case now handle `ethereum.NotFound`
- Replaced field-by-field checks with `MarshalJSON()` equality
  - Deploy case now validates `receipt.ContractAddress`
- Added `TransactionReceipt` checks for deploy + call
  - Assert `receipt.Status == types.ReceiptStatusSuccessful`
  - Use receipt for block metadata in `TransactionInBlock`

One issue surfaced: `ec.BlockByNumber` / `BlockByHash` fail due to non-hex (`0x`-prefixed) fields in gateway block marshaling (`LogsBloom`, `ExtraData`). I’ve skipped block-fetch here and left a TODO. Happy to send a separate fix PR if preferred.

I’ll follow up with the ERC-20 receipt-status tests as a separate PR and updated the PR

ale-linux (approved):

PR #126, opened 2026-04-29
gateway/api: resolving `BlockNumberOrHash` correctly and reject unknown block hashes. Closes #125

### LINKED ISSUE #125: eth_* state queries ignore blockHash and silently fall back to latest state
Right now, several state-query RPC methods accept a `BlockNumberOrHash` parameter, but when a block hash is provided, it’s effectively ignored. Instead of fetching state from that specific block, the methods quietly fall back to using the latest state. The request still succeeds, which makes the issue harder to detect, but the data returned may be wrong for historical queries.

### I have listed the affected methods

* `eth_getBalance`
* `eth_getCode`
* `eth_getStorageAt`
* `eth_getTransactionCount`
* `eth_call`

### Why this matters

When a client specifies a block hash, it’s usually because they need an exact snapshot of the blockchain at that point in time. Ignoring the hash breaks that expectation and can lead to subtle but serious inconsistencies.

In short, this behavior undermines trust in historical queries: everything looks like it worked correctly, but the data doesn’t actually correspond to the requested block.

rootp1:
## Summary
This PR fixes state-query block selection in the Ethereum RPC API. Closes #125 
Previously, when callers provided a `blockHash` (via `BlockNumberOrHash`), the API could silently fall back to latest state.  
Now, hash-based queries are resolved through `GetBlockByHash`, and unknown hashes return `NotFound` instead of reading latest state.

## Testing
- Ran with local Go 1.26.2 toolchain:
  - `toolchains/go1.26.2/bin/go test ./gateway/api`
- Result: `ok github.com/hyperledger/fabric-x-evm/gateway/api`

### DIFF
--- a/integration/ethereum_test.go
+++ b/integration/ethereum_test.go
@@ -119,10 +119,6 @@ func filterSlowTests(files []string, slow map[string]struct{}, want_very_slow bo
 //
 // This follows the same approach as Besu, Geth, and other Ethereum clients.
 func TestEthereumTests(t *testing.T) {
-	if testing.Short() {
-		t.Skip("skipping ethereum/tests corpus in short mode")
-	}
-
 	grpclog.SetLoggerV2(grpclog.NewLoggerV2(io.Discard, os.Stderr, os.Stderr)) // disable grpc logging
 
 	// Load slow
@@ -137,7 +133,7 @@ func TestEthereumTests(t *testing.T) {
 	// 1) GeneralStateTests
 	testsDir := filepath.Join("..", "testdata", "ethereum-tests", "GeneralStateTests")
 	if _, err := os.Stat(testsDir); os.IsNotExist(err) {
-		t.Skipf("ethereum/tests corpus not found at %s; run `git submodule update --init --recursive`", testsDir)
+		t.Fatalf("ethereum/tests corpus not found at %s; run `git submodule update --init --recursive`", testsDir)
 	}
 	allFiles, err := findJSONFiles(testsDir)
 	if err != nil {
--- a/integration/ethereum_test.go
+++ b/integration/ethereum_test.go
@@ -119,6 +119,10 @@ func filterSlowTests(files []string, slow map[string]struct{}, want_very_slow bo
 //
 // This follows the same approach as Besu, Geth, and other Ethereum clients.
 func TestEthereumTests(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping ethereum/tests corpus in short mode")
+	}
+
 	grpclog.SetLoggerV2(grpclog.NewLoggerV2(io.Discard, os.Stderr, os.Stderr)) // disable grpc logging
 
 	// Load slow
@@ -132,6 +136,9 @@ func TestEthereumTests(t *testing.T) {
 
 	// 1) GeneralStateTests
 	testsDir := filepath.Join("..", "testdata", "ethereum-tests", "GeneralStateTests")
+	if _, err := os.Stat(testsDir); os.IsNotExist(err) {
+		t.Skipf("ethereum/tests corpus not found at %s; run `git submodule update --init --recursive`", testsDir)
+	}
 	allFiles, err := findJSONFiles(testsDir)
 	if err != nil {
 		t.Fatalf("Failed to find test files: %v", err)
--- a/integration/state_test.go
+++ b/integration/state_test.go
@@ -21,6 +21,10 @@ import (
 // TestSingleAdd11 runs only the add11.json test from LegacyTests/Constantinople/GeneralStateTests/stExample
 // This test is ported from go-ethereum's tests/state_test.go
 func TestSingleAdd11(t *testing.T) {
+	if testing.Short() {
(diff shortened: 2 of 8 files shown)


rootp1:
Hey @arner, @ale-linux. 
Please have a look whenever you have a chance to.

integration/ethereum_test.go
```diff
@@ -119,6 +119,10 @@ func filterSlowTests(files []string, slow map[string]struct{}, want_very_slow bo
 //
 // This follows the same approach as Besu, Geth, and other Ethereum clients.
 func TestEthereumTests(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping ethereum/tests corpus in short mode")
+	}
+
```
ale-linux: We already have a way to handle those tests in short mode by randomly skipping a selection (inspired by geth). I don't think we need this, could you pls remove?

integration/state_test.go
```diff
@@ -21,6 +21,10 @@ import (
 // TestSingleAdd11 runs only the add11.json test from LegacyTests/Constantinople/GeneralStateTests/stExample
 // This test is ported from go-ethereum's tests/state_test.go
 func TestSingleAdd11(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping ethereum state test in short mode")
+	}
+
```
ale-linux: We already have a way to handle those tests in short mode by randomly skipping a selection (inspired by geth). I don't think we need this, could you pls remove?

integration/ethereum_test.go
```diff
@@ -132,6 +136,9 @@ func TestEthereumTests(t *testing.T) {
 
 	// 1) GeneralStateTests
 	testsDir := filepath.Join("..", "testdata", "ethereum-tests", "GeneralStateTests")
+	if _, err := os.Stat(testsDir); os.IsNotExist(err) {
+		t.Skipf("ethereum/tests corpus not found at %s; run `git submodule update --init --recursive`", testsDir)
+	}
```
ale-linux, line 137: We don't want to skip those tests but I agree that it would be useful to add instructions somewhere (I guess in the README file) to remind users to run `git submodule update` is useful. Do you want to do that?

integration/state_test.go
```diff
@@ -29,6 +33,9 @@ func TestSingleAdd11(t *testing.T) {
 	// Load the test file
 	file, err := os.Open(testPath)
 	if err != nil {
+		if os.IsNotExist(err) {
+			t.Skipf("ethereum test file not found at %s; run `git submodule update --init --recursive`", testPath)
+		}
```
ale-linux, line 34: We don't want to skip those tests but I agree that it would be useful to add instructions somewhere (I guess in the README file) to remind users to run `git submodule update` is useful. Do you want to do that?

ale-linux (changes requested):
Many thanks for your contribution @rootp1! A couple of comments

integration/ethereum_test.go
rootp1: @ale-linux Implemented, thanks for flagging this.

integration/state_test.go
rootp1: same in here

integration/ethereum_test.go
rootp1: Agreed @ale-linux. I changed this from t.Skipf to t.Fatalf (so we don’t silently skip) and added setup instructions in the main README

integration/state_test.go
rootp1: same in here. Thankyou

rootp1:
> Many thanks for your contribution @rootp1! A couple of comments

Thankyou for the appreciation @ale-linux.

ale-linux (approved):

ale-linux:
Once again, thanks very much @rootp1

rootp1:
> Once again, thanks very much @rootp1

Thankyou for the appreciation @ale-linux,
I just want to confirm, rest of the 70+ issues are open for solving or they are associated to anything else

ale-linux:
in theory they are all open but in practice we have priorities so it would be good to discuss briefly before you start work on any of the open issues, just to avoid duplicate work. Simplest way would be to leave a comment on the issues and we can chat there. We could also communicate on discord and we should probably start having community calls @arner , @kushnireyal

rootp1:
> in theory they are all open but in practice we have priorities so it would be good to discuss briefly before you start work on any of the open issues, just to avoid duplicate work. Simplest way would be to leave a comment on the issues and we can chat there. We could also communicate on discord and we should probably start having community calls @arner , @kushnireyal

Thanks for the clarification @ale-linux ! That makes sense, I’ll make sure to comment on issues before picking them up.

Also, just to share a bit of context from my side,
I’m really hoping to spend my summer contributing more meaningfully, and I’ve been looking for a place where I can apply and improve my Go skills in a real world project. This project seems like a great fit for me, and I loved the combination of EVM, go and Fabric

And yes, community calls would be great opportunity for contributors like me and that would really help with coordination and learning from everyone.

PR #127, opened 2026-04-29
gateway: track pending txs so TransactionByHash returns isPending correctly

### LINKED ISSUE #116: `TransactionByHash` always returns `false`

Storm1289:
## Summary

Fix `Gateway.TransactionByHash` to correctly report pending transactions via the `isPending` flag, aligning with geth’s API behavior.

---

## Background

`Gateway.TransactionByHash` always returned `false` for `isPending`, regardless of whether a transaction was:

- pending (submitted but not yet committed)
- committed
- unknown

This breaks geth compatibility and clients relying on:
- `isPending`
- or `BlockNumber == nil` (ethclient heuristic)

---

### `gateway/core/api.go`

- Added `pendingTxs sync.Map` keyed by `common.Hash`
  - Stores `pendingTx{tx, from}` to avoid repeated sender recovery

- Updated `SendTransaction`:
  - Stores `(tx.Hash(), pendingTx{tx, from})` after validation and before enqueue

- Updated `TransactionByHash` logic:
  - If found in store:
    - delete pending entry (lazy cleanup)
    - return `(tx, false, nil)`
  - If not in store but in `pendingTxs`:
    - return synthesized tx via `pendingDomainTx`
    - return `(tx, true, nil)`
  - Otherwise:
    - return `(nil, false, nil)`

- Added `pendingDomainTx`:
  - Populates:
    - `TxHash`, `FromAddress`, `RawTx`
    - `ToAddress` (calls)
    - or `ContractAddress = crypto.CreateAddress(from, nonce)` (deploys)
  - Leaves block fields empty → signals pending state

---

### `gateway/api/models.go`

- Fixed panic in `rpcTransaction`:
  - `common.Hash(tx.BlockHash)` could panic for empty hash (pending tx)
- Now guarded:
  - `BlockHash`, `BlockNumber`, `TransactionIndex` remain `nil` for pending txs
- Aligns with geth behavior (`BlockNumber == nil` → `isPending=true`)

---

### `gateway/core/api_test.go` (new)

Added unit tests using a stub store:

- `TestTransactionByHash_PendingThenCommitted`
  - Validates transition from pending → committed
  - Ensures pending entry cleanup

- `TestTransactionByHash_DeployPending`
  - Verifies deploy tx returns correct `ContractAddress`

- `TestTransactionByHash_Unknown`
  - Returns `(nil, false, nil)` for unknown hash

---

## Cleanup Behavior

- Pending entries are removed lazily when:
  - transaction is found in store via `TransactionByHash`
- Unqueried txs remain in memory

This matches typical client behavior (they query shortly after submission).  
More aggressive cleanup (TTL or block hook) can be added later if needed.

---

## Test Plan

- [x] `make checks`
- [x] `make unit-tests`
- [x] `go test -run '^TestLocalX$' ./integration -v`
- [x] `go test -run TestTransactionByHash ./gateway/core -v`

All tests passing.

---

Closes #116

### DIFF
--- a/gateway/api/models.go
+++ b/gateway/api/models.go
@@ -160,17 +160,21 @@ func rpcTransaction(tx *domain.Transaction) *RPCTransaction {
 		return nil
 	}
 
-	blockHash := common.Hash(tx.BlockHash)
-	blockNumber := hexutil.Big(*big.NewInt(int64(tx.BlockNumber)))
-	txIndex := hexutil.Uint64(tx.TxIndex)
-
-	return &RPCTransaction{
-		tx:               ethTx,
-		From:             common.BytesToAddress(tx.FromAddress),
-		BlockHash:        &blockHash,
-		BlockNumber:      &blockNumber,
-		TransactionIndex: &txIndex,
+	out := &RPCTransaction{
+		tx:   ethTx,
+		From: common.BytesToAddress(tx.FromAddress),
 	}
+	// Pending txs (no committed block yet) leave block metadata nil so eth
+	// clients see isPending=true.
+	if len(tx.BlockHash) == common.HashLength {
+		blockHash := common.Hash(tx.BlockHash)
+		blockNumber := hexutil.Big(*big.NewInt(int64(tx.BlockNumber)))
+		txIndex := hexutil.Uint64(tx.TxIndex)
+		out.BlockHash = &blockHash
+		out.BlockNumber = &blockNumber
+		out.TransactionIndex = &txIndex
+	}
+	return out
 }
 
 // rpcBlock returns a block in the form the RPC API can return. Some values are mocked.
--- a/gateway/core/api.go
+++ b/gateway/core/api.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/params"
 	cmn "github.com/hyperledger/fabric-x-evm/common"
 	"github.com/hyperledger/fabric-x-evm/gateway/domain"
@@ -46,9 +47,16 @@ type Gateway struct {
 	chainConfig *params.ChainConfig
 	signer      types.Signer
 	txQueue     *TxQueue
+	pendingTxs  sync.Map // common.Hash -> pendingTx
 	workerCount int
 	wg          sync.WaitGroup
 	stopOnce    sync.Once
+}
+
+// pendingTx is the data we keep about a tx accepted by SendTransaction but not yet committed.
+type pendingTx struct {
+	tx   *types.Transaction
+	from common.Address
 }
 
 type Store interface {
@@ -131,6 +139,11 @@ func (g *Gateway) SendTransaction(ctx context.Context, tx *types.Transaction) er
 	if err := ValidateTx(ctx, tx, g.chainConfig, g.signer, g); err != nil {
 		return err
 	}
+	from, err := types.Sender(g.signer, tx)
+	if err != nil {
+		return err
+	}
+	g.pendingTxs.Store(tx.Hash(), pendingTx{tx: tx, from: from})
 	g.txQueue.Enqueue(tx)
 	return nil
 }
@@ -246,25 +259,48 @@ func (g *Gateway) NonceAt(ctx context.Context, account common.Address, blockNumb
 
 // Transactions
 
-// TransactionByHash retrieves a past transaction data from local storage, reconstructs a Transaction object and returns it.
-// We always return false for "is pending".
+// TransactionByHash returns isPending=true for a tx that was accepted by
+// SendTransaction but is not yet visible in a committed block, and
+// isPending=false once the local store has indexed it.
 func (g *Gateway) TransactionByHash(ctx context.Context, hash common.Hash) (*domain.Transaction, bool, error) {
 	tx, err := g.store.GetTransactionByHash(ctx, hash.Bytes())
 	if err != nil {
 		return nil, false, err
 	}
-	if tx == nil {
-		return nil, false, nil
+	if tx != nil {
+		// Lazy cleanup so the pending map doesn't grow forever.
+		g.pendingTxs.Delete(hash)
+
+		logs, err := g.store.GetLogsByTxHash(ctx, hash.Bytes())
+		if err != nil {
+			return nil, false, err
+		}
+		tx.Logs = logs
+		return tx, false, nil
 	}
 
-	// Fetch logs for the transaction (needed for receipts)
-	logs, err := g.store.GetLogsByTxHash(ctx, hash.Bytes())
-	if err != nil {
-		return nil, false, err
+	if v, ok := g.pendingTxs.Load(hash); ok {
+		p := v.(pendingTx)
+		return pendingDomainTx(p.tx, p.from), true, nil
 	}
-	tx.Logs = logs
+	return nil, false, nil
+}
 
-	return tx, false, nil
+// pendingDomainTx synthesises a domain.Transaction for a queued-but-uncommitted
+// tx. Block-context fields (BlockHash/Number, TxIndex, Status, Logs) are zero.
+func pendingDomainTx(tx *types.Transaction, from common.Address) *domain.Transaction {
+	raw, _ := tx.MarshalBinary()
+	d := &domain.Transaction{
+		TxHash:      tx.Hash().Bytes(),
+		RawTx:       raw,
+		FromAddress: from.Bytes(),
+	}
+	if to := tx.To(); to != nil {
+		d.ToAddress = to.Bytes()
+	} else {
+		d.ContractAddress = crypto.CreateAddress(from, tx.Nonce()).Bytes()
+	}
+	return d
 }
 
 // GetTransactionByBlockHashAndIndex retrieves a transaction based on block hash in the transaction index in that block.
--- a/gateway/core/api_test.go
+++ b/gateway/core/api_test.go
@@ -1,0 +1,145 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+*/
+
+package core
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/hyperledger/fabric-x-evm/gateway/domain"
+	"github.com/stretchr/testify/require"
+)
+
+// stubStore is a Store with overridable lookups; everything else is a no-op.
+type stubStore struct {
+	getTxFn   func(ctx context.Context, hash []byte) (*domain.Transaction, error)
+	getLogsFn func(ctx context.Context, txHash []byte) ([]domain.Log, error)
+}
+
+func (s *stubStore) BlockNumber(context.Context) (uint64, error)              { return 0, nil }
+func (s *stubStore) LatestBlock(context.Context, bool) (*domain.Block, error) { return nil, nil }
+func (s *stubStore) GetBlockByNumber(context.Context, uint64, bool) (*domain.Block, error) {
+	return nil, nil
+}
+func (s *stubStore) GetBlockByHash(context.Context, []byte, bool) (*domain.Block, error) {
+	return nil, nil
+}
+func (s *stubStore) GetBlockTxCountByHash(context.Context, []byte) (int64, error)   { return 0, nil }
+func (s *stubStore) GetBlockTxCountByNumber(context.Context, uint64) (int64, error) { return 0, nil }
+func (s *stubStore) GetTransactionByHash(ctx context.Context, hash []byte) (*domain.Transaction, error) {
+	if s.getTxFn != nil {
+		return s.getTxFn(ctx, hash)
+	}
+	return nil, nil
+}
+func (s *stubStore) GetTransactionByBlockHashAndIndex(context.Context, []byte, int64) (*domain.Transaction, error) {
+	return nil, nil
+}
+func (s *stubStore) GetTransactionByBlockNumberAndIndex(context.Context, uint64, int64) (*domain.Transaction, error) {
+	return nil, nil
+}
+func (s *stubStore) GetLogs(context.Context, domain.LogFilter) ([]domain.Log, error) { return nil, nil }
+func (s *stubStore) GetLogsByTxHash(ctx context.Context, txHash []byte) ([]domain.Log, error) {
+	if s.getLogsFn != nil {
+		return s.getLogsFn(ctx, txHash)
+	}
+	return nil, nil
+}
+
+func newTestGateway(t *testing.T, store Store) *Gateway {
+	t.Helper()
+	gw, err := New(nil, nil, store, testChainID, 1)
+	require.NoError(t, err)
+	return gw
+}
+
+func signedTx(t *testing.T, key *ecdsa.PrivateKey, to *common.Address) *types.Transaction {
+	t.Helper()
+	var inner *types.LegacyTx
+	if to != nil {
+		inner = &types.LegacyTx{Nonce: 7, To: to, Gas: 21_000, GasPrice: big.NewInt(1), Value: big.NewInt(0)}
+	} else {
+		inner = &types.LegacyTx{Nonce: 7, Gas: 60_000, GasPrice: big.NewInt(1), Data: []byte{0x60}}
+	}
+	signed, err := types.SignTx(types.NewTx(inner), types.NewEIP155Signer(big.NewInt(testChainID)), key)
+	require.NoError(t, err)
+	return signed
+}
+
+func TestTransactionByHash_PendingThenCommitted(t *testing.T) {
+	store := &stubStore{}
+	gw := newTestGateway(t, store)
+
+	key := newKey(t)
+	to := common.HexToAddress("0x2222222222222222222222222222222222222222")
+	tx := signedTx(t, key, &to)
+	from := crypto.PubkeyToAddress(key.PublicKey)
+
+	gw.pendingTxs.Store(tx.Hash(), pendingTx{tx: tx, from: from})
+
+	domTx, isPending, err := gw.TransactionByHash(context.Background(), tx.Hash())
+	require.NoError(t, err)
+	require.True(t, isPending)
+	require.NotNil(t, domTx)
+	require.Equal(t, tx.Hash().Bytes(), domTx.TxHash)
+	require.Equal(t, from.Bytes(), domTx.FromAddress)
+	require.Equal(t, to.Bytes(), domTx.ToAddress)
+	require.NotEmpty(t, domTx.RawTx)
+
+	// Simulate commit: store now returns the tx → isPending must flip and the
+	// pending entry must be cleaned up.
+	committed := &domain.Transaction{
+		TxHash:      tx.Hash().Bytes(),
+		BlockNumber: 42,
+		BlockHash:   []byte("block-hash"),
+		Status:      1,
+		FromAddress: from.Bytes(),
+		ToAddress:   to.Bytes(),
+	}
+	store.getTxFn = func(_ context.Context, _ []byte) (*domain.Transaction, error) {
+		return committed, nil
+	}
+
+	domTx, isPending, err = gw.TransactionByHash(context.Background(), tx.Hash())
+	require.NoError(t, err)
+	require.False(t, isPending)
+	require.Equal(t, committed, domTx)
+
+	_, stillPending := gw.pendingTxs.Load(tx.Hash())
+	require.False(t, stillPending, "pending entry should be cleared once committed")
+}
+
+func TestTransactionByHash_DeployPending(t *testing.T) {
+	gw := newTestGateway(t, &stubStore{})
+
+	key := newKey(t)
+	tx := signedTx(t, key, nil) // deploy
+	from := crypto.PubkeyToAddress(key.PublicKey)
+
+	gw.pendingTxs.Store(tx.Hash(), pendingTx{tx: tx, from: from})
+
+	domTx, isPending, err := gw.TransactionByHash(context.Background(), tx.Hash())
+	require.NoError(t, err)
+	require.True(t, isPending)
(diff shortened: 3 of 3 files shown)


Storm1289:
Hi @ale-linux @arner ,This is a Follow-up to #113 / #122, addresses #116.
The fix is small: `SendTransaction` stores accepted txs in a `sync.Map`, and `TransactionByHash` checks it before falling back to "not found".
One issue surfaced during this: `rpcTransaction` in `gateway/api/models.go` used `common.Hash(tx.BlockHash)`, which panics on short slices. This only appeared once a real pending tx (with empty `BlockHash`) reached the RPC layer. The conversion is now guarded, and pending txs correctly return `BlockNumber=nil`, matching geth behavior (used by `ethclient` to set `isPending=true`).
Pending entries are currently cleaned up lazily on the next `TransactionByHash` call, so a tx that is never queried will remain in memory. I suggest either adding a TTL-based cleanup or hooking into the synchronizer’s block-applied path for deterministic removal happy to implement either approach based on your preference.

arner:
While it's nice to be able to return more geth-like (or generic ethereum-style) results, @ale-linux and I had a chat and think it might be a bit early for this, now that we're still working on the design for the mempool solution, retries and other things.

I propose to close this PR for now, until the mempool solution is solidified a bit more and we can go for a more holistic approach.

ale-linux:
I gave it some thought and came up with a better approach which already makes use of the existing mempool queue we have and goes in the direction where we want that queue to go. I typed it up and had an AI assistant polish that. here goes:

# Detailed Implementation Plan: Transaction Pending Status Tracking

## Introduction: High-Level Approach

### The Problem
Currently, `TransactionByHash()` in `gateway/core/api.go` always returns `isPending = false`, which is incorrect. Ethereum clients expect transactions to be marked as "pending" from the moment they're submitted until they're included in a committed block. This is critical for user experience, as clients poll this endpoint to track transaction status.

### The Solution Architecture
We will implement a **two-queue tracking system** combined with a **callback notification mechanism**:

1. **Two-Queue System**: Instead of immediately removing transactions from the queue when workers pick them up, we'll maintain two data structures:
   - A **pending queue** (FIFO) for transactions waiting to be processed
   - An **in-progress map** (hash → transaction) for transactions currently being endorsed and submitted by workers

2. **Pending Status Query**: When `TransactionByHash()` is called, it will first check if the transaction exists in either the pending queue or in-progress map. If found, it returns `isPending = true`. Otherwise, it queries the database for committed transactions.

3. **Completion Notification**: When a block is committed (in `Chain.Handle()`), we extract all transaction hashes from that block and invoke a callback to notify the Gateway. The Gateway then removes these transactions from the in-progress map, completing the lifecycle.

4. **Component Wiring**: The Gateway and Chain components need to be connected so the Chain can notify the Gateway about completed transactions. This requires passing a Chain reference to the Gateway and registering a callback.

### Transaction Lifecycle States
- **Pending (in queue)**: Transaction submitted via `SendTransaction()`, waiting in `pendingQueue`
- **In Progress (being processed)**: Worker has dequeued the transaction, moved it to `inProgressMap`, and is endorsing/submitting it
- **Completed (in database)**: Block containing the transaction has been committed, transaction removed from `inProgressMap` and stored in database
- **Not Found**: Transaction hash doesn't exist in any of the above locations

---

## Phase 1: Refactor TxQueue to Support Pending State Tracking

**Goal**: Transform the simple FIFO queue into a sophisticated two-queue system that can track both pending and in-progress transactions, enabling the Gateway to answer "is this transaction pending?" queries.

### Task 1.1: Modify TxQueue Structure
**What to do**: Refactor the `TxQueue` struct in `gateway/core/txqueue.go` to maintain two separate data structures instead of one.

**Details**:
- Replace the single `queue []*types.Transaction` slice with:
  - `pendingQueue []*types.Transaction` - FIFO queue for transactions waiting to be processed
  - `inProgressMap map[common.Hash]*types.Transaction` - Map for fast lookup of transactions currently being processed by workers
- Keep the existing `mu sync.Mutex` and `cond *sync.Cond` for synchronization
- Keep the `done bool` flag for shutdown signaling
- The map allows O(1) lookup by transaction hash, which is essential for the `IsPending()` query

**Why**: The current implementation loses track of transactions once they're dequeued. By keeping them in `inProgressMap`, we can answer pending status queries accurately.

### Task 1.2: Add IsPending Method
**What to do**: Create a new public method `IsPending(txHash common.Hash) bool` that checks if a transaction is in either the pending queue or in-progress map.

**Details**:
- Lock the mutex before checking both data structures
- Iterate through `pendingQueue` to check if any transaction has the matching hash
- Check if the hash exists as a key in `inProgressMap`
- Return `true` if found in either location, `false` otherwise
- This method will be called by `TransactionByHash()` to determine pending status

**Why**: This is the core query method that enables the Gateway to report accurate pending status to Ethereum clients.

### Task 1.3: Modify Dequeue to Mark In-Progress
**What to do**: Change the `Dequeue()` method's behavior from "remove and return" to "move to in-progress and return".

**Details**:
- When a transaction is dequeued from `pendingQueue`, don't discard it
- Instead, compute its hash using `tx.Hash()` and add it to `inProgressMap[hash] = tx`
- Remove it from the front of `pendingQueue` as before
- Return the transaction to the worker for processing
- The transaction now exists only in `inProgressMap` until it's completed or fails

**Why**: This keeps the transaction tracked during the endorsement and submission process, which can take several seconds. Without this, there's a gap where the transaction appears to not exist.

### Task 1.4: Add MarkCompleted Method
**What to do**: Create a new public method `MarkCompleted(txHash common.Hash)` that removes a transaction from the in-progress map.

**Details**:
- Lock the mutex
- Delete the entry from `inProgressMap` using `delete(inProgressMap, txHash)`
- This method will be called by the Gateway's callback when a block is committed
- Should be idempotent (safe to call multiple times with the same hash)

**Why**: This completes the transaction lifecycle. Once a transaction is in the database, it should no longer be tracked as pending.

### Task 1.5: Update Close Method
**What to do**: Modify the `Close()` method to properly handle both the pending queue and in-progress map during shutdown.

**Details**:
- Set `done = true` as before
- Broadcast to wake up all waiting workers
- Consider whether to clear `inProgressMap` or leave it (document the decision)
- Add comments explaining what happens to in-progress transactions on shutdown

**Why**: Clean shutdown behavior is important. In-progress transactions at shutdown time will be lost (they'll be resubmitted by clients if needed), but this should be documented.

### Task 1.6: Thread-Safety Review
**What to do**: Review and ensure all new map operations are properly protected by the mutex.

**Details**:
- Verify that every access to `inProgressMap` (read or write) is within a `mu.Lock()`/`mu.Unlock()` block
- Consider potential deadlocks if callbacks are invoked while holding the lock
- Add comments documenting the locking strategy
- Consider whether `IsPending()` could be called from multiple goroutines simultaneously

**Why**: Maps in Go are not thread-safe. Concurrent access without proper locking will cause panics and data races.

---

## Phase 2: Integrate Pending Status into TransactionByHash

**Goal**: Modify the `TransactionByHash()` API method to accurately report whether a transaction is pending by consulting the queue before querying the database.

### Task 2.1: Modify TransactionByHash Logic
**What to do**: Update the `TransactionByHash()` method in `gateway/core/api.go` to check the queue first.

**Details**:
- At the start of the method, call `isPending := g.txQueue.IsPending(hash)`
- If `isPending` is true, return `nil, true, nil` immediately
  - First return value (nil): No transaction data available yet (it hasn't been committed)
  - Second return value (true): Transaction is pending
  - Third return value (nil): No error occurred
- If not pending, proceed with the existing database query logic
- If found in database, return the transaction with `isPending = false`

**Why**: This implements the correct Ethereum behavior where pending transactions are reported as such, and clients can distinguish between "pending", "completed", and "not found" states.

### Task 2.2: Handle Three Transaction States
**What to do**: Ensure the return logic properly distinguishes between all three possible states.

**Details**:
- **State 1 - Pending**: Transaction in queue → return `nil, true, nil`
- **State 2 - Completed**: Transaction in database → return `tx, false, nil`
- **State 3 - Not Found**: Transaction in neither location → return `nil, false, nil`
- Add comments explaining each case
- Consider logging for debugging (e.g., "Transaction %s is pending" or "Transaction %s not found")

**Why**: Ethereum clients rely on these three states to provide accurate feedback to users. Incorrect status reporting breaks user experience.

---

## Phase 3: Create Callback Mechanism for Transaction Completion

**Goal**: Establish a notification system where the Chain component can inform the Gateway when transactions have been committed to blocks, enabling the Gateway to update its pending transaction tracking.

### Task 3.1: Define Callback Type
**What to do**: Define a callback function type in `gateway/core/api.go` that will be invoked when transactions are completed.

**Details**:
- Add type definition: `type TxCompletionCallback func(txHashes []common.Hash)`
- The callback receives a slice of transaction hashes that were just committed
- Document that this callback should be fast and non-blocking
- Consider whether the callback should return an error (probably not needed)

**Why**: A well-defined callback interface makes the contract between Chain and Gateway explicit and type-safe.

### Task 3.2: Add Callback Registration to Gateway
**What to do**: Add a method to the Gateway struct that allows external components to register a completion callback.

**Details**:
- Add a field to Gateway struct: `txCompletionCallback TxCompletionCallback`
- Add method: `func (g *Gateway) RegisterTxCompletionCallback(callback TxCompletionCallback)`
- Store the callback in the struct field
- Consider thread-safety if the callback might be registered after Start() is called

**Why**: This allows the Chain to register its callback with the Gateway, creating the notification pathway.

### Task 3.3: Modify Chain.Handle to Extract and Notify
**What to do**: Update the `Handle()` method in `gateway/core/chain.go` to extract transaction hashes from committed blocks and invoke the callback.

**Details**:
- After successfully inserting the block (after `c.Store.InsertBlock(ctx, ebl)`), extract transaction hashes
- Iterate through `ebl.Transactions` and collect all `TxHash` values into a `[]common.Hash` slice
- If a callback is registered, invoke it: `if c.txCompletionCallback != nil { c.txCompletionCallback(txHashes) }`
- Ensure callback errors don't break block processing (wrap in recover or handle errors gracefully)
- Add logging for debugging: "Notifying completion of %d transactions"

**Why**: This is where the Chain notifies the Gateway that transactions have been finalized, triggering the removal from the in-progress map.

### Task 3.4: Add Callback Support to Chain Struct
**What to do**: Add the necessary fields and methods to the Chain struct to support callback registration.

**Details**:
- Add field to Chain struct: `txCompletionCallback TxCompletionCallback`
- Add method: `func (c *Chain) SetTxCompletionCallback(callback TxCompletionCallback)`
- Store the callback in the struct field
- Document that the callback is optional (Chain can work without it)
- Consider thread-safety if callback might be set after Chain is already processing blocks

**Why**: The Chain needs to store the callback so it can invoke it during block processing.

---

## Phase 4: Wire Gateway and Chain Together

**Goal**: Connect the Gateway and Chain components so they can communicate about transaction completion, establishing the full notification pipeline from block commitment to pending status updates.

### Task 4.1: Add Chain Reference to Gateway
**What to do**: Modify the Gateway struct to hold a reference to the Chain instance.

**Details**:
- Add field to Gateway struct: `chain *Chain`
- This creates a dependency from Gateway to Chain
- Document why this reference is needed (for callback registration)
- Consider whether this creates any circular dependency issues (Gateway → Chain → Store, Gateway uses Store)

**Why**: The Gateway needs access to the Chain to register its callback during initialization.

### Task 4.2: Update Gateway Constructor
**What to do**: Modify the `New()` function in `gateway/core/api.go` to accept a Chain parameter.

**Details**:
- Change signature from: `func New(ec *EndorsementClient, submitter Submitter, store Store, chainID int64, workerCount int) (*Gateway, error)`
- To: `func New(ec *EndorsementClient, submitter Submitter, store Store, chain *Chain, chainID int64, workerCount int) (*Gateway, error)`
- Store the chain parameter in `g.chain = chain`
- Update the return statement to include the chain field
- Note: The Store parameter might become redundant since Chain embeds Store, but keep it for now for backward compatibility

**Why**: The Gateway needs the Chain reference at construction time to set up the callback relationship.

### Task 4.3: Register Callback in Gateway Initialization
**What to do**: In the Gateway's `Start()` method or immediately after construction, register a callback with the Chain that removes completed transactions from the queue.

**Details**:
- Add code after Gateway construction or in `Start()`:
  ```go
  g.chain.SetTxCompletionCallback(func(txHashes []common.Hash) {
      for _, hash := range txHashes {
          g.txQueue.MarkCompleted(hash)
      }
  })
  ```
- Consider adding logging: "Registered transaction completion callback"
- Ensure this happens before any blocks are processed
- Consider whether to do this in `New()` or `Start()` (probably `New()` is safer)

**Why**: This is the critical connection that closes the loop: when Chain commits a block, it notifies Gateway, which updates the queue.

### Task 4.4: Update Gateway Instantiation Sites
**What to do**: Find all places where `Gateway.New()` is called and update them to pass the Chain instance.

**Details**:
- Search for calls to `core.New()` or `gateway.New()` (likely in `cmd/` directory)
- Update each call site to pass the Chain instance as a parameter
- Ensure Chain is created before Gateway in the initialization sequence
- Update any tests that create Gateway instances
- Verify compilation after changes

**Why**: The API change to the constructor requires updating all call sites, or the code won't compile.

---

## Phase 5: Handle Edge Cases and Error Scenarios

**Goal**: Ensure the system behaves correctly in failure scenarios, prevents data corruption, and handles edge cases gracefully.

### Task 5.1: Handle Transaction Processing Failures
**What to do**: Decide what happens when `processTx()` fails and implement appropriate cleanup.

**Details**:
- In `worker()` method in `gateway/core/api.go`, when `processTx()` returns an error:
  - Here we shouldn't actually do anything yet - we should clearly document in the code using TODO and FIXME comments to highlight the error scenarios.

### Task 5.2: Prevent Duplicate Submissions
**What to do**: Add duplicate detection to `SendTransaction()` to prevent the same transaction from being queued multiple times.

**Details**:
- In `SendTransaction()` in `gateway/core/api.go`, before calling `g.txQueue.Enqueue(tx)`:
  - Check if transaction already exists: `if g.txQueue.IsPending(tx.Hash()) { return fmt.Errorf("transaction already pending") }`
- Consider whether this should be an error or silently succeed
- Add test case for duplicate submission
- Document the behavior in comments

**Why**: Duplicate submissions waste resources and could cause confusion. Ethereum clients typically don't resubmit the same transaction, but defensive programming is good.

### Task 5.3: Document Gateway Restart Behavior
**What to do**: Clearly document what happens to pending and in-progress transactions when the Gateway is restarted.

**Details**:
- Add comments to `TxQueue` explaining that the queue is in-memory only
- Document that on restart:
  - Pending queue is empty (transactions lost)
  - In-progress map is empty (transactions lost)
  - Clients will resubmit transactions if they don't see them confirmed
- Consider whether to add persistence in the future (probably not needed for MVP)
- Add a note in the README or architecture docs about this behavior

**Why**: Operators need to understand the implications of restarting the Gateway. Lost pending transactions are acceptable since clients handle resubmission.

### Task 5.4: Handle Callback Errors Gracefully
**What to do**: Ensure that errors in the callback don't break block processing in the Chain.

**Details**:
- In `Chain.Handle()`, wrap the callback invocation in error handling:
  ```go
  if c.txCompletionCallback != nil {
      defer func() {
          if r := recover(); r != nil {
              log.Printf("Transaction completion callback panicked: %v", r)
          }
      }()
      c.txCompletionCallback(txHashes)
  }
  ```
- Consider whether callback should return an error (probably not - it's fire-and-forget)
- Add logging if callback fails
- Document that callback failures are logged but don't affect block processing

**Why**: Block processing is critical and must not fail due to callback issues. The callback is a convenience feature, not a requirement for correctness.

---

## Phase 6: Testing and Validation

**Goal**: Ensure the implementation works correctly through comprehensive unit and integration tests, covering both happy paths and edge cases.

### Task 6.1: Unit Tests for TxQueue
**What to do**: Create comprehensive unit tests for all new TxQueue methods in `gateway/core/txqueue_test.go`.

**Details**:
- Test `IsPending()`:
  - Returns true for transaction in pending queue
  - Returns true for transaction in in-progress map
  - Returns false for transaction in neither
- Test `Dequeue()` moves transaction from pending to in-progress
- Test `MarkCompleted()` removes transaction from in-progress
- Test concurrent access:
  - Multiple goroutines calling `IsPending()` simultaneously
  - Enqueue and Dequeue happening concurrently
  - MarkCompleted called while IsPending is running
- Use Go's race detector: `go test -race`

**Why**: Unit tests catch bugs early and ensure thread-safety, which is critical for concurrent code.

### Task 6.2: Integration Test for Pending Transaction Flow
**What to do**: Create an end-to-end integration test that verifies the complete pending transaction lifecycle.

**Details**:
- Test scenario:
  1. Start Gateway and Chain
  2. Submit a transaction via `SendTransaction()`
  3. Immediately call `TransactionByHash()` - should return `isPending = true`
  4. Wait for block to be committed (may need to trigger block creation in test)
  5. Call `TransactionByHash()` again - should return `isPending = false` with transaction data
- Add assertions at each step
- Use test helpers to create and sign transactions
- Consider adding timing assertions (pending status should last at least X milliseconds)

**Why**: Integration tests verify that all components work together correctly, catching issues that unit tests miss.

### Task 6.3: Test Callback Mechanism
**What to do**: Create tests specifically for the callback notification system.

**Details**:
- Test that callback is invoked when block is committed
- Test that callback receives correct transaction hashes
- Test that transactions are removed from in-progress map after callback
- Test callback with multiple transactions in one block
- Test callback with empty block (no transactions)
- Test that callback errors don't break block processing (use a callback that panics)

**Why**: The callback mechanism is the critical link between Chain and Gateway. Bugs here would break pending status tracking.

### Task 6.4: Test Edge Cases
**What to do**: Create tests for all the edge cases identified in Phase 5.

**Details**:
- Test transaction failure during processing:
  - Mock `processTx()` to return an error
  - Verify transaction is removed from in-progress map
  - Verify subsequent `IsPending()` returns false
- Test duplicate submission:
  - Submit same transaction twice
  - Verify second submission is rejected or handled correctly
- Test Gateway shutdown with pending transactions:
  - Enqueue transactions
  - Call `Stop()`
  - Verify clean shutdown
  - Document that pending transactions are lost
- Test race conditions using `-race` flag

**Why**: Edge cases are where bugs hide. Thorough testing of failure scenarios ensures robustness in production.

---

## Phase 7: Documentation and Cleanup

**Goal**: Ensure the code is well-documented, maintainable, and ready for production use by adding comprehensive comments and updating documentation.

### Task 7.1: Document Transaction Lifecycle
**What to do**: Add detailed comments explaining the complete transaction lifecycle from submission to completion.

**Details**:
- Add a package-level comment in `gateway/core/api.go` or `txqueue.go` explaining:
  - The four states: pending, in-progress, completed, not found
  - How transactions move between states
  - The role of the callback in state transitions
- Add inline comments in key methods (`SendTransaction`, `worker`, `TransactionByHash`)
- Create a diagram if helpful (ASCII art or reference to external doc)

**Why**: Future maintainers need to understand the design to make changes safely. The lifecycle is complex enough to warrant detailed documentation.

### Task 7.2: Document Two-Queue Architecture
**What to do**: Add comprehensive documentation to the `TxQueue` struct and file explaining the two-queue design.

**Details**:
- Add a comment block above the `TxQueue` struct explaining:
  - Why two data structures are needed
  - The purpose of each (pending queue vs in-progress map)
  - Thread-safety guarantees
  - Performance characteristics (O(1) lookup, O(n) pending check)
- Document the locking strategy
- Explain why transactions aren't immediately removed on dequeue

**Why**: The two-queue design is non-obvious. Clear documentation prevents future developers from "simplifying" it back to a single queue.

### Task 7.3: Add Godoc Comments
**What to do**: Add or improve godoc comments for all new public methods and types.

**Details**:
- Add godoc comments for:
  - `TxCompletionCallback` type
  - `IsPending()` method
  - `MarkCompleted()` method
  - `RegisterTxCompletionCallback()` method
  - `SetTxCompletionCallback()` method
- Follow Go conventions (start with method name, describe what it does)
- Include examples if helpful
- Run `go doc` to verify formatting

**Why**: Godoc is the standard Go documentation system. Good godoc comments make the API self-documenting.

### Task 7.4: Update Architecture Documentation
**What to do**: Update any existing architecture documentation (like `docs/ARCHITECTURE.md`) to reflect the new pending transaction tracking system.

**Details**:
- Add a section on "Transaction Lifecycle and Pending Status"
- Explain the Gateway-Chain callback relationship
- Document the in-memory nature of pending tracking (lost on restart)
- Add diagrams showing the flow from submission to completion
- Update any sequence diagrams to include the callback

**Why**: Architecture documentation helps new team members understand the system design and makes design reviews easier.

### Task 7.5: Clean Up TODO Comments
**What to do**: Find and update or remove any TODO comments related to pending transaction status.

**Details**:
- Search codebase for TODO comments mentioning "pending" or "transaction status"
- In `gateway/core/api.go` line 256, there's a comment "We always return false for 'is pending'" - update or remove this
- Remove any temporary TODO comments added during implementation
- Add new TODOs for future improvements if needed (e.g., "TODO: Consider persisting pending transactions for restart resilience")

**Why**: Outdated TODO comments are confusing and misleading. Clean code has accurate, up-to-date comments.

---

## Summary

This plan provides a complete roadmap for implementing proper transaction pending status tracking. The implementation follows a logical progression:

1. **Foundation** (Phase 1): Build the two-queue tracking system
2. **Integration** (Phase 2): Connect the tracking to the API
3. **Notification** (Phase 3): Create the callback mechanism
4. **Connection** (Phase 4): Wire everything together
5. **Robustness** (Phase 5): Handle edge cases
6. **Validation** (Phase 6): Test thoroughly
7. **Polish** (Phase 7): Document everything

Each phase builds on the previous one, and tasks within each phase can often be done in parallel. The total effort is 29 discrete tasks organized into 7 logical phases.

arner:
Looks like a nice solution! For the implementation for Task 4: instead of a callback, we can also consider registering the completion callback as another (final) block handler, to keep the block processing pipeline simple and have looser coupling. If that works with the constructor orchestration at least. Not sure if that would be better, but something to consider at least.

ale-linux:
@arner I completely agree with you: no callback, just registering a new `Handle` function (could simply be the queue itself implementing that) that parses the block, gets txes and bumps them from the pending map...

rootp1:
@arner, @ale-linux, if it is okay, can I start working with the phase 1, task 1.1

ale-linux:
Sure - we can then close this one. I've updated https://github.com/hyperledger/fabric-x-evm/issues/116 with sub-tasks breakdown as per our conversation - anyone can pick them up and work on them

Storm1289:
Thanks for the thorough plan @ale-linux and the simplification @arner agreed, the two-queue + block-handler approach is cleaner and lines up with the mempool direction. I'll close this PR. Happy to pick up one of the later phases on #116 once @rootp1's Task 1.1 lands.

PR #130, opened 2026-04-30
Add Hardhat Test Support for OpenZeppelin ERC20

kushnireyal:
This PR adds support for running Hardhat tests against Fabric-EVM using OpenZeppelin's ERC20 test suite.

## Changes

- **Test Infrastructure**: Added `make hardhat-tests` target and test script
- **OpenZeppelin Integration**: Added OpenZeppelin contracts as a submodule
- **Test RPC Support**: Implemented `eth_accounts` and `eth_sendTransaction` for Hardhat compatibility
- **Hardhat Helpers**: Added stub implementations for Hardhat-specific RPC methods

## Test Results

- ✅ **58 tests passing** - Core ERC20 functionality (transfers, approvals, minting, burning, events)
- ⏭️ **~50 tests skipped** - Revert detection tests (known Fabric-EVM limitation)

## Known Limitations

- Transaction reverts are not reported like Ethereum, causing revert-expecting tests to timeout
- Tests are skipped using Mocha grep pattern (documented in script)
- Some Hardhat helper methods are stubs (evm_snapshot, evm_revert, etc.) but not used by passing tests

## Usage

```bash
make hardhat-tests

### CODE REVIEW

File: cmd/fxevm/main.go [hunk 1]
```diff
@@ -49,34 +48,49 @@ func main() {
...
 	var protocol string
+	var enableTestRPC bool
+	var testAccountsPath string
 
 	cmd := &cobra.Command{
 		Use:   "start",
 		Short: "Start the EVM gateway with embedded endorsers (single-process mode)",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runStart(cmd.Context(), protocol)
+			return runStart(cmd.Context(), protocol, enableTestRPC, testAccountsPath)
```
arner: Considering this is a completely different mode of starting the network, I propose to have a completely different entrypoint. Like, instead of `fxevm start`, do `fxevm testnode` or something. We can then (safely) have all the test settings we like. 

Hot take: we can probably even use the `local` backend from the SDK instead of a real Fabric-X backend! This would be the developer mode, where people can test their smart contracts against a system that behaves exactly like a Fabric-X EVM network. With the additional test features like the testaccounts, impersonate, snapshot/revert, etc. And it can just run in a single go process or container, without even a config file. And we still learn the same things about compatibility. What do you think?

gateway/testimpl/hardhat_helpers.go
```diff
@@ -0,0 +1,122 @@
...
+func (api *EvmAPI) Snapshot(ctx context.Context) (string, error) {
+	api.mu.Lock()
+	defer api.mu.Unlock()
+
+	api.snapshotCounter++
+	snapshotID := fmt.Sprintf("0x%x", api.snapshotCounter)
+
+	// Stub: return snapshot ID without actually preserving state
+	// In a full implementation, this would:
+	// 1. Create a copy of the current StateDB using StateDB.Copy()
```
arner, line 74: We can talk about the best way to do this (when the time comes), we may be able to just store the BlockHeight when this is called, and when they request the revert, we just delete all world state with a higher block number than that point.

gateway/testimpl/hardhat_helpers.go
```diff
@@ -0,0 +1,122 @@
...
+	// 2. Restore the StateDB from the snapshot
+	// 3. Remove snapshots created after this one
+	// 4. Return success
+
+	return true, nil
+}
+
+// Mine mines a new block (evm_mine).
+// This is a stub that returns success. In fabric-evm, blocks are created
+// by the Fabric consensus mechanism, not by mining.
```
arner, line 101: This might be challenging but if we go for the local backend instead of real Fabric-X, we can add hooks to trigger a block creation.

gateway/testimpl/hardhat_helpers.go
```diff
@@ -0,0 +1,122 @@
...
+func (api *EvmAPI) Mine(ctx context.Context) (string, error) {
+	// Stub: In fabric-evm, blocks are created by Fabric consensus
+	// Return success to allow tests to proceed
+	return "0x0", nil
+}
+
+// IncreaseTime increases the timestamp of the next block (evm_increaseTime).
+// This is a stub that returns the time increase amount.
+func (api *EvmAPI) IncreaseTime(ctx context.Context, seconds hexutil.Uint64) (hexutil.Uint64, error) {
```
arner, line 110: Oh dear... Bigger discussion on block timestamps.... Still not sure if we can have them consistently across the network! (TBD)

integration/config.go
```diff
@@ -98,8 +98,14 @@ func FabricSamplesConfig(testdataDir string) config.Config {
 				MspID:  "Org1MSP",
 				MSPDir: path.Join(user, "msp"),
 			},
-			DbConnStr:  "file:gateway.db?mode=memory&cache=shared",
-			TrieDBPath: filepath.Join(testdataDir, "triedb"),
+			DbConnStr:        "file:gateway.db?mode=memory&cache=shared",
+			TrieDBPath:       filepath.Join(testdataDir, "triedb"),
```
arner, line 102: It's not consistent that we store the TrieDB but only keep the block db in memory. That causes issue every time we try to run the tests twice.

Propose to have an empty string for the triedb (which makes it default to memory)

arner (commented):
Ok, this requires some more changes than I expected! IMO if we manage to keep the code for testing clean (and as isolated as possible), it's a good feature to offer, to give developers confidence that their contracts work as expected.

gateway/api/eth.go
```diff
@@ -247,7 +247,9 @@ func (api *EthAPI) Call(ctx context.Context, args map[string]any, block rpc.Bloc
 
 // eth_estimateGas
 func (api *EthAPI) EstimateGas(ctx context.Context, args map[string]any, block *rpc.BlockNumberOrHash) (*hexutil.Uint64, error) {
-	u := hexutil.Uint64(0)
+	// Return a reasonable default gas limit for contract deployments and transactions
+	// This is a simplified implementation - a full implementation would simulate the transaction
+	u := hexutil.Uint64(10000000) // 10M gas should be sufficient for most operations
```
ale-linux: Can't you override this into gateway/testimpl/test_eth_api.go?

ale-linux (commented):
Sw33t! 👍 Thanks @kushnireyal, LGTM modulo maybe moving the gas estimate override to the testimpl.

gateway/api/eth.go
```diff
@@ -248,9 +248,9 @@ func (api *EthAPI) Call(ctx context.Context, args map[string]any, block rpc.Bloc
 
 // eth_estimateGas
 func (api *EthAPI) EstimateGas(ctx context.Context, args map[string]any, block *rpc.BlockNumberOrHash) (*hexutil.Uint64, error) {
-	// Gas is not metered; return a constant that always satisfies the intrinsic-gas
-	// check in ValidateTx. Matches the endorser's own zero-gas fallback.
-	u := hexutil.Uint64(5_000_000)
+	// TODO: Implement proper gas estimation by simulating the transaction
+	// For now, return 0 to indicate gas estimation is not implemented
+	u := hexutil.Uint64(0)
```
arner: If at zero, `cast` fails to submit transactions (and so does Metamask if I remember correctly). It decides to create transactions with zero gas, but now that we enabled more checks in ValidateTx, it rejects those.

arner: Sorry just realised this was about the same thing that @ale-linux just commented on... 😅

arner (approved):
Yessss let's go! Nice one!

PR #140, opened 2026-05-01
Remove unused variables

ale-linux:


integration/test_helpers.go
```diff
@@ -504,15 +498,15 @@ func newEndorser(t *testing.T, logger sdk.Logger, cfg econf.Endorser, channel, n
 	}
 
 	end, err := endorser.New(
-		endorser.NewEVMEngine(namespace, writeDB, evmConfig, monotonicVersions),
+		endorser.NewEVMEngine(namespace, endorserDB, evmConfig, monotonicVersions),
```
arner, line 501: We might lose some throughput now that we don't have a read only db for the EVMEngine. SQLite does not have concurrent writes, so we were using a writedb with only one connection, while the readdb had 4 connections who can read concurrently. No problem for now but something to keep in mind, depending how the statedb evolves.

arner (approved):

integration/test_helpers.go
ale-linux: yeah I'm working on the light KVS now... let's see what that brings :)

PR #141, opened 2026-05-01
gateway/api/rpcerr: add typed JSON-RPC error package

Storm1289:
## Summary
First PR of a four-part plan to bring the gateway's JSON-RPC error responses to parity with geth, tracked in #46.

Adds a new leaf package `gateway/api/rpcerr` that exposes typed errors implementing go-ethereum's `rpc.Error` and `rpc.DataError` interfaces. This enables later PRs to return Ethereum-compatible JSON-RPC error codes from handlers and validation paths.

This PR is **purely additive** no existing call sites are modified and no behavior changes.

Refs #46.

## Background
Currently, all RPC handlers (`gateway/api/eth.go`) return plain Go `error`. The go-ethereum RPC server maps non-`rpc.Error` values to:

```json
{ "code": -32603, "message": "<msg>" }
```

This makes all failures appear as "Internal error", breaking compatibility with Ethereum clients that rely on error codes (ethers.js, Hardhat, MetaMask, Blockscout, etc.).

### Package structure
A separate `rpcerr` package is introduced to avoid dependency cycles:
- No internal dependencies
- Can be used by both `gateway/api` and `gateway/core`

### Types
- `Error` → implements `rpc.Error`
- `DataError` → implements `rpc.Error` + `rpc.DataError`

### Constructors
- `InvalidParams(...)` → -32602
- `Internal(err)` → -32603 (nil-safe)
- `TxRejected(err)` → -32003 (nil-safe)
- `ExecutionReverted(reason, data)` → -32000

### Notes
- `Internal` and `TxRejected` return `nil` if input is `nil`
- `InvalidParams` and `ExecutionReverted` always return non-nil
- Minimal API: 4 constants, 2 types, 4 functions

## Files

- `gateway/api/rpcerr/rpcerr.go`
- `gateway/api/rpcerr/rpcerr_test.go`

## Test Plan

- `go vet ./...` clean  
- `gofmt -l gateway/api/rpcerr/` clean  
- `go test -race -count=1 ./gateway/api/rpcerr/...` all 7 tests passing  

All tests passing.

## Follow-ups

Planned as separate PRs:

1. Wire validation errors in `SendRawTransaction`
2. Add `eth_call` revert handling
3. Add read-side coverage, integration tests, and docs (Closes #46)

Issue #46, opened 2026-03-26
Validate Ethereum Error Code Compatibility


ale-linux:
Validate that fabric-x-evm returns Ethereum-compatible error codes and error messages for all failure scenarios to ensure client applications can handle errors correctly. Ethereum clients expect specific error codes for different failure types such as insufficient gas, invalid nonce, contract execution revert, and transaction validation failures. Implement comprehensive error mapping from Fabric and EVM errors to standard Ethereum JSON-RPC error codes. Create tests that deliberately trigger various error conditions and verify the returned error codes and messages match Ethereum client expectations. Document the error code mappings and ensure error messages provide useful diagnostic information without exposing internal implementation details that could confuse users expecting standard Ethereum behavior.

---

**Quarter:** Q2

Storm1289:
Hi @ale-linux,I would like to work on this.

Storm1289:
Hi @ale-linux, @arner  here's a full implementation plan for #46. Posting before coding so we can align on scope and sequencing.

## Goal

Make the gateway's JSON-RPC error responses match the geth contract that Ethereum clients (ethers.js, web3.py, native ethclient, Metamask, etc.) expect. Today every failure is reported as `-32603 Internal Error`, because all handlers return bare `error` and the go-ethereum RPC server falls back to `-32603` for anything that does not implement `rpc.Error`.

## Non-overlap with #131

This work is fully independent of @rootp1's Task 1.1 on #131. The only file that *could* be touched by both is `gateway/core/api.go`, but #131 modifies `SendTransaction`'s enqueue path while #46 only changes how errors surface in `gateway/api/eth.go` handlers. No structural overlap, trivial rebase if both are in flight.

## Background current state

I surveyed the JSON-RPC surface. Findings:

- All 19 user-facing handlers in `gateway/api/eth.go` return bare `error`. go-ethereum's RPC server converts any non-`rpc.Error` into `-32603 Internal Error` regardless of cause.
- `gateway/core/validate.go` already uses geth's typed errors (`ErrNonceTooLow`, `ErrIntrinsicGas`, `ErrTxTypeNotSupported`, `ErrMaxInitCodeSizeExceeded`, `txpool.ErrInvalidSender`, `errUnprotectedTx`), but several are wrapped with `fmt.Errorf("...: %v", err)` which destroys type identity at the RPC boundary.
[...]

PR #142, opened 2026-05-01
gateway: return Ethereum-compatible JSON-RPC codes for tx submission errors

Storm1289:
## Summary
Second PR of the four-part plan on #46. Wires validation and decoding errors through the typed-error package added in #141, so Ethereum clients receive the **standard JSON-RPC codes** instead of `-32603 Internal Error` for every failure on the write path.

Refs #46.

## What changes for clients

| Failure | Before | After |
|--------|--------|-------|
| Send tx with `nonce too low` | `-32603 Internal Error` | `-32003 Transaction rejected` |
| Send tx below intrinsic gas | `-32603` | `-32003` |
| Send tx with insufficient funds | `-32603` | `-32003` |
| Send unsupported tx type (e.g. blob over RPC) | `-32603` | `-32003` |
| Send tx with init code over EIP-3860 limit | `-32603` | `-32003` |
| Send tx with bad signature / invalid sender | `-32603` | `-32003` |
| Send unprotected (non-EIP-155) tx | `-32603` | `-32003` |
| Send malformed raw tx (cannot RLP-decode) | `-32603` | `-32602 Invalid params` |
| `eth_call` bad hex (`gas`, `gasPrice`, `value`, `input`, `data`, `maxFeePerGas`, `maxPriorityFeePerGas`) | `-32603` | `-32602` |
| Internal state lookup failure | `-32603` | `-32603` (unchanged) |

Wallets, simulators, and dApps that branch on JSON-RPC codes (ethers.js `UNPREDICTABLE_GAS_LIMIT`, Hardhat, MetaMask, Blockscout, etc.) will now receive correct signals.

## Design

Per @arner's review on #46, **`gateway/api` does not depend on `gateway/core`**.

```
gateway/core ──> gateway/api/rpcerr
gateway/api  ──> gateway/api/rpcerr
gateway/api  ─X─> gateway/core
```

- Classification lives in `core`
- `SendTransaction` returns `rpc.Error`
- API propagates errors unchanged
- API directly uses `rpcerr.InvalidParams` only for:
  - raw tx decode
  - hex decoding in `argsToCallMsg`

## Files

| File | Change |
|------|--------|
| `gateway/core/rpcerrors.go` | New classifier (`classifyValidationError`) |
| `gateway/core/rpcerrors_test.go` | Tests for sentinel mapping and behavior |
| `gateway/core/validate.go` | Adds `errNonceLookup` + multi-`%w` wrapping |
| `gateway/core/api.go` | Applies classifier in `SendTransaction` |
| `gateway/api/eth.go` | Maps decode errors → `rpcerr.InvalidParams` |
| `gateway/api/eth_test.go` | Adds invalid payload + bad hex tests |

## errNonceLookup handling

Before:
```go
nonce, err := state.NonceAt(ctx, from, nil)
if err != nil {
    return fmt.Errorf("look up nonce: %w", err)
}
```

After:
```go
return fmt.Errorf("%w: %w", errNonceLookup, err)
```

- Enables `errors.Is(err, errNonceLookup)`
- Preserves original error chain
- Required for correct classification

## Test plan

- `go vet ./...` clean
- `gofmt -l ./...` clean
- `go test -race -count=1 ./gateway/...` all passing
- `go test -count=1 -run '^TestLocalX$' ./integration/...`  passing
- Existing validation tests (10) unchanged and passing


gateway/core/rpcerrors.go
```diff
@@ -0,0 +1,32 @@
...
+// and the unprotected-tx sentinel all flow through TxRejected because
+// they describe a tx the node refuses to admit, not a server fault.
+func classifyValidationError(err error) error {
+	if err == nil {
+		return nil
+	}
+	if errors.Is(err, errNonceLookup) {
+		return rpcerr.Internal(err)
+	}
+	return rpcerr.TxRejected(err)
```
arner: Just an idea, but I'm wondering if we can have even cleaner dependency graphs. It seems to me that the core package should not "know" anything about the presentation layer (RPC API, in this case). In clean architecture, it should be easy to replace an entire layer, and here the errors leak some implementation details across layers. What do you think of having an `errors.go` in the `domain` package, with the sentinel errors you defined in api.go, errUnprotectedTx and errNonceLookup, but export them? The translation you're doing in this rpcerrors.go could then move to the API package. I don't know which other changes you have planned, but with the previous and current PR this seems doable. Maybe a tiny bit more translation but cleaner separation. Maybe the rpcerr package can then be flattened into the api package.

Maybe I'm wrong, there might be some subtle issues with this approach, but could you give it some thought?

Storm1289: Good point ,the import direction in rpcerrors.go (core → api/rpcerr) is the wrong way around. I'll move the translation to the API layer. One question on placement: the sentinels (errUnprotectedTx, errNonceLookup) are validation concepts rather than domain types ,would you prefer them in gateway/domain/errors.go alongside Block/Transaction, or exported from gateway/core (next to the validator that emits them)? Either works for the layering; the second keeps domain focused on data shapes. Happy to go with whichever you prefer. Will also keep this pattern in mind for PR 3 (revert handling), which has the same core→api translation shape.

arner: API does not depend on core, so yes let's go for domain :)

Storm1289:
@arner, Done sentinels moved to gateway/domain/errors.go (ErrUnprotectedTx, ErrNonceLookup) and the classifier now lives in gateway/api/rpcerrors.go, called from SendRawTransaction. core no longer imports api/rpcerr; the dependency direction is now core → domain and api → domain/api/rpcerr. I'll keep this layering for PR 3 (revert handling) too likely a domain.ErrExecutionReverted carrying the revert data, translated to -32000 in the api layer.

arner (approved):
LGTM

ale-linux (approved):
Many thanks @Storm1289

Issue #145, opened 2026-05-03
Gateway: Hardening Fabric-X Transaction Status Compatibility Across Committer Versions


rootp1:
While implementing async mempool processing and retry behavior, I found a real compatibility gap in Fabric-X transaction status handling.

In some environments, the block `TRANSACTIONS_FILTER` status value observed for effectively committed transactions differs from what the current enum mapping expects. 
The Gateway currently includes a pragmatic compatibility mapping so runtime processing succeeds, but this is not a long-term contract and may break with other committer versions or future status changes.

We should make status interpretation explicit and test-covered.

### Proposed Work

1. Introduction of explicit status compatibility strategy:
version-based mapping, capability detection, or configurable status profile.
Removing/containing hardcoded pragmatic fallback once robust strategy is in place.
2. Adding startup validation:
detecting unsupported/unknown status semantics early and fail fast or warn loudly.
3. Improving observability:
metrics and structured logs for status-code distribution, unknown codes, and mapping path used.
4. Adding tests

### Acceptance Criteria:

1. Gateway interprets commit/invalid/retryable statuses correctly for all supported committer versions.
2. Unknown status semantics are surfaced clearly (metrics + actionable logs).
3. Compatibility behavior is covered by automated tests (including regression tests for previously failing status mappings).
4. Runtime no longer depends on ad-hoc fallback assumptions without visibility.

arner:
See https://github.com/hyperledger/fabric-x-evm/pull/146#issuecomment-4388507347, this does not appear to be an issue for fabric-evm.

PR #146, opened 2026-05-03
Hardened Fabric-X transaction status compatibility for async mempool processing

rootp1:
This PR fixes a runtime [compatibility issue](https://github.com/hyperledger/fabric-x-evm/issues/145) I hit in the Fabric-X path while validating async mempool behavior. 

In some committer environments, the transaction status values in block metadata did not line up with the status semantics the gateway was assuming. That mismatch caused transactions that Ire effectively committed to be interpreted as invalid, which then led to incorrect retry behavior and missing state updates in integration runs.

To address this, I introduced a Fabric-X compatibility synchronizer/parser in the gateway core and moved Fabric-X sync wiring to use it in both app runtime and integration harness setup. The parser now preserves per-transaction status from `TRANSACTIONS_FILTER` and applies a compatibility mapping that supports legacy and current committer encodings observed in our environments. This keeps transaction validity decisions stable across those versions and removes the runtime failure mode I Ire seeing.

I also made integration startup behavior safer by waiting for synchronizer readiness before worker processing begins in the harness, reducing race-related noise during startup. Alongside this, I added focused unit tests for status mapping and parser behavior so this compatibility path is locked in and regressions are easier to catch.

Validation was done with gateway/core tests plus full Fabric-X integration (`make test-x`) on fresh committer state. With this patch, the async mempool flow remains intact and the Fabric-X status handling is now robust for the currently supported committer variants.

### DIFF
--- a/gateway/app/app.go
+++ b/gateway/app/app.go
@@ -117,7 +117,7 @@ func buildApp(cfg config.Config, gwSigner sdk.Signer, logger sdk.Logger, endorse
 	case "fabric":
 		gwSync, err = nfab.NewSynchronizer(chain, cfg.Network.Channel, cfg.Gateway.Committer.ToPeerConf(), gwSigner, logger, chain)
 	case "fabric-x", "":
-		gwSync, err = nfabx.NewSynchronizer(chain, cfg.Network.Channel, cfg.Gateway.Committer.ToPeerConf(), gwSigner, logger, chain)
+		gwSync, err = core.NewFabricXSynchronizer(chain, cfg.Network.Channel, cfg.Gateway.Committer.ToPeerConf(), gwSigner, logger, chain)
 	default:
 		return nil, fmt.Errorf("unsupported protocol: %q", cfg.Network.Protocol)
 	}
--- a/gateway/core/fabricx_sync.go
+++ b/gateway/core/fabricx_sync.go
@@ -83,7 +83,9 @@ func (p *fabricXCompatParser) Parse(b *cb.Block) (blocks.Block, error) {
 func isFabricXCommittedStatus(status int) bool {
 	const legacyCommittedStatusV2 = 2
 	// Compatibility across committer status encodings:
-	// - legacy: COMMITTED=0
+	// - legacy: committed was encoded as numeric 0.
+	//   In newer protobuf enums, value 0 is named STATUS_UNSPECIFIED.
+	//   We intentionally treat 0 as committed for backward compatibility.
 	// - current: COMMITTED=1
 	// - some deployed test committers: COMMITTED=2
 	return status == int(committerpb.Status_STATUS_UNSPECIFIED) ||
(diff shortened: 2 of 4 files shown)


rootp1:
@arner, @ale-linux
Please have a look whenever you have a chance to

ale-linux:
Thanks for your PR @rootp1 ! A couple of Qs:

1. was the issue visible only on fabric-x? Is it easy to reproduce it?
2. shouldn't we just push the fix to the fabric-x synchroniser directly on fabric-x-sdk? @arner, wdyt?

rootp1:
> Thanks for your PR @rootp1 ! A couple of Qs:
> 
> 1. was the issue visible only on fabric-x? Is it easy to reproduce it?
> 2. shouldn't we just push the fix to the fabric-x synchroniser directly on fabric-x-sdk? @arner, wdyt?

Thanks for the review @ale-linux
1. Was the issue visible only on Fabric-X? Is it easy to reproduce?
Yes, this is Fabric-X specific. The issue is in how the Fabric-X block TRANSACTIONS_FILTER metadata encodes per-transaction commit status, this metadata path doesn't exist in the standard Fabric gateway path. 
It surfaced when running against a committer that encodes committed as 0 (`STATUS_UNSPECIFIED` in protobuf3) or 2 rather than the current COMMITTED=1, which caused the gateway to mark effectively committed transactions as invalid, triggering incorrect retries and missing state updates. It's reproducible with make test-x against a committer using a non-1 committed encoding, though with the current committer default (COMMITTED=1) you won't see it. 
The unit tests in fabricx_sync_test.go cover the affected status values directly.
2. Shouldn't we push the fix to the fabric-x-sdk directly?
That's a fair point, and I agree the long-term right place for this is in fabric-x-sdk's blocksfx.BlockParser itself 
since ideally the SDK parser would interpret `TRANSACTIONS_FILTER` status codes with the same compatibility awareness. This PR's compat parser (`fabricXCompatParser in gateway/core/fabricx_sync.go`) decorates the SDK's BlockParser as a near-term stopgap so we're unblocked right now. 

I'd suggest we:
- Merge this as-is so the gateway is resilient today (with tests locking in the mapping)
- Open a corresponding issue on fabric-x-sdk to move status interpretation into the SDK's parser natively
- Once the SDK ships the fix, we can remove the compat decorator here and delegate directly
@arner @ale-linux, does that approach sound reasonable to you? 
Happy to open the SDK side issue if we're aligned.

arner:
Hi, I doubt this. Looking at the protobufs https://github.com/hyperledger/fabric-x-common/blob/main/api/committerpb/status.proto committed = 1. If it has ever been different doesn't matter, we don't have to support older versions of the test committer - and we're about to move to a newer version anyway.

If there is a problem, it should be fixed in the SDK. Closing this PR, please open an issue on the SDK with steps to reproduce if you encounter it again.

Issue #65, opened 2026-03-26
Validate Full Ledger Replay Recovery


ale-linux:
Validate that the Gateway can be completely destroyed and recreated from scratch by replaying the entire Fabric ledger without any data loss or manual intervention. Implement and test a recovery procedure where a new Gateway instance starts with an empty database, connects to the Fabric network, and synchronizes the complete blockchain history from genesis to the current block. Verify that after synchronization completes, the Gateway has identical state to a Gateway that has been running continuously, including all blocks, transactions, logs, and indexes. Test this recovery process with ledgers of various sizes to ensure it scales appropriately. Document the recovery procedure and expected synchronization time. This capability is critical for disaster recovery and enables operational flexibility in managing Gateway deployments.

---

**Quarter:** Q3

PR #147, opened 2026-05-03
tests: add replay recovery parity test for persisted gateway state

Ayush3941:
## Summary

Adds recovery coverage for gateway persistence.

Verifies that gateway state can be:

- persisted to SQLite + trie storage
- reopened without drift
- rebuilt from committed ledger history with the same result

Refs #65.

## Test plan

- `gofmt -l .`
- `go vet $(go list ./... | grep -v '/integration')`
- `go tool staticcheck ./...`
- `go test -count=1 $(go list ./... | grep -v '/integration')`


Ayush3941:
This PR is intentionally limited to gateway persistence recovery parity from committed ledger history.
It does not fully close #65 yet. If maintainers like this direction, I can extend this PR with additional #65 coverage

ale-linux:
LGTM but I'd like to hear what @arner thinks

arner:
As discussed, sorry for letting this become stale @Ayush3941 , and thanks for the effort anyway!

PR #152, opened 2026-05-04
Use lightweight in-memory key-value store

ale-linux:
This commit introduces a new LightKVS implementation that provides versioned key-value storage with snapshot isolation semantics, replacing the previous database-backed state management approach.

The core change replaces direct database access with a snapshot-based model where each transaction execution operates on an isolated point-in-time view of the state. The new LightKVS maintains versioned state in memory and provides snapshot creation through a KVSSnapshotter interface, ensuring that concurrent transactions see consistent state without interference. This approach eliminates the need for external database connections during endorsement while maintaining proper MVCC semantics.

The EVMEngine has been updated to accept a KVSSnapshotter instead of a ReadStore, allowing it to create fresh snapshots for each transaction execution. The executor now properly manages snapshot lifecycle with explicit Close calls to prevent resource leaks. The synchronizer components in both Fabric and Fabric-X protocols have been adapted to work with the new storage abstraction, passing the same LightKVS instance for both read and write operations.

### DIFF
--- a/endorser/executor.go
+++ b/endorser/executor.go
@@ -18,7 +18,6 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus/misc/eip4844"
 	"github.com/ethereum/go-ethereum/core"
-	ethstate "github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/params"
@@ -49,9 +48,8 @@ type EVMEngine struct {
 	monotonicVersions bool
 
 	// LightKVS provides versioned storage with snapshot isolation
-	kvs        KVSSnapshotter
-	ethStateDB *ethstate.StateDB
-	evmConfig  EVMConfig
+	kvs       KVSSnapshotter
+	evmConfig EVMConfig
 }
 
 // NewEVMEngine creates a new EVMEngine.
@@ -61,18 +59,7 @@ func NewEVMEngine(namespace string, kvs KVSSnapshotter, evmConfig EVMConfig, mon
 		kvs:               kvs,
 		monotonicVersions: monotonicVersions,
 		evmConfig:         evmConfig,
-		ethStateDB:        nil, // Will be set when priming state
 	}
-}
-
-// SetEthStateDB sets the ethStateDB for the EVMEngine.
-func (e *EVMEngine) SetEthStateDB(ethStateDB *ethstate.StateDB) {
-	e.ethStateDB = ethStateDB
-}
-
-// GetEthStateDB returns the ethStateDB from the EVMEngine.
-func (e *EVMEngine) GetEthStateDB() *ethstate.StateDB {
-	return e.ethStateDB
 }
 
 // Execute runs a state-changing transaction and returns the EVM result,
@@ -86,7 +73,7 @@ func (e *EVMEngine) Execute(blockInfo *utils.BlockInfo, tx *types.Transaction) (
 	}
 	defer ex.Close()
 
-	ret, err := ex.send(tx)
+	ret, err := ex.Send(tx)
 	if err != nil {
 		return endorsement.ExecutionResult{}, err
 	}
@@ -115,7 +102,7 @@ func (e *EVMEngine) Call(msg ethereum.CallMsg, blockNumber *big.Int) ([]byte, er
 	}
 	defer ex.Close()
 
-	return ex.call(msg)
+	return ex.Call(msg)
 }
 
 func (e *EVMEngine) BalanceAt(_ context.Context, account common.Address, blockNumber *big.Int) (*big.Int, error) {
@@ -156,7 +143,7 @@ func (e *EVMEngine) NonceAt(_ context.Context, account common.Address, blockNumb
 
 // newExecutor creates a fresh executor with an isolated StateDB.
 // stateBlockNum selects the Fabric block height for the state snapshot (0 = latest).
-func (e *EVMEngine) newExecutor(blockInfo *utils.BlockInfo, stateBlockNum uint64) (*executor, error) {
+func (e *EVMEngine) newExecutor(blockInfo *utils.BlockInfo, stateBlockNum uint64) (*Executor, error) {
 	// Begin a new reader to get snapshot isolation
 	reader := e.kvs.NewSnapshot()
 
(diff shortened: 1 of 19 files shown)


Storm1289:
Hi @ale-linux, when you get a chance, could you please review the plan for phase 3 of issue #46. The first two phases are already completed.

endorser/executor.go
```diff
@@ -37,22 +37,29 @@ type EVMConfig struct {
...
+
+	// LightKVS provides versioned storage with snapshot isolation
+	kvs        KVSSnapshotter
+	ethStateDB *ethstate.StateDB
+	evmConfig  EVMConfig
 }
 
 // NewEVMEngine creates a new EVMEngine.
-func NewEVMEngine(namespace string, db ReadStore, evmConfig EVMConfig, monotonicVersions bool) *EVMEngine {
+// The db parameter should be a KVSSnapshotter to provide snapshot isolation.
```
arner: it is, isn't it? I mean, there is no db parameter, only a kvs snapshotter

endorser/executor.go
```diff
@@ -37,22 +37,29 @@ type EVMConfig struct {
...
 type EVMEngine struct {
 	namespace         string
 	monotonicVersions bool
-	db                ReadStore
-	ethStateDB        *ethstate.StateDB
-	evmConfig         EVMConfig
+
+	// LightKVS provides versioned storage with snapshot isolation
+	kvs        KVSSnapshotter
+	ethStateDB *ethstate.StateDB
```
arner: I forgot, is the ethStateDB only here for testing? Maybe we should write that down in a comment, our state story is a bit confusing 😅

endorser/executor.go
```diff
@@ -105,87 +114,104 @@ func (e *EVMEngine) Call(msg ethereum.CallMsg, blockNumber *big.Int) ([]byte, er
...
 	if evmConfig.ChainConfig == nil {
 		return nil, fmt.Errorf("evmConfig.ChainConfig must be set")
 	}
 	if blockInfo == nil {
-		// Note: stateDB.Version() is a Fabric block number, not an Ethereum block number — these are
-		// separate namespaces. With all forks active from block 0 this is harmless,
-		// but callers executing real transactions should always supply blockInfo explicitly.
 		blockInfo = &utils.BlockInfo{
-			BlockNumber: new(big.Int).SetUint64(stateDB.Version()),
+			BlockNumber: new(big.Int),
```
arner, line 200: With this the smart contract can't call `block.number`, or at least not get the right response. Can we inject the current (or at least the one of the simulation store) blockheight some other way?

endorser/lightkvs.go
```diff
@@ -0,0 +1,359 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+*/
+
+// Package endorser implements a lightweight versioned key-value store (LightKVS)
```
arner: I think package endorser does more than that, this description is about the LightKVS

endorser/lightkvs.go
```diff
@@ -0,0 +1,359 @@
...
+	kvs.current.Store(initial)
+	return kvs
+}
+
+// NewSnapshot starts a new read transaction and returns a Reader.
+// The Reader will see a consistent snapshot of the store at this point in time,
+// regardless of subsequent writes.
+//
+// Readers must call Close() when done to allow garbage collection of old snapshots.
+func (kvs *LightKVS) NewSnapshot() ReadStore {
```
arner, line 131: We're not actively using it at the time, but before this we supported historic queries. The snapsnot has an optional block number which allows the user to choose a moment in time to execute against. It's an ethereum feature.

While I don't mind temporarily losing this feature in favour of speed, it would be good if we can reinstate it if we move to pebble (or equivalent).

endorser/statedb.go
```diff
@@ -707,11 +698,6 @@ func (s *StateDB) Logs() []Log {
 	return s.logs
 }
 
-// Version returns the block height of this snapshot.
```
arner, line 710: See above, I'm not sure if we can just get rid of this, we need to know which block we're at... Unless there's another way, maybe it's a bit odd that we let the database tell the executor which block they're at, but still...

integration/test_helpers.go
```diff
@@ -474,11 +501,11 @@ func newEndorser(t *testing.T, cfg econf.Endorser, channel, namespace string, ev
...
 
-	endorserDB, err := state.NewWriteDB(channel, cfg.DbConnStr)
-	if err != nil {
-		t.Fatalf("NewWriteDB: %v", err)
-	}
-	t.Cleanup(func() { endorserDB.Close() })
+	// Create LightKVS for versioned key-value storage with snapshot isolation
+	lightKVS := endorser.NewLightKVS()
+
+	// Cleanup is a no-op for LightKVS
```
arner, line 431: Then why do it? 🤔

arner (commented):
There are some features of the VersionedDB not covered by the KVS: timetravel, persistence, and solidity's block.number. It's nice to see the VersionedDBWrapper, do you think it might be possible to make the evm engine constructor depend on an interface so it's still easy to plug in the VersionedDBWrapper instead of the KVS, in case we need any of these features?

endorser/executor.go
ale-linux: Yeah sorry, that was a comment line from one of the earlier drops - gone now because indeed, we are expecting an interface

arner:
> There are some features of the VersionedDB not covered by the KVS: timetravel, persistence, and solidity's block.number. It's nice to see the VersionedDBWrapper, do you think it might be possible to make the evm engine constructor depend on an interface so it's still easy to plug in the VersionedDBWrapper instead of the KVS, in case we need any of these features?

Oops never mind, it already relies on an interface!

endorser/executor.go
ale-linux: yes this was ugly - I ended up creating endorser/engine/excutor wrappers for our ethereum tests only, to keep our base implementations free of all that stuff. It looks a lot better now - thanks for challenging that

endorser/executor.go
ale-linux: ah yes, good point... how about we create an issue for that?

endorser/lightkvs.go
ale-linux: :) Bob got carried away

endorser/lightkvs.go
ale-linux: yup - as per our convo, let's go with this for now and create an issue to fix it?

integration/test_helpers.go
ale-linux: because pebble will need that and I'm shooting for a drop-in replacement

endorser/testimpl/executor_wrapper.go
```diff
@@ -0,0 +1,81 @@
+/*
```
arner, line 1: Ooh very nice!

arner (commented):
LGTM

arner (approved):

PR #154, opened 2026-05-05
add release targets for docker image

arner:
This PR will publish a multi-arch docker image when we push a version tag. Heavily inspired by https://github.com/hyperledger/fabric-x-committer/blob/main/.github/workflows/docker-release.yml.

Added a health check to the binary, so we can have an image `FROM scratch`.


ale-linux (approved):
Thanks @arner! We need to rebase it manually first tho

PR #155, opened 2026-05-05
gateway: surface eth_call reverts as JSON-RPC -32000 with revert data

Storm1289:
## Summary
Phase 3 of #46.

Closes the EVM revert gap on both the read and write paths so the gateway now matches geth’s JSON-RPC behavior end-to-end.

Before this PR:
- `eth_call` reverts returned `-32603 Internal Error` with no revert payload
- reverting transactions were dropped before commit and never produced receipts

After this PR:
- `eth_call` reverts return `-32000` with the raw revert payload in `error.data`
- reverting transactions are committed with `receipt.Status = 0`

Builds on:
- #141 (`gateway/api/rpcerr`)
- #142 (validation mapping)

Follows layering rule:
```text
api → domain only
(no api → core)
```

Refs #46.

---

### endorser/endorser.go
- `response()` now forwards revert payloads even on error
- revert responses use `Status: 201`
- genuine failures remain `500`

### gateway/domain/errors.go

Added:

```go
type RevertError struct {
    Reason string
    Data   []byte
}
```

Also adds:
- `ErrExecutionReverted`
- `errors.Is` / `errors.As` support

### gateway/core/endorse.go
- detects revert responses
- returns `*domain.RevertError`
- non-revert failures unchanged

### gateway/api/rpcerrors.go
- `RevertError` → `rpcerr.ExecutionReverted`
- other backend failures → `rpcerr.Internal`

### gateway/api/eth.go
- `Call()` now routes backend errors through the classifier

### Commit / receipt path
Reverting transactions are now:
- endorsed
- sequenced
- committed

Committed revert txs correctly surface as:

```go
receipt.Status = types.ReceiptStatusFailed
```

---

### Before

```json
{
  "code": -32603,
  "message": "query response was not successful, error code 500, msg execution reverted: ERC20: insufficient allowance"
}
```

### After

```json
{
  "code": -32000,
  "message": "execution reverted: ERC20: insufficient allowance",
  "data": "0x08c379a0..."
}
```

---

## Tests

10 new unit tests added:

- `gateway/core/endorse_test.go`
- `gateway/api/rpcerrors_test.go`
- `gateway/api/eth_test.go`

New integration subtest:
```text
testRevertHandling
```

Assertions:
- `eth_call` → `-32000` + revert data
- committed tx → `receipt.Status == types.ReceiptStatusFailed`

Runs under:
- `TestLocal`
- `TestLocalX`

---

## Docs

Updated:
- `docs/COMPATIBILITY.md`

---

## Test Plan

- `go build ./...`  clean
- `go test ./gateway/api/... ./gateway/core/... ./gateway/domain/... ./endorser/...`  all passing
- `go test -short ./...`  full module green

---

Refs #46

### DIFF
--- a/common/events.go
+++ b/common/events.go
@@ -49,3 +49,37 @@ func UnmarshalLogs(event []byte) ([]state.Log, error) {
 
 	return logs, nil
 }
+
+// EventNameRevert is the ChaincodeEvent name used to signal an EVM revert
+// from the endorser to the committer, so receipts can be marked status=0.
+const EventNameRevert = "revert"
+
+// MarshalRevert wraps the raw revert payload in a ChaincodeEvent with name
+// "revert" so the committer can detect the revert via the block-parsed events.
+func MarshalRevert(payload []byte, namespace, txID string) ([]byte, error) {
+	return proto.Marshal(&peer.ChaincodeEvent{
+		Payload:     payload,
+		ChaincodeId: namespace,
+		TxId:        txID,
+		EventName:   EventNameRevert,
+	})
+}
+
+// IsRevertEvent reports whether the given event bytes represent an EVM revert.
+// The SDK Endorse builder wraps res.Event in an outer ChaincodeEvent
+// (EventName "log"), so the marker we set in the inner event is one
+// proto-unmarshal deeper.
+func IsRevertEvent(event []byte) bool {
+	if len(event) == 0 {
+		return false
+	}
+	var outer peer.ChaincodeEvent
+	if err := proto.Unmarshal(event, &outer); err != nil {
+		return false
+	}
+	var inner peer.ChaincodeEvent
+	if err := proto.Unmarshal(outer.Payload, &inner); err != nil {
+		return false
+	}
+	return inner.EventName == EventNameRevert
+}
--- a/common/events.go
+++ b/common/events.go
@@ -8,6 +8,7 @@ package common
 
 import (
 	"encoding/json"
+	"strings"
 
 	"github.com/hyperledger/fabric-protos-go-apiv2/peer"
 	"github.com/hyperledger/fabric-x-sdk/state"
@@ -50,18 +51,20 @@ func UnmarshalLogs(event []byte) ([]state.Log, error) {
 	return logs, nil
 }
 
-// EventNameRevert is the ChaincodeEvent name used to signal an EVM revert
-// from the endorser to the committer, so receipts can be marked status=0.
-const EventNameRevert = "revert"
+// eventNameRevertPrefix is the prefix of the inner ChaincodeEvent name used
+// to signal an EVM revert. The Fabric txid is appended so the full name is
+// unique per transaction and cannot be forged by an EVM contract.
+const eventNameRevertPrefix = "revert:"
 
-// MarshalRevert wraps the raw revert payload in a ChaincodeEvent with name
-// "revert" so the committer can detect the revert via the block-parsed events.
+// MarshalRevert wraps the raw revert payload in a ChaincodeEvent whose name
+// is "revert:<txID>" so the committer can detect the revert and the marker
+// cannot collide with any name an EVM contract could produce.
 func MarshalRevert(payload []byte, namespace, txID string) ([]byte, error) {
 	return proto.Marshal(&peer.ChaincodeEvent{
 		Payload:     payload,
 		ChaincodeId: namespace,
 		TxId:        txID,
-		EventName:   EventNameRevert,
+		EventName:   eventNameRevertPrefix + txID,
 	})
 }
 
@@ -81,5 +84,5 @@ func IsRevertEvent(event []byte) bool {
 	if err := proto.Unmarshal(outer.Payload, &inner); err != nil {
 		return false
 	}
-	return inner.EventName == EventNameRevert
+	return strings.HasPrefix(inner.EventName, eventNameRevertPrefix)
 }
--- a/endorser/endorser.go
+++ b/endorser/endorser.go
@@ -17,6 +17,7 @@ import (
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/hyperledger/fabric-protos-go-apiv2/peer"
 	"github.com/hyperledger/fabric-x-evm/common"
 	"github.com/hyperledger/fabric-x-evm/utils"
@@ -65,29 +66,22 @@ func New(engine *EVMEngine, builder endorsement.Builder, chainID int64) (*Endors
 	}, nil
 }
 
-// ProcessEVMTransaction processes an Ethereum transaction and returns a signed proposal response
+// ProcessEVMTransaction processes an Ethereum transaction and returns a signed proposal response.
+// Reverts are endorsed and submitted (so the receipt records status=0); pre-execution
+// failures (nonce, gas, signer, EIP-3860, etc.) are rejected before any envelope is cut.
 func (f *Endorser) ProcessEVMTransaction(ctx context.Context, inv endorsement.Invocation, ethTx *types.Transaction, blockInfo *utils.BlockInfo) (*peer.ProposalResponse, error) {
-	// Validate the ethereum transaction signature
 	if _, err := types.Sender(f.ethSigner, ethTx); err != nil {
 		return nil, fmt.Errorf("invalid ethereum signature: %w", err)
 	}
 
-	// Execute the transaction
 	res, err := f.Engine.Execute(blockInfo, ethTx)
 	if err != nil {
-		// Distinguish between pre-execution validation errors and execution errors.
-		// Pre-execution errors (from ApplyMessage) indicate the transaction is invalid
-		// and should be rejected. Execution errors (from result.Err) indicate the
-		// transaction executed but failed, and should be included in the response.
 		if isPreExecutionError(err) {
-			// Pre-execution validation error: reject the transaction
 			return nil, err
 		}
-		// Execution error: include in response with error status
 		return response(nil, err), nil
 	}
 
-	// Build and sign the endorsement
 	return f.builder.Endorse(inv, res)
 }
 
@@ -135,12 +129,16 @@ func (f *Endorser) ProcessStateQuery(ctx context.Context, query common.StateQuer
 
 func response(res []byte, err error) *peer.ProposalResponse {
 	if err != nil {
-		// Carry res through on error so callers receive the EVM revert payload
-		// for ErrExecutionReverted; for non-revert errors res is empty.
+		// 201 marks an EVM revert: success-range so protoutil cuts a tx, but
+		// distinguishable from 200 so the gateway and committer can tell.
+		status := int32(500)
+		if errors.Is(err, vm.ErrExecutionReverted) {
+			status = 201
+		}
 		return &peer.ProposalResponse{
 			Version: 1,
 			Response: &peer.Response{
-				Status:  500,
+				Status:  status,
 				Message: err.Error(),
 				Payload: res,
 			},
--- a/endorser/endorser.go
+++ b/endorser/endorser.go
@@ -135,11 +135,14 @@ func (f *Endorser) ProcessStateQuery(ctx context.Context, query common.StateQuer
 
 func response(res []byte, err error) *peer.ProposalResponse {
 	if err != nil {
+		// Carry res through on error so callers receive the EVM revert payload
+		// for ErrExecutionReverted; for non-revert errors res is empty.
 		return &peer.ProposalResponse{
 			Version: 1,
 			Response: &peer.Response{
 				Status:  500,
 				Message: err.Error(),
+				Payload: res,
 			},
 		}
 	}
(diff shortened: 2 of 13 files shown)


Storm1289:
@arner @ale-linux, when you get a chance, could you please take a look at this pr?

arner (approved):
Nice one!

ale-linux (changes requested):
I'm currently unable to leave inline comments becaue of a [github outage](https://www.githubstatus.com/incidents/f5pb5d5mr9yh). However, here's a couple of comments:

1. we don't need to care about `CallContract`, it's meant to be a read-only call anyway
2. we should instead focus on `ExecuteTransaction` which is the code path taken for sent transactions
3. this PR needs integration tests - unit tests are not sufficient. I tried this quick-and-dirty

```
diff --git a/integration/integration_test.go b/integration/integration_test.go
index b07cf1f..027801b 100644
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -314,6 +314,8 @@ func testTetherToken(t *testing.T, th *TestHarness) {
 
        // Deploy contract
        addr := deploySmartContract(t, node1, ethOwner, deploySupply, tokenName, tokenSymbol, tokenDecimals)
+       approveAmount := big.NewInt(100_000)
+       callSmartContract(t, ethA, addr, node2, "transferFrom", nil, ethOwner.Address(), ethA.Address(), approveAmount)
 
        // -------------------- full test body --------------------
 
@@ -333,7 +335,7 @@ func testTetherToken(t *testing.T, th *TestHarness) {
        querySmartContractExpect(t, ethOwner, addr, th, toA, "balanceOf", ethA.Address())
 
        // -------------------- Owner approves user2 --------------------
-       approveAmount := big.NewInt(100_000)
+       approveAmount = big.NewInt(100_000)
        callSmartContract(t, ethOwner, addr, node2, "approve", nil, ethA.Address(), approveAmount)
        querySmartContractExpect(t, ethOwner, addr, th, approveAmount, "allowance", ethOwner.Address(), ethA.Address())
 ```

   which should cause a revert. But the transaction is never actually even submitted because the code in the fabric-x-sdk refuses to accept it (I tried on Local/Fabric) - @arner: we hit [this](https://github.com/hyperledger/fabric-x-common/blob/main/protoutil/txutils.go#L274) error.

At this point I suggest the following route:
1. we need to change the return code from a reverted transaction. If we pick an HTTP error code then the protoutil will refuse to cut a tx
2. `ProcessEVMTransaction` needs to be changes:

```go
// ProcessEVMTransaction processes an Ethereum transaction and returns a signed proposal response
func (f *Endorser) ProcessEVMTransaction(ctx context.Context, inv endorsement.Invocation, ethTx *types.Transaction, blockInfo *utils.BlockInfo) (*peer.ProposalResponse, error) {
	// Validate the ethereum transaction signature
	if _, err := types.Sender(f.ethSigner, ethTx); err != nil {
		return nil, fmt.Errorf("invalid ethereum signature: %w", err)
	}

	// Execute the transaction
	res, err := f.Engine.Execute(blockInfo, ethTx)
	if err != nil {
		// Distinguish between pre-execution validation errors and execution errors.
		// Pre-execution errors (from ApplyMessage) indicate the transaction is invalid
		// and should be rejected. Execution errors (from result.Err) indicate the
		// transaction executed but failed, and should be included in the response.
		if isPreExecutionError(err) {
			// Pre-execution validation error: reject the transaction
			return nil, err
		}
		// Execution error: include in response with error status
		return response(nil, err), nil
	}

	// Build and sign the endorsement
	return f.builder.Endorse(inv, res)
}
```

   If it gets an error from `f.Engine.Execute`, and that error is a revert, it should not error out early and not execute `f.builder.Endorse(inv, res)`.

ale-linux:
I made a little more progress (it's still very hacky though...)

```
diff --git a/endorser/endorser.go b/endorser/endorser.go
index 17c3b22..5a90276 100644
--- a/endorser/endorser.go
+++ b/endorser/endorser.go
@@ -12,6 +12,7 @@ import (
        "fmt"
        "math/big"
        "regexp"
+       "strings"
 
        "github.com/ethereum/go-ethereum"
        ethcommon "github.com/ethereum/go-ethereum/common"
@@ -19,6 +20,7 @@ import (
        "github.com/ethereum/go-ethereum/core/types"
        "github.com/hyperledger/fabric-protos-go-apiv2/peer"
        "github.com/hyperledger/fabric-x-evm/common"
+       "github.com/hyperledger/fabric-x-evm/gateway/domain"
        "github.com/hyperledger/fabric-x-evm/utils"
        "github.com/hyperledger/fabric-x-sdk/endorsement"
 )
@@ -74,21 +76,21 @@ func (f *Endorser) ProcessEVMTransaction(ctx context.Context, inv endorsement.In
 
        // Execute the transaction
        res, err := f.Engine.Execute(blockInfo, ethTx)
-       if err != nil {
+       switch {
+       case err == nil || strings.HasPrefix(err.Error(), domain.ErrExecutionReverted.Error()):
+               // Success
+               return f.builder.Endorse(inv, res)
+
+       case isPreExecutionError(err):
                // Distinguish between pre-execution validation errors and execution errors.
                // Pre-execution errors (from ApplyMessage) indicate the transaction is invalid
                // and should be rejected. Execution errors (from result.Err) indicate the
                // transaction executed but failed, and should be included in the response.
-               if isPreExecutionError(err) {
-                       // Pre-execution validation error: reject the transaction
-                       return nil, err
-               }
-               // Execution error: include in response with error status
-               return response(nil, err), nil
-       }
+               return nil, err
 
-       // Build and sign the endorsement
-       return f.builder.Endorse(inv, res)
+       default:
+               panic("when does this happen?")
+       }
 }
 
 // ProcessCall processes an Ethereum call (query) and returns a proposal response
@@ -140,7 +142,7 @@ func response(res []byte, err error) *peer.ProposalResponse {
                return &peer.ProposalResponse{
                        Version: 1,
                        Response: &peer.Response{
-                               Status:  500,
+                               Status:  201,
                                Message: err.Error(),
                                Payload: res,
                        },
diff --git a/endorser/executor.go b/endorser/executor.go
index 212385f..65b2dc8 100644
--- a/endorser/executor.go
+++ b/endorser/executor.go
@@ -12,6 +12,7 @@ import (
        "errors"
        "fmt"
        "math/big"
+       "strings"
 
        "github.com/ethereum/go-ethereum"
        "github.com/ethereum/go-ethereum/accounts/abi"
@@ -21,6 +22,7 @@ import (
        "github.com/ethereum/go-ethereum/core/types"
        "github.com/ethereum/go-ethereum/core/vm"
        "github.com/ethereum/go-ethereum/params"
+       "github.com/hyperledger/fabric-x-evm/gateway/domain"
        "github.com/hyperledger/fabric-x-evm/utils"
        "github.com/hyperledger/fabric-x-sdk/endorsement"
 )
@@ -74,7 +76,7 @@ func (e *EVMEngine) Execute(blockInfo *utils.BlockInfo, tx *types.Transaction) (
        defer ex.Close()
 
        ret, err := ex.Send(tx)
-       if err != nil {
+       if err != nil && !strings.HasPrefix(err.Error(), domain.ErrExecutionReverted.Error()) {
                return endorsement.ExecutionResult{}, err
        }
        var logs []byte
diff --git a/integration/integration_test.go b/integration/integration_test.go
index b07cf1f..ff1d663 100644
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -314,6 +314,33 @@ func testTetherToken(t *testing.T, th *TestHarness) {
 
        // Deploy contract
        addr := deploySmartContract(t, node1, ethOwner, deploySupply, tokenName, tokenSymbol, tokenDecimals)
+       approveAmount := big.NewInt(100_000)
+       callSmartContract(t, ethA, addr, node2, "transferFrom", nil, ethOwner.Address(), ethA.Address(), approveAmount)
+
+       node := th.Gateways[0]
+       ec, err := NewNativeEthClient(node)
+       if err != nil {
+               t.Fatal(err)
+       }
+
+       tx, err := ethA.TxForCall(t.Context(), node2, &addr, "transferFrom", nil, ethOwner.Address(), ethA.Address(), approveAmount)
+       if err != nil {
+               t.Fatal(err)
+       }
+
+       err = ec.SendTransaction(t.Context(), tx)
+       if err != nil {
+               t.Fatal(err)
+       }
+
+       waitForCommitT(t, ec, tx)
+
+       r, err := ec.TransactionReceipt(t.Context(), tx.Hash())
+       if err != nil {
+               t.Fatal(err)
+       }
+
+       fmt.Println(r.Status)
 
        // -------------------- full test body --------------------
 
@@ -333,7 +360,7 @@ func testTetherToken(t *testing.T, th *TestHarness) {
        querySmartContractExpect(t, ethOwner, addr, th, toA, "balanceOf", ethA.Address())
 
        // -------------------- Owner approves user2 --------------------
-       approveAmount := big.NewInt(100_000)
+       approveAmount = big.NewInt(100_000)
        callSmartContract(t, ethOwner, addr, node2, "approve", nil, ethA.Address(), approveAmount)
        querySmartContractExpect(t, ethOwner, addr, th, approveAmount, "allowance", ethOwner.Address(), ethA.Address())

```

with this patch I can see the revert (the `callSmartContract` from an account that hasn't been approved yet at least committing. Now of course `r.Status` is successful because we can't tell that it's invalid... We need a way to signal that it is not valid. The committers definitely won't be able to determine it's invalid and so we'll have to parse the status from the endorsers (which we should set to a specific HTTP code which is "successful" so as not to break the tx packager, and yet not 200, so that we can tell "ah, it's a revert".

Storm1289:
Hi @ale-linux, thanks for the patch the committer-side gap you pointed out makes the full flow much clearer now.

Reading through it, my understanding is that the remaining piece is on the gateway side: when chain.go ingests a committed block, it needs to read the endorser Status from the proposal response and map reverted executions (201) to ethStatus=0 instead of the normal success path (200 -> ethStatus=1). That’s what makes the receipt come back as reverted in the integration test you added.

I’m happy to take this either way:

 (a) land all four pieces in this PR your three endorser patches + the `gateway/core/chain.go` `ethStatus` ingestion + the existing `eth_call -> -32000` mapping + the `testTetherToken` revert regression subtest

 (b) keep the current PR scoped to your patches + the `eth_call` mapping, and land the committer-side `ethStatus` ingestion as a small follow-up PR

I lean slightly toward (b) for reviewability and to stay closer to the original Phase 3 sequencing, but I’m also fine with (a) if you’d prefer a single self-contained flow.

ale-linux:
Thanks again @Storm1289. As @arner pointed out to me

> we don't need to care about CallContract, it's meant to be a read-only call anyway

this assertion is incorrect: we _also_ care because reverted calls must be surfaced correctly.

And yes, as I was saying, we should also handle reverts on send correctly. Given how close the code paths are, I'd like to see this in one single PR (broken down in different commits if you prefer), because I'd like to evaluate them together, given their similarities.

Also, I'd like to see integration tests driven directly by requests to the gw API through `NewNativeEthClient` - it should be easy to trigger them both on an ERC-20 smart contract as I proposed in my patch.

Storm1289:
Hi @ale-linux, pushed the write-path work as a second commit on top of the existing eth_call mapping both paths are now covered. Could you take another look when you get a chance .

endorser/endorser.go
```diff
@@ -65,29 +66,22 @@ func New(engine *EVMEngine, builder endorsement.Builder, chainID int64) (*Endors
 	}, nil
 }
 
-// ProcessEVMTransaction processes an Ethereum transaction and returns a signed proposal response
+// ProcessEVMTransaction processes an Ethereum transaction and returns a signed proposal response.
+// Reverts are endorsed and submitted (so the receipt records status=0); pre-execution
+// failures (nonce, gas, signer, EIP-3860, etc.) are rejected before any envelope is cut.
 func (f *Endorser) ProcessEVMTransaction(ctx context.Context, inv endorsement.Invocation, ethTx *types.Transaction, blockInfo *utils.BlockInfo) (*peer.ProposalResponse, error) {
-	// Validate the ethereum transaction signature
```
ale-linux, line 70: Could we please keep the inline comments?

endorser/executor.go
```diff
@@ -74,9 +76,23 @@ func (e *EVMEngine) Execute(blockInfo *utils.BlockInfo, tx *types.Transaction) (
 	defer ex.Close()
 
 	ret, err := ex.Send(tx)
-	if err != nil {
+	if err != nil && !errors.Is(err, vm.ErrExecutionReverted) {
```
ale-linux, line 79: @arner @Storm1289 : do we know under which circumstances we could hit these conditions? It would be great to have an integration test to verify that

common/events.go
```diff
@@ -49,3 +49,37 @@ func UnmarshalLogs(event []byte) ([]state.Log, error) {
 
 	return logs, nil
 }
+
+// EventNameRevert is the ChaincodeEvent name used to signal an EVM revert
+// from the endorser to the committer, so receipts can be marked status=0.
+const EventNameRevert = "revert"
```
ale-linux: Is it possible that a transaction will emit exactly this event for its own reasons? If so, we could just prepend or append the fabric txid to the event name - that will make it unique and very likely impossible to have been generated by the ethereum smart contract

ale-linux (commented):
Fantastic work @Storm1289, this is exactly what I had in mind 👍 

I left you a last bunch of annoying comments and also a Q for @arner - I think we could merge soon! Thanks!

endorser/executor.go
Storm1289: @ale-linux @arner non-revert errors that reach this branch are the EVM execution failures other than revert; out-of-gas and invalid-opcode are the main ones. Pre-execution paths (signer, nonce, intrinsic gas) are filtered upstream by isPreExecutionError before we get here.
Should the integration test cover both out-of-gas and invalid-opcode, or would one be enough?

endorser/endorser.go
Storm1289: Good catch ,will restore them in the next commit.

common/events.go
Storm1289: Today there's no path for an EVM contract to set the inner EventName contracts only produce Ethereum logs (LOG0–LOG4), which we serialize as JSON, and the SDK builder hardcodes the outer event name to "log". So no collision risk in practice.
But the txid prefix is good defense-in-depth will apply in the next commit.

ale-linux (approved):
Fantastic work @Storm1289, thanks a lot!

PR #157, opened 2026-05-05
Run hardhat tests on CI

ale-linux:


kushnireyal:
reopened here #159 because of some mess I created (initially thought it will be quick and easy :) )

PR #161, opened 2026-05-06
Use yaml files to configure the application instead of hardcoded values

arner:
Using viper, like other fabric-x components.


endorser/versioned_db_wrapper.go
```diff
@@ -70,3 +70,23 @@ func (s *VersionedDBSnapshot) Get(namespace, key string) (*blocks.WriteRecord, e
...
+
+// BlockNumber returns the last processed block number.
+func (w *VersionedDBWrapper) BlockNumber(ctx context.Context) (uint64, error) {
+	return w.db.BlockNumber(ctx)
+}
+
+// Close closes the underlying VersionedDB.
+func (w *VersionedDBWrapper) Close() error {
+	return w.db.Close()
+}
```
ale-linux, line 92: Why? Doesn't this code just duplicate the underlying API with pass-through stuff?

ale-linux (approved):
LGTM

Issue #66, opened 2026-03-26
Ensure Ledger-Derived State Only


ale-linux:
Verify that the Gateway does not maintain any authoritative data that cannot be reconstructed from the Fabric ledger, eliminating the need for separate backup and recovery procedures for Gateway state. All data in the Gateway's SQLite database should be derivable from the ledger through deterministic processing of blocks and transactions. Audit the database schema to identify any data that is not ledger-derived and either eliminate it or ensure it can be safely lost and reconstructed. Document which data is ephemeral versus durable and ensure operators understand that the Gateway database is a cache that can be rebuilt from the ledger at any time. This design principle simplifies operations by reducing the number of systems that require backup and ensures the ledger remains the single source of truth.

---

**Quarter:** Q3

arner:
Closing, this is more a design principle than a task.

PR #162, opened 2026-05-06
tests: prove query state is ledger-derived and rebuildable

Ayush3941:
## Summary

  Adds ledger-derived-state coverage for gateway persistence.

  Verifies that gateway state is rebuildable from committed ledger history by checking that:

  - the gateway query store contains only ledger-derived user tables
  - the same ledger-derived block data reproduces the same query results in a fresh store
  - local gateway state can be discarded and rebuilt from committed ledger history with the same result

  Refs #66.

  ## Test plan

  - `gofmt -l .`
  - `go vet $(go list ./... | grep -v '/integration')`
  - `go tool staticcheck ./...`
  - `go test -count=1 $(go list ./... | grep -v '/integration')`

### DIFF
--- a/gateway/core/chain_ledger_derived_state_test.go
+++ b/gateway/core/chain_ledger_derived_state_test.go
@@ -1,0 +1,233 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+*/
+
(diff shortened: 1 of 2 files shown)


arner:
As discussed, sorry for letting this become stale @Ayush3941 , and thanks for the effort anyway!

PR #163, opened 2026-05-07
Port performance testing infrastructure from private repo

kushnireyal:
This ports the `perf` branch from the private IBM repo (`github.ibm.com/decentralized-trust-research/fabric-evm`) to this repo. The branch adds performance testing infrastructure for replaying real USDC transactions from Ethereum mainnet.

Note: the `withTrie` changes that were also in `perf` are already here, so this only adds what's missing.

**What's new**

The main addition is balance priming — a way to run tests against real transaction data without having to pre-fund thousands of accounts. `endorser/statedb_wrapper.go` wraps the StateDB and intercepts ERC-20 balance reads, returning a high balance (1B tokens) instead of zero. This is wired into the executor via a new optional `BalancePriming` field on `EVMConfig`. When not set, there's zero overhead — it's just a nil check.

The executor integration uses a `SenderAware` interface rather than a concrete type assertion, so it's not coupled to this specific wrapper.

On top of that, there's the actual test infrastructure under `integration/perf/`: a test harness that replays pre-generated USDC transactions (`replay_json_dataset_test.go`), a fetcher tool to pull contract code from mainnet and generate the transaction dataset, and some Python scripts for visualizing results.

**Dataset files**

The two large dataset files (`dataset.tsv.gz` ~10MB, `USDC_dataset.json.gz` ~27MB) are not committed. See `integration/perf/USDC_testing.md` for how to generate them locally using `fetcher.go`.

**Running the perf tests**

Gated behind a build tag so they don't run in CI:

```bash
go test -tags=perf ./integration/perf/...
```

**Existing tests**

All pass — `go test ./endorser/... -short` and `go build ./...` are clean.


endorser/executor.go
```diff
@@ -422,6 +463,12 @@ func (h *Executor) Execute(msg *core.Message) ([]byte, error) {
 	// passes preCheck and executes.
 	gp := new(core.GasPool).AddGas(h.blockCtx.GasLimit)
 
+	// If the state implements SenderAware, inform it of the transaction sender.
+	// This allows wrappers like BalancePrimingWrapper to perform sender-specific optimizations.
+	if sa, ok := h.state.(SenderAware); ok {
+		sa.SetSender(msg.From)
+	}
+
```
ale-linux, line 471: Why do we also have to do it here?

ale-linux (changes requested):
LGTM but I wrote most of this code so I shouldn't review it.

I only want to flag an issue with DCO - I should also co-sign those commits as per LFDT DCO requirements. Here's how we can handle it: I created https://github.com/hyperledger/fabric-x-evm/tree/perf with the stuff I wrote, @kushnireyal you can just push a delta with the changes you made. 

At this point @arner must be the one reviewing the code, and when we squash-merge, the commit will have DCO signatures from both.

endorser/executor.go
kushnireyal: Right, it's redundant, will remove

PR #165, opened 2026-05-08
Add block-number-based snapshot retrieval to LightKVS

ale-linux:
Modified the NewSnapshot method in LightKVS to accept a blockNumber parameter, enabling retrieval of snapshots at specific block heights. The implementation searches through the current snapshot and a configurable history buffer to find matching block numbers. When blockNumber is 0, it returns the current snapshot for backward compatibility. If a requested block number is higher than the current block, it returns the current snapshot. The history buffer uses a ring buffer pattern to maintain the most recent old snapshots.

This enables 6 more hardhat tests that required this feature.

### DIFF
--- a/gateway/storage/store.go
+++ b/gateway/storage/store.go
@@ -12,6 +12,7 @@ import (
 	_ "embed"
 	"errors"
 	"strings"
+	"sync/atomic"
 
 	"github.com/hyperledger/fabric-x-evm/gateway/domain"
 )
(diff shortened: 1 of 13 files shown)


gateway/storage/store.go
```diff
@@ -20,8 +21,9 @@ import (
 var ddl string
 
 type Store struct {
-	queries *Queries
-	db      *sql.DB
+	queries     *Queries
+	db          *sql.DB
+	blockNumber atomic.Uint64 // cached block number for fast reads
```
arner, line 26: Nice!

arner (approved):
Looks good

PR #167, opened 2026-05-11
Enable full OpenZeppelin Hardhat test suite compatibility

ale-linux:


endorser/app/factory.go
```diff
@@ -25,28 +26,37 @@ import (
...
 	case "sqlite":
 		writeDB, err := state.NewWriteDB(network.Channel, cfg.Database.ConnString)
 		if err != nil {
-			return nil, nil, fmt.Errorf("failed to initialize store: %w", err)
+			return nil, nil, nil, fmt.Errorf("failed to initialize store: %w", err)
 		}
 		kvs = endorser.NewVersionedDBWrapper(writeDB)
 	default:
-		kvs = endorser.NewLightKVS()
+		kvs = endorser.NewLightKVS(2)
```
kushnireyal: Redundant set, since immediately overwritten

endorser/executor.go
```diff
@@ -446,6 +464,7 @@ func (h *Executor) Execute(msg *core.Message) ([]byte, error) {
 	// Note: result.Err contains execution errors (e.g., revert, out of gas, code size exceeded)
 	// These are not fatal errors - the transaction is included but failed
 	if result.Err != nil {
+		fmt.Println("========+> apply failed:", result.Err)
```
kushnireyal: Debug prints left on purpose?

endorser/executor.go
```diff
@@ -461,5 +480,6 @@ func formatRevert(ret []byte, err error) error {
 	if errUnpack != nil {
 		return err
 	}
+	fmt.Println("===> revert reason", fmt.Errorf("%w: %v", vm.ErrExecutionReverted, reason))
```
kushnireyal: Same- debug prints left on purpose?

gateway/core/chain.go
```diff
@@ -127,7 +129,7 @@ func (c *Chain) convertToDomain(b blocks.Block) domain.Block {
 
 		etx, err := convertTransaction(tx.InputArgs[1], b.Hash, b.Number, tx.Number, tx.ID, status, tx.Status, tx.Events, &logIndex)
 		if err != nil {
-			continue // ?
+			panic(err) // we surface this for now instead of swallowing it
```
kushnireyal, line 132: Just wondering, why?

gateway/app/app.go
```diff
@@ -63,21 +64,31 @@ func NewWithSigner(cfg config.Config, gwSigner sdk.Signer) (*App, error) {
 	// Create endorsers and their synchronizers.
 	endorsers := make([]core.Endorser, 0, len(cfg.Endorsers))
 	endorserSyncs := make([]*network.Synchronizer, 0, len(cfg.Endorsers))
+	var firstKVS interface{} // Keep first endorser's KVS for test server
 	for i, ecfg := range cfg.Endorsers {
-		end, sync, err := eapp.NewEndorser(ecfg, cfg.Network, logger, false)
+		// Set history size: always 128 for test RPC (snapshot/revert), else default to 2 if not set
```
kushnireyal, line 69: Why 128?

endorser/statedb.go
```diff
@@ -293,18 +293,22 @@ func (s *StateDB) getStateFromStore(key string) ([]byte, error) {
...
-			return nil, nil
-		}
-		if s.monotonicVersions {
-			read.Version = &blocks.Version{BlockNum: record.Version}
-		} else {
-			read.Version = &blocks.Version{
-				BlockNum: record.BlockNum,
-				TxNum:    record.TxNum,
+		// Use the value from the record, even if IsDelete is true
+		// (IsDelete: true with non-nil Value happens during snapshot revert)
```
kushnireyal, line 297: Is this test-only scenario? If so, maybe move somehow to "testimpl" or something like that?

common/proposal.go
```diff
@@ -15,7 +15,7 @@ import (
 type ProposalType byte
 
 const (
-	ProposalTypeEVMTx ProposalType = iota
+	ProposalTypeEVMTx ProposalType = 0xfb
```
kushnireyal, line 18: Why is that?

endorser/testimpl/lightkvs.go
```diff
@@ -0,0 +1,314 @@
...
+	availableBlocks := make([]uint64, 0, count+1)
+	for i := 0; i < count; i++ {
+		snapshot := kvs.History[i].Load()
+		if snapshot != nil {
+			availableBlocks = append(availableBlocks, snapshot.BlockNumber)
+		}
+	}
+	availableBlocks = append(availableBlocks, current.BlockNumber)
+
+	log.Printf("[DEBUG] LightKVSExt.NewSnapshot() called: requested=%d current=%d historyCount=%d available=%v\n",
```
kushnireyal: Consider use flogging

endorser/statedb_logger.go
kushnireyal, line 1: I see it's only use (in `endorser/executor.go`) is commented out. Is this file needed?

endorser/executor.go
```diff
@@ -170,16 +166,28 @@ func (e *EVMEngine) NonceAt(_ context.Context, account common.Address, blockNumb
...
 	// Create StateDB with the reader
 	stateDB, err := NewStateDB(context.TODO(), reader, e.namespace, stateBlockNum, e.monotonicVersions)
 	if err != nil {
 		reader.Close()
 		return nil, err
 	}
+
+	// in case logging is required - consider enabling programmatically
+	// sdl := NewStateDBLogger(stateDB)
+	// flogging.ActivateSpec("DEBUG")
```
kushnireyal, line 190: Is this needed?

gateway/core/chain.go
kushnireyal: Maybe a log + continue is better?

endorser/testimpl/lightkvs.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	newSnapshot := &endorser.Snapshot{
+		BlockNumber: blockNum,
+		Data:        newData,
+	}
+
+	// Sequential storage: append to history without wrapping
+	count := int(kvs.NextIndex.Load())
+	if count >= len(kvs.History) {
+		panic(fmt.Sprintf("snapshot history exhausted at block %d", blockNum))
```
kushnireyal, line 155: Does the kvs live for the entire test suite run? if so `count >= len(kvs.History)` (while `kvs.History=128`) might not be enough i think

endorser/lightkvs.go
kushnireyal, line 1: Are we sure we want to export all the previously unexported (lowercase) fields here?

endorser/app/factory.go
```diff
@@ -25,28 +26,37 @@ import (
...
 // NewEndorser creates a single endorser instance with its synchronizer.
 // This is the canonical way to create an endorser, whether embedded or standalone.
+// Returns the endorser, synchronizer, and the LightKVS instance (or extended version) for state management.
 func NewEndorser(
 	cfg config.Endorser,
 	network common.Network,
 	logger sdk.Logger,
 	skipAllNonceChecks bool,
-) (*endorser.Endorser, *sdknet.Synchronizer, error) {
+	testImpl bool,
```
kushnireyal, line 35: Consider separate test factory, something like `NewEndorserForTesting(...)`, that lives in "testimpl"

endorser/versioned_db_wrapper.go
kushnireyal, line 1: All RPC handlers now have debug log statements added at entry and exit. Is this intentional or something that was added for debugging during the work?

gateway/storage/store_test.go
```diff
@@ -378,7 +378,8 @@ func TestBlockNumber(t *testing.T) {
 	store := setupTestDB(t)
 
 	// Empty database should return 0
-	num, err := store.BlockNumber(t.Context())
+	num := store.CachedBlockNumber.Load()
+	err := error(nil)
 	if err != nil {
 		t.Fatalf("BlockNumber error: %v", err)
 	}
```
kushnireyal: Redundant

gateway/storage/store.go
```diff
@@ -20,14 +21,15 @@ import (
 var ddl string
 
 type Store struct {
-	queries *Queries
-	db      *sql.DB
+	queries           *Queries
+	DB                *sql.DB
+	CachedBlockNumber atomic.Uint64 // cached block number for fast reads
```
kushnireyal, line 26: Maybe should be private?

gateway/storage/store_test.go
```diff
@@ -378,7 +378,8 @@ func TestBlockNumber(t *testing.T) {
 	store := setupTestDB(t)
 
 	// Empty database should return 0
-	num, err := store.BlockNumber(t.Context())
+	num := store.CachedBlockNumber.Load()
```
kushnireyal, line 381: Tests should probably use the "regular user API" (e.g., `BlockNumber`), unless i didn't understand correctly and `CachedBlockNumber` is actually meant to be part of the "regular user API"

gateway/storage/store_test.go
kushnireyal: I mean, dead code. `err` is always nil

kushnireyal (commented):
Looks good. See comments

common/proposal.go
ale-linux: I wanted to avoid having a generic

```go
[]byte{0x00}
```

being mistaken for an evm proposal so I thought a non-zero `0xfb` byte would be less likely to happen there by accident. Wdyt?

endorser/app/factory.go
ale-linux: I thought about that, but I couldn't see a clean way of doing it without a lot of code duplication.. wdyt?

endorser/app/factory.go
ale-linux: good catch - this was a rebasing snafu...

endorser/testimpl/lightkvs.go
ale-linux: good point, done

endorser/testimpl/lightkvs.go
ale-linux: yeah it does run for the whole suite but it is sufficient (tests do indeed pass). Note that we rewind every time we revert so actually a number far smaller than 128 is sufficient

endorser/testimpl/lightkvs.go
```diff
@@ -0,0 +1,314 @@
...
+	"log"
+	"maps"
+
+	"github.com/hyperledger/fabric-x-evm/endorser"
+)
+
+// LightKVSExt extends LightKVS with additional test-specific functionality.
+// It embeds the base LightKVS and adds methods for snapshot management and revert operations.
+// Uses the base LightKVS.NextIndex for sequential storage instead of ring buffer.
+type LightKVSExt struct {
```
arner, line 23: Wondering if LightKVS itself isn't a testimpl, as it only runs in-memory? Merging them could make it easier to grasp.

endorser/lightkvs.go
arner: See my earlier comment, if we merge them we don't have to

endorser/executor.go
ale-linux: So here's the thing: I wanted to add logging to the StateDB but refrained from it because the DB is on the hotpath and I felt that any call to logging would slow things down. At the same time for debugging purposes it's useful to have the possibility of enabling logging. I created `statedb_logger.go` so that we can wrap the DB and create a logger, but refrained from enabling by default - I think we should have another PR to enable logging and left this as the trace for how to do it - wdyt?

endorser/executor.go
ale-linux: whoops... :)

endorser/executor.go
ale-linux: Thanks! Gone

endorser/lightkvs.go
ale-linux: yeah I think we have to... that was the cleanest way to wrap this object with one that implements the test-relevant functionality

endorser/statedb.go
ale-linux: Yeah this is a big Q. I _think_ it's okay to leave it here... the point is anyway that in the production code, if `record.IsDelete` is set, then `record.Value` is nil and so it's okay to bubble it out of the if.

We take advantage of this in the test code by setting `record.IsDelete` to true _AND_ `record.Value` to be non-nil so that we can revert safely without breaking MVCC.

Wdyt? @arner ?

endorser/statedb_logger.go
ale-linux: yeah - see here https://github.com/hyperledger/fabric-x-evm/pull/167/changes#r3226952715

endorser/versioned_db_wrapper.go
ale-linux: I left that intentionally - ultimately I think all components should have debug logs.. note that by default `DEBUG` is muted anyway

gateway/app/app.go
ale-linux: A nice power-of-two... 😓 128 was large enough so that the window was sufficient to keep the whole history for the hardhat tests. In practice probably a smaller value is also ok but in doing so we don't sacrifice much because this is almost for free

endorser/versioned_db_wrapper.go
arner: +1. We have the loggingHandler in gateway/api/server.go, let's just reuse that one and expand it if needed (at least change it to flogging)?

gateway/core/chain.go
ale-linux: Yeah so ultimately yes, we should nicely handle this without panicking. At the same time, swallowing the error forced me to spend a lot of time chasing one parsing issue.. for now I vote to leave the panic so that we see if anything breaks, and I'll create an issue to then switch to a cleaner error. Point is, if we hit this error, we have a serious problem because it means this namespace ordered and committed a valid ethereum tx which we don't recognise and can't process...

gateway/storage/store.go
ale-linux: gateway/testimpl/store.go needs it to cleanly wrap this and extend it with the snap/revert functionality

gateway/storage/store_test.go
ale-linux: yes, CachedBlockNumber should also be a part of the default impl. The idea is that we query the current block number quite often and it took one trip to the DB for each - now I just cache it and return that number instead.

gateway/core/chain.go
arner: I tend to agree, if we have a dirty ledger that causes this panic, a whole network will be dead (every node will keep choking on it) and it cannot recover without a code change and everyone upgrading... So for the future yes, log & continue... And more, I think we need some prometheus metrics or other tools for this to make it very visible to an operator that they should do something!

Counterpoint, if something causes it during this stage of development, we would definitely want to know. Maybe keep it for now, and include the removal of the panic in one of the "productionalization" stories, like the one for error handling?

gateway/storage/store_test.go
ale-linux: 🤦  sorry...

ale-linux (commented):
Thanks @kushnireyal - see new drop + my answers

gateway/testimpl/hardhat_helpers.go
```diff
@@ -16,8 +16,12 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/hyperledger/fabric-lib-go/common/flogging"
+	endorsertestimpl "github.com/hyperledger/fabric-x-evm/endorser/testimpl"
 )
 
+var hardhatLogger = flogging.MustGetLogger("gateway.testimpl.hardhat")
```
arner, line 23: I haven't noticed the switch to flogging (but I like it). Can we make a github issue for making the logging consistent? (flogging everywhere, which also means that we only inject it in the SDK -- our own constructors don't need the logger as argument anymore).

gateway/testimpl/store.go
```diff
@@ -0,0 +1,267 @@
+/*
```
arner, line 1: You know my point of view on this one, I'm gonna just skip reviewing it 😅

integration/fablo.hardhat.yaml
```diff
@@ -0,0 +1,59 @@
...
+      endpoint:
+        host: 127.0.0.1
+        port: 7041
+      tls:
+        mode: tls
+        server-name: peer0.org1.example.com
+        ca-cert-paths:
+          - ../testdata/fablo/fablo-target/fabric-config/crypto-config/peerOrganizations/org1.example.com/peers/peer0.org1.example.com/tls/ca.crt
+    database:
+      connection-string: "file:endorser-org1.db"
```
arner: This can be removed because we default to in-memory LightKVS

gateway/core/chain.go
ale-linux: https://github.com/hyperledger/fabric-x-evm/issues/171 tracks this under the code hygiene epic

endorser/executor.go
ale-linux: https://github.com/hyperledger/fabric-x-evm/issues/170 tracks the work needed to do this cleanly

endorser/versioned_db_wrapper.go
ale-linux: not sure what you mean - are you happy with the debug statements? I think I use flogging everywhere...

endorser/testimpl/lightkvs.go
ale-linux: Hmm, I'm not yet sure... I haven't discounted the possibility yet that we just use exactly this code for in-mem caching + persist in pebble every so often. We need to design this properly. What's for sure is that what I parked under `testimpl` will never be production code. For this file, the jury is still out imho

gateway/testimpl/hardhat_helpers.go
ale-linux: Done! https://github.com/hyperledger/fabric-x-evm/issues/172

gateway/testimpl/store.go
ale-linux: :grin:

endorser/statedb.go
arner: I think you're right @ale-linux ... as long as we define the RWS (we do) and we define how to parse to it (we do), and we test it with both Fabric and Fabric-X (we do?), we should be ok.

Do we have integration tests that cover this?

arner (approved):
LGTM!

PR #169, opened 2026-05-12
Perf: Implement wrap‑around replay logic

kushnireyal:
Adds configurable wrap-around replay for the USDC perf dataset and enables repeated replay of the same signed transactions in the perf/demo harness. This change introduces replay window and wrap controls, demo-scoped nonce bypass handling across gateway and endorser paths, and documentation updates for full-dataset runs and local testing guidance

### DIFF
--- a/integration/perf/replay_json_dataset_test.go
+++ b/integration/perf/replay_json_dataset_test.go
@@ -27,6 +27,7 @@ import (
 	econf "github.com/hyperledger/fabric-x-evm/endorser/config"
 	"github.com/hyperledger/fabric-x-evm/endorser/testimpl"
 	gwcore "github.com/hyperledger/fabric-x-evm/gateway/core"
+	gwtestimpl "github.com/hyperledger/fabric-x-evm/gateway/testimpl"
 	"github.com/hyperledger/fabric-x-evm/integration"
 	"github.com/hyperledger/fabric-x-evm/utils"
 	"github.com/hyperledger/fabric-x-sdk/endorsement"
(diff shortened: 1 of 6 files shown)


endorser/executor.go
```diff
@@ -27,6 +27,30 @@ import (
 	"github.com/hyperledger/fabric-x-sdk/endorsement"
 )
 
+// BalancePrimingConfig configures automatic ERC-20 balance priming.
```
ale-linux: I think the idea was to leave this file as-is for production and just extend https://github.com/hyperledger/fabric-x-evm/blob/main/endorser/testimpl/balance_priming_executor.go with the nonce priming (it already takes care of balance priming)

integration/test_helpers.go
```diff
@@ -115,7 +115,7 @@ func (th *TestHarness) PrimeStateFromJSON(ctx context.Context, jsonFilePath stri
 //
 // Sync goroutines are started in the background using ctx. The returned synchronizers
 // can be used by callers that need to wait for the initial sync to complete.
-func buildTestHarness(t *testing.T, logger sdk.Logger, cfg config.Config, evmConfig endorser.EVMConfig, primeDBPath string, bypass bool, ends []core.Endorser, dbs []endorser.KVS, builders []endorsement.Builder) (*TestHarness, *network.Synchronizer, error) {
+func buildTestHarness(t *testing.T, logger sdk.Logger, cfg config.Config, evmConfig endorser.EVMConfig, primeDBPath string, bypass bool, ends []core.Endorser, dbs []endorser.KVS, builders []endorsement.Builder, skipNonceValidation bool) (*TestHarness, *network.Synchronizer, error) {
```
ale-linux: I think if we do things right, we don't have to touch this code

ale-linux (changes requested):
Thanks Eyal - I think we can contain these changes in the testimpl like I did for the balance priming..

endorser/executor.go
kushnireyal: Ah sorry, you're right

integration/test_helpers.go
kushnireyal: right, fixed

integration/perf/replay_json_dataset_test.go
```diff
@@ -99,21 +162,42 @@ func runReplayTest(t *testing.T, processingWorkerCount int, submittingWorkerCoun
...
+			continue
+		}
+		sender, err := types.Sender(signerForTx(tx), tx)
+		if err != nil {
+			continue
+		}
+		if _, seen := senderStartNonce[sender]; !seen {
+			senderStartNonce[sender] = tx.Nonce()
+		}
+	}
```
ale-linux: This appears unused

ale-linux (commented):
LGTM - modulo what appears to be dead code

ale-linux (approved):

Issue #43, opened 2026-03-26
Implement Gas Handling for Tests


ale-linux:
Enhance the testing environment to properly handle Ethereum gas mechanics if required by compatibility tests. While fabric-x-evm tracks gas for metering purposes without charging fees, some Ethereum tests may require accurate gas accounting, gas limit enforcement, or gas price handling. Implement gas estimation for transactions, ensure gas limits are respected during EVM execution, and return accurate gas used values in transaction receipts. Configure appropriate gas limits for test scenarios and ensure the EVM execution engine correctly tracks and reports gas consumption. Document any differences between fabric-x-evm's gas handling and standard Ethereum to help users understand the system's behavior.

---

**Quarter:** Q2

arner:
Closing, as this is now covered by:

- COMPATIBILITY.md
- https://github.com/hyperledger/fabric-x-evm/issues/241
- https://github.com/hyperledger/fabric-x-evm/issues/257

PR #174, opened 2026-05-13
gateway/core: surface tx-conversion errors instead of panicking

### LINKED ISSUE #171: Remove panic in gateway/core/chain.go
gateway/core/chain.go panics when a transaction fails parsing in `convertTransaction`. This is a terminal error because this transaction has already been endorsed, ordered and committed, and so this must never happen. For now we want to surface all of these errors but as we cut a release, we should remove panics.

Storm1289:
## Summary

`Chain.convertToDomain` panicked when `convertTransaction` failed to parse a transaction's ETH payload, with a `// we surface this for now instead of swallowing it` comment. The path is genuinely "should never happen" (the tx is already endorsed, ordered and committed), but panicking inside the block-ingest goroutine takes down the gateway and gives the `BlockHandler` caller no chance to react.

This PR replaces the panic with proper error propagation, matching the maintainer's note on #171 surface the error now, decide on stronger handling closer to release.

## Changes

- **[gateway/core/chain.go](gateway/core/chain.go)**
  - `convertToDomain` signature changed to `(domain.Block, error)`. On `convertTransaction` failure it now returns `fmt.Errorf("convert tx %s: %w", tx.ID, err)` instead of panicking. The failing Fabric tx id is included so operators can correlate with the ledger.
  - `Handle` (which already returned `error` for `blocks.BlockHandler`) propagates the error untouched.

- **[gateway/core/chain_test.go](gateway/core/chain_test.go)**
  - All 5 existing `convertToDomain` call sites updated for the new return signature.
  - New test `TestConvertToDomain_ReturnsErrorOnUnparseableTx` exercises the previously-panicking path: a tx with a valid `ProposalTypeEVMTx` marker but unparseable ETH bytes. Asserts an error is returned and that the tx id appears in the message.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./gateway/core/` —10/10 pass, including the new error-path test

Closes #171

### DIFF
--- a/gateway/core/chain.go
+++ b/gateway/core/chain.go
@@ -72,7 +72,10 @@ func NewChain(dbConnStr, triePath string, withTrie bool) (*Chain, error) {
 // Handle implements blocks.BlockHandler. It commits the block's write sets to the trie,
 // then persists the block and its transactions to the database.
 func (c *Chain) Handle(ctx context.Context, b blocks.Block) error {
-	ebl := c.convertToDomain(b)
+	ebl, err := c.convertToDomain(b)
+	if err != nil {
+		return err
+	}
 
 	if c.ts != nil {
 		stateRoot, err := c.ts.Commit(ctx, b)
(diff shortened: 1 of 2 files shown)


Storm1289:
Hi @arner @ale-linux, I have drafted the PR for issue #171 whenever you get a chance, could you please take a look?
Also, I’d like to start the discussion/work for issues #43 .

Issue #134, opened 2026-04-30
Wire Gateway and Chain Together


ale-linux:
**Goal**: Connect the Gateway and Chain components so they can communicate about transaction completion, establishing the full notification pipeline from block commitment to pending status updates.

### Task 4.1: Add Chain Reference to Gateway
**What to do**: Modify the Gateway struct to hold a reference to the Chain instance.

**Details**:
- Add field to Gateway struct: `chain *Chain`
- This creates a dependency from Gateway to Chain
- Document why this reference is needed (for callback registration)
- Consider whether this creates any circular dependency issues (Gateway → Chain → Store, Gateway uses Store)

**Why**: The Gateway needs access to the Chain to register its callback during initialization.

### Task 4.2: Update Gateway Constructor
**What to do**: Modify the `New()` function in `gateway/core/api.go` to accept a Chain parameter.

**Details**:
- Change signature from: `func New(ec *EndorsementClient, submitter Submitter, store Store, chainID int64, workerCount int) (*Gateway, error)`
- To: `func New(ec *EndorsementClient, submitter Submitter, store Store, chain *Chain, chainID int64, workerCount int) (*Gateway, error)`
- Store the chain parameter in `g.chain = chain`
- Update the return statement to include the chain field
- Note: The Store parameter might become redundant since Chain embeds Store, but keep it for now for backward compatibility

**Why**: The Gateway needs the Chain reference at construction time to set up the callback relationship.

### Task 4.3: Register Callback in Gateway Initialization
**What to do**: In the Gateway's `Start()` method or immediately after construction, register a callback with the Chain that removes completed transactions from the queue.

**Details**:
- Add code after Gateway construction or in `Start()`:
  ```go
  g.chain.SetTxCompletionCallback(func(txHashes []common.Hash) {
      for _, hash := range txHashes {
          g.txQueue.MarkCompleted(hash)
      }
  })
  ```
- Consider adding logging: "Registered transaction completion callback"
- Ensure this happens before any blocks are processed
- Consider whether to do this in `New()` or `Start()` (probably `New()` is safer)

**Why**: This is the critical connection that closes the loop: when Chain commits a block, it notifies Gateway, which updates the queue.

### Task 4.4: Update Gateway Instantiation Sites
**What to do**: Find all places where `Gateway.New()` is called and update them to pass the Chain instance.

**Details**:
- Search for calls to `core.New()` or `gateway.New()` (likely in `cmd/` directory)
[...]

Issue #133, opened 2026-04-30
Create Callback Mechanism for Transaction Completion


ale-linux:
**Goal**: Establish a notification system where the Chain component can inform the Gateway when transactions have been committed to blocks, enabling the Gateway to update its pending transaction tracking.

### Task 3.1: Define Callback Type
**What to do**: Define a callback function type in `gateway/core/api.go` that will be invoked when transactions are completed.

**Details**:
- Add type definition: `type TxCompletionCallback func(txHashes []common.Hash)`
- The callback receives a slice of transaction hashes that were just committed
- Document that this callback should be fast and non-blocking
- Consider whether the callback should return an error (probably not needed)

**Why**: A well-defined callback interface makes the contract between Chain and Gateway explicit and type-safe.

### Task 3.2: Add Callback Registration to Gateway
**What to do**: Add a method to the Gateway struct that allows external components to register a completion callback.

**Details**:
- Add a field to Gateway struct: `txCompletionCallback TxCompletionCallback`
- Add method: `func (g *Gateway) RegisterTxCompletionCallback(callback TxCompletionCallback)`
- Store the callback in the struct field
- Consider thread-safety if the callback might be registered after Start() is called

**Why**: This allows the Chain to register its callback with the Gateway, creating the notification pathway.

### Task 3.3: Modify Chain.Handle to Extract and Notify
**What to do**: Update the `Handle()` method in `gateway/core/chain.go` to extract transaction hashes from committed blocks and invoke the callback.

**Details**:
- After successfully inserting the block (after `c.Store.InsertBlock(ctx, ebl)`), extract transaction hashes
- Iterate through `ebl.Transactions` and collect all `TxHash` values into a `[]common.Hash` slice
- If a callback is registered, invoke it: `if c.txCompletionCallback != nil { c.txCompletionCallback(txHashes) }`
- Ensure callback errors don't break block processing (wrap in recover or handle errors gracefully)
- Add logging for debugging: "Notifying completion of %d transactions"

**Why**: This is where the Chain notifies the Gateway that transactions have been finalized, triggering the removal from the in-progress map.

### Task 3.4: Add Callback Support to Chain Struct
**What to do**: Add the necessary fields and methods to the Chain struct to support callback registration.

**Details**:
- Add field to Chain struct: `txCompletionCallback TxCompletionCallback`
- Add method: `func (c *Chain) SetTxCompletionCallback(callback TxCompletionCallback)`
- Store the callback in the struct field
[...]

Issue #132, opened 2026-04-30
Integrate Pending Status into TransactionByHash


ale-linux:
**Goal**: Modify the `TransactionByHash()` API method to accurately report whether a transaction is pending by consulting the queue before querying the database.

### Task 2.1: Modify TransactionByHash Logic
**What to do**: Update the `TransactionByHash()` method in `gateway/core/api.go` to check the queue first.

**Details**:
- At the start of the method, call `isPending := g.txQueue.IsPending(hash)`
- If `isPending` is true, return `nil, true, nil` immediately
  - First return value (nil): No transaction data available yet (it hasn't been committed)
  - Second return value (true): Transaction is pending
  - Third return value (nil): No error occurred
- If not pending, proceed with the existing database query logic
- If found in database, return the transaction with `isPending = false`

**Why**: This implements the correct Ethereum behavior where pending transactions are reported as such, and clients can distinguish between "pending", "completed", and "not found" states.

### Task 2.2: Handle Three Transaction States
**What to do**: Ensure the return logic properly distinguishes between all three possible states.

**Details**:
- **State 1 - Pending**: Transaction in queue → return `nil, true, nil`
- **State 2 - Completed**: Transaction in database → return `tx, false, nil`
- **State 3 - Not Found**: Transaction in neither location → return `nil, false, nil`
- Add comments explaining each case
- Consider logging for debugging (e.g., "Transaction %s is pending" or "Transaction %s not found")

**Why**: Ethereum clients rely on these three states to provide accurate feedback to users. Incorrect status reporting breaks user experience.

Storm1289:
Hi @ale-linux I'd like to pick this up as the next step after @rootp1's PR for #131 lands. The integration here is a thin layer on top of `IsPending` / `MarkCompleted`, so it makes sense to wait for the underlying queue refactor to merge before opening this PR. Will
reference #131 in the PR description once it's in.

arner:
Done in https://github.com/hyperledger/fabric-x-evm/pull/176

Issue #131, opened 2026-04-30
Refactor TxQueue to Support Pending State Tracking


ale-linux:
**Goal**: Transform the simple FIFO queue into a sophisticated two-queue system that can track both pending and in-progress transactions, enabling the Gateway to answer "is this transaction pending?" queries.

### Task 1.1: Modify TxQueue Structure
**What to do**: Refactor the `TxQueue` struct in `gateway/core/txqueue.go` to maintain two separate data structures instead of one.

**Details**:
- Replace the single `queue []*types.Transaction` slice with:
  - `pendingQueue []*types.Transaction` - FIFO queue for transactions waiting to be processed
  - `inProgressMap map[common.Hash]*types.Transaction` - Map for fast lookup of transactions currently being processed by workers
- Keep the existing `mu sync.Mutex` and `cond *sync.Cond` for synchronization
- Keep the `done bool` flag for shutdown signaling
- The map allows O(1) lookup by transaction hash, which is essential for the `IsPending()` query

**Why**: The current implementation loses track of transactions once they're dequeued. By keeping them in `inProgressMap`, we can answer pending status queries accurately.

### Task 1.2: Add IsPending Method
**What to do**: Create a new public method `IsPending(txHash common.Hash) bool` that checks if a transaction is in either the pending queue or in-progress map.

**Details**:
- Lock the mutex before checking both data structures
- Iterate through `pendingQueue` to check if any transaction has the matching hash
- Check if the hash exists as a key in `inProgressMap`
- Return `true` if found in either location, `false` otherwise
- This method will be called by `TransactionByHash()` to determine pending status

**Why**: This is the core query method that enables the Gateway to report accurate pending status to Ethereum clients.

### Task 1.3: Modify Dequeue to Mark In-Progress
**What to do**: Change the `Dequeue()` method's behavior from "remove and return" to "move to in-progress and return".

**Details**:
- When a transaction is dequeued from `pendingQueue`, don't discard it
- Instead, compute its hash using `tx.Hash()` and add it to `inProgressMap[hash] = tx`
- Remove it from the front of `pendingQueue` as before
- Return the transaction to the worker for processing
- The transaction now exists only in `inProgressMap` until it's completed or fails

**Why**: This keeps the transaction tracked during the endorsement and submission process, which can take several seconds. Without this, there's a gap where the transaction appears to not exist.

### Task 1.4: Add MarkCompleted Method
[...]

PR #176, opened 2026-05-18
Implement transaction tracking with pending status

ale-linux:
This commit transforms the Gateway's simple FIFO transaction queue into a two-queue system that tracks both pending and in-progress transactions, enabling accurate responses to "is this transaction pending?" queries from Ethereum clients.

The core change refactors TxQueue to maintain two separate data structures: a pendingQueue slice for transactions waiting to be processed, and an inProgressMap for fast O(1) lookup of transactions currently being endorsed and submitted by workers. This eliminates the previous gap where transactions would disappear from tracking once dequeued, making them appear non-existent to clients during the multi-second endorsement process.

To support this, the Backend interface's TransactionByHash signature was simplified to return only the transaction and error, with pending status now indicated by BlockNumber being 0 (which the API layer converts to null in JSON responses). The Gateway was registered as a block handler in the synchronizer, positioned after the chain handler to ensure blocks are persisted before transactions are marked complete.

A critical test regression was discovered in Hardhat tests. The solution was to override SendRawTransaction in the test API to make it synchronous by polling until the transaction is committed, mimicking Hardhat's auto-mining behavior.

Closes: 

- https://github.com/hyperledger/fabric-x-evm/issues/131
- https://github.com/hyperledger/fabric-x-evm/issues/132
- https://github.com/hyperledger/fabric-x-evm/issues/133
- https://github.com/hyperledger/fabric-x-evm/issues/134

### DIFF
--- a/gateway/api/eth.go
+++ b/gateway/api/eth.go
@@ -49,7 +49,8 @@ type Backend interface {
 	CallContract(ctx context.Context, call ethereum.CallMsg, blockNumber *big.Int) ([]byte, error) // ethereum.ContractCaller
 
 	// Transactions. Our transactions also include the status, so we can build receipts out of the same data.
-	TransactionByHash(ctx context.Context, hash common.Hash) (tx *domain.Transaction, isPending bool, err error)
+	// For pending transactions, BlockNumber will be 0 (converted to null in JSON response).
+	TransactionByHash(ctx context.Context, hash common.Hash) (*domain.Transaction, error)
 	GetTransactionByBlockHashAndIndex(ctx context.Context, hash common.Hash, idx int64) (*domain.Transaction, error)
 	GetTransactionByBlockNumberAndIndex(ctx context.Context, num uint64, idx int64) (*domain.Transaction, error)
 	GetLogs(ctx context.Context, query domain.LogFilter) ([]domain.Log, error)
@@ -256,7 +257,7 @@ func (api *EthAPI) SendRawTransaction(ctx context.Context, input hexutil.Bytes) 
 // eth_getTransactionByHash
 func (api *EthAPI) GetTransactionByHash(ctx context.Context, hash common.Hash) (*RPCTransaction, error) {
 	logger.Debugf("EthAPI.GetTransactionByHash() called with hash=%s", hash.Hex())
-	tx, _, err := api.b.TransactionByHash(ctx, hash)
+	tx, err := api.b.TransactionByHash(ctx, hash)
 	if err != nil {
 		logger.Debugf("EthAPI.GetTransactionByHash() returning error: %v", err)
 		return nil, err
@@ -301,7 +302,7 @@ func (api *EthAPI) GetTransactionByBlockNumberAndIndex(ctx context.Context, num 
 // eth_getTransactionReceipt
 func (api *EthAPI) GetTransactionReceipt(ctx context.Context, hash common.Hash) (*rpcReceipt, error) {
 	logger.Debugf("EthAPI.GetTransactionReceipt() called with hash=%s", hash.Hex())
-	r, _, err := api.b.TransactionByHash(ctx, hash)
+	r, err := api.b.TransactionByHash(ctx, hash)
 	if err != nil {
 		logger.Debugf("EthAPI.GetTransactionReceipt() returning error: %v", err)
 		return nil, err
--- a/gateway/api/rpcerrors_test.go
+++ b/gateway/api/rpcerrors_test.go
@@ -9,63 +9,14 @@ package api
 import (
 	"errors"
 	"fmt"
-	"math/big"
 	"testing"
 
-	"github.com/ethereum/go-ethereum/common"
 	ethcore "github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/txpool"
-	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/hyperledger/fabric-x-evm/gateway/api/rpcerr"
 	"github.com/hyperledger/fabric-x-evm/gateway/domain"
 )
-
-func TestNilBlockNum(t *testing.T) {
-	nonce := uint64(0)
-	to := common.HexToAddress("0xcafe")
-	value := big.NewInt(1e18)
-	gasLimit := uint64(21000)
-	maxFeePerGas := big.NewInt(20000000000)
-	maxPriorityFeePerGas := big.NewInt(2000000000)
-	tx := types.NewTx(&types.DynamicFeeTx{
-		ChainID:   big.NewInt(1),
-		Nonce:     nonce,
-		GasFeeCap: maxFeePerGas,
-		GasTipCap: maxPriorityFeePerGas,
-		Gas:       gasLimit,
-		To:        &to,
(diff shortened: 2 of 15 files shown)


ale-linux:
FYI @rootp1, @Storm1289 - your review is welcome

gateway/api/eth.go
```diff
@@ -301,7 +302,7 @@ func (api *EthAPI) GetTransactionByBlockNumberAndIndex(ctx context.Context, num
 // eth_getTransactionReceipt
 func (api *EthAPI) GetTransactionReceipt(ctx context.Context, hash common.Hash) (*rpcReceipt, error) {
 	logger.Debugf("EthAPI.GetTransactionReceipt() called with hash=%s", hash.Hex())
-	r, _, err := api.b.TransactionByHash(ctx, hash)
+	r, err := api.b.TransactionByHash(ctx, hash)
```
Storm1289, line 305: Hi @ale-linux, found something while going through this - dropping the `isPending` bool from `TransactionByHash` here looks like it might cause `eth_getTransactionReceipt` to panic for any pending tx.

The new signal `Gateway.TransactionByHash` uses for pending (`BlockHash=nil`, `BlockNumber=0`) flows straight into `receipt(r)` a few lines below, and `receipt()` only checks if `r == nil` before casting `common.Hash(r.BlockHash)` at `models.go:83`. With `BlockHash=nil` that cast panics (length-0 slice → `[32]byte`).

Tried reproducing it locally with a stub backend returning a pending-shaped tx, and the panic confirms:

```text id="e8q6xl"
runtime error: cannot convert slice with length 0 to array or pointer to array with length 32
  at gateway/api/models.go:83  →  BlockHash: common.Hash(r.BlockHash)
  at gateway/api/eth.go:310    →  result := receipt(r)
```

The integration test doesn't surface it because the test API override only polls `TransactionByHash` directly, never `GetTransactionReceipt`. But real clients (ethers.js, Hardhat, web3.py) submit and immediately start polling the receipt, so they'd hit this during the pending window.

A small guard returning `nil` when `r == nil || r.BlockNumber == 0` would fix it and also match the JSON-RPC spec (`null` for not-yet-mined txs).

ale-linux: Great catch @Storm1289, thanks. Fixed

gateway/testimpl/test_eth_api.go
```diff
@@ -132,10 +132,47 @@ func (api *TestEthAPI) SendTransaction(ctx context.Context, args TransactionArgs
...
+
+	// Poll until the transaction is committed (has a block number > 0)
+	// This mimics Hardhat's auto-mining behavior where transactions are mined immediately
+	for {
+		select {
+		case <-ctx.Done():
+			return common.Hash{}, ctx.Err()
+		default:
+			// Check if transaction is committed
+			tx, err := api.backend.TransactionByHash(ctx, txHash)
```
kushnireyal, line 162: TransactionByHash can return (nil, nil) (if transaction not found in queue or database), so we need to add if clause that continues if this is the case

gateway/testimpl/test_eth_api.go
```diff
@@ -132,10 +132,47 @@ func (api *TestEthAPI) SendTransaction(ctx context.Context, args TransactionArgs
...
+			if err != nil {
+				// Transaction not found yet, continue polling
```
kushnireyal, line 164: Consider adding some logging or something to print the error if it's non-nil

gateway/core/chain_test.go
```diff
@@ -104,7 +106,8 @@ func TestConvertToDomain_SkipsInsufficientInputArgs(t *testing.T) {
 	}
 
 	c := &Chain{}
-	got := c.convertToDomain(b)
+	_ = c
```
kushnireyal: if 'c' is not used, just remove

gateway/core/chain_test.go
```diff
@@ -120,15 +123,17 @@ func TestConvertToDomain_SkipsInvalidEthBytes(t *testing.T) {
...
```
kushnireyal: same comment, if 'c' is not used, just remove

kushnireyal: and there are other places in this file with the same thing

gateway/api/rpcerrors_test.go
```diff
@@ -9,15 +9,64 @@ package api
...
 	ethcore "github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/txpool"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/hyperledger/fabric-x-evm/gateway/api/rpcerr"
 	"github.com/hyperledger/fabric-x-evm/gateway/domain"
 )
 
+func TestNilBlockNum(t *testing.T) {
```
kushnireyal: Does this test belong to this test suite?

gateway/testimpl/test_eth_api.go
kushnireyal: with this in mind, maybe it will be good to add a unittest that covers this case

kushnireyal (approved):
Looks good! see comments

gateway/testimpl/test_eth_api.go
```diff
@@ -132,10 +132,47 @@ func (api *TestEthAPI) SendTransaction(ctx context.Context, args TransactionArgs
...
+			}
+
+			// Check if transaction has been included in a block
+			// BlockNumber is 0 for pending transactions, > 0 for committed
+			if tx.BlockNumber > 0 {
+				// Transaction is committed
+				return txHash, nil
+			}
+
+			// Transaction is still pending, continue polling
```
arner, line 182: maybe a tiny (20ms or so) sleep would be good to give the database a break?

arner (approved):

gateway/testimpl/test_eth_api.go
ale-linux: well spotted - but actually I think this must never happen right? The call to SendRawTransaction will synchronously add it to the pending queue so it will either be there, or in the inprogress map, or committed in the DB. `tx == nil` there would be a programming error of some kind that we don't understand yet. It's good to bubble it up though, so I added a `panic` there (which is probably fine since this is a test impl..). Should we maybe still remove it at some point? But to do what? If we don't find it once, we might never find it and loop forever?

gateway/testimpl/test_eth_api.go
ale-linux: Done

gateway/core/chain_test.go
ale-linux: :facepalm:... gone

gateway/api/rpcerrors_test.go
ale-linux: Good point - moved

gateway/testimpl/test_eth_api.go
ale-linux: I wouldn't want a whole sleep because it might be too much... how about 

```go
runtime.Gosched()
```

?

arner: It would still be a bit of a DoS attack on the db so I'd go for sleep but your call!

ale-linux: well but this is only for the test gateway right? I mean, we are never meant to use it anywhere near production..

PR #183, opened 2026-05-21
validate config and add logging config

arner:
fail fast with invalid config, and add some examples to the readme.


ale-linux (approved):
LGTM - let's just rebase and let the CI run

PR #184, opened 2026-05-21
Fix polling loop and implement ERC-20-aware mempool

ale-linux:
This commit fixes a critical bug in the transaction polling loop that was causing premature test failures and implements an ERC-20-aware mempool to eliminate MVCC conflicts through intelligent transaction scheduling.

The polling loop in integration/perf/replay_json_dataset_test.go was incorrectly timing out after approximately 100ms (100 iterations with minimal delay), which was insufficient for transactions to be committed during high-throughput testing. The fix removes the iteration counter and allows the loop to poll indefinitely until the transaction is found or an actual error occurs.

The main enhancement is a new dependency-aware transaction queue in gateway/core/txqueue.go that acts as an intelligent mempool. The queue tracks in-progress participants (both senders and recipients) for all transactions and gates any pending transaction that shares a participant with an in-flight transaction. This ensures transactions touching the same ERC-20 token accounts are properly serialized, preventing concurrent state modifications that cause MVCC conflicts. The implementation extracts participants from both regular transfers and ERC-20 token transfers by parsing the transfer function signature and recipient address from transaction data.

### DIFF
--- a/endorser/testimpl/balance_priming_statedb.go
+++ b/endorser/testimpl/balance_priming_statedb.go
@@ -8,16 +8,14 @@ package testimpl
 
 import (
 	"fmt"
-	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/holiman/uint256"
 	"github.com/hyperledger/fabric-x-evm/endorser"
 )
 
 // Default prime value: 1 billion tokens (with 6 decimals for USDC)
-var primeValue = new(uint256.Int).Mul(uint256.NewInt(1_000_000_000), uint256.NewInt(1_000_000))
+var primeValue = new(uint256.Int).Mul(uint256.NewInt(1_000_000_000_000), uint256.NewInt(1_000_000_000_000))
 
 // BalancePrimingWrapper wraps a StateDB and intercepts GetState calls to prime
 // ERC-20 balance slots with a high value when they are zero.
@@ -25,7 +23,6 @@ type BalancePrimingWrapper struct {
 	endorser.ExtendedStateDB
 	contractAddr    common.Address // The ERC-20 contract address
 	senderAddr      common.Address // The sender address to prime
-	balanceSlot     common.Hash    // The storage slot for the sender's balance
 	mappingPosition uint64         // The position of the balances mapping in storage
 	enabled         bool           // Whether priming is enabled
 	expectedNonce   uint64         // The nonce the current transaction expects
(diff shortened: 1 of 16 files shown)


gateway/core/txqueue_helpers.go
```diff
@@ -0,0 +1,74 @@
...
+		return common.Address{}, false
+	}
+
+	data := tx.Data()
+	if len(data) < 4+32+32 {
+		return common.Address{}, false
+	}
+
+	// Extract recipient from ERC20 transfer calldata (offset 4 + 12 bytes)
+	recipientOffset := 4 + 12
```
kushnireyal, line 60: This will catch `approve` calls as well, which are non blocking for transfer, so we'll get false positives in `dequeueReady`. Not super crucial, maybe just document it

gateway/core/txqueue.go
```diff
@@ -119,7 +201,33 @@ func (q *TxQueue) IsPending(txHash common.Hash) *types.Transaction {
...
+					delete(q.inProgressParticipants, participant)
+				} else {
+					q.inProgressParticipants[participant] = count - 1
+				}
+			}
+		}
+	}
+
+	if len(q.pendingQueue) > 0 {
+		q.cond.Signal()
```
kushnireyal: Shouldn't it be `Broadcast()`? For the case that there are multiple workers sleeping

gateway/core/txqueue.go
```diff
@@ -130,6 +238,13 @@ func (q *TxQueue) Handle(ctx context.Context, block *domain.Block) error {
 	// Mark all transactions in the block as complete
 	for _, tx := range block.Transactions {
 		txHash := common.BytesToHash(tx.TxHash)
+		q.total++
+		if tx.Status == 0 {
+			q.invalid++
```
kushnireyal, line 139: There might be a race condition here since you didn't locked `q.mu.Lock()` like we do in `Complete()`

endorser/testimpl/balance_priming_statedb.go
```diff
@@ -44,8 +44,6 @@ func NewBalancePrimingWrapper(stateDB endorser.ExtendedStateDB, contractAddr com
 
 // SetSender sets the sender address and calculates the balance slot.
 func (w *BalancePrimingWrapper) SetSender(sender common.Address) {
-	w.senderAddr = sender
-	w.balanceSlot = GetERC20BalanceSlot(sender, w.mappingPosition)
```
kushnireyal, line 48: Why did you removed these?

integration/perf/replay_json_dataset_test.go
```diff
@@ -240,8 +269,7 @@ func runReplayTest(t *testing.T, processingWorkerCount int, submittingWorkerCoun
 					}
 
 					// Wait for transaction to be committed
-					ctr := 0
-					for pending := true; pending && ctr < 100; ctr++ {
+					for pending := true; pending; {
```
kushnireyal, line 273: Kind of worried to remove the bailout altogether. Wdyt?

integration/perf/replay_json_dataset_test.go
```diff
@@ -370,9 +406,10 @@ func TestReplayJSONDataset(t *testing.T) {
 	if testing.Short() {
 		t.Skip("skipping in short mode")
 	}
+	// flogging.ActivateSpec("gateway.core.txqueue_v2=debug")
 
 	// Run the test with single worker configuration
-	_, _, _ = runReplayTest(t, 1, 1, loadReplayConfigFromEnv(t))
+	_, _, _ = runReplayTest(t, 1, 8, loadReplayConfigFromEnv(t))
```
kushnireyal: What's the reason for this change?

endorser/testimpl/balance_priming_statedb.go
ale-linux: Two reasons:

1. I still occasionally saw some failed txes
2. computing the hash on every transaction was expensive

This is simpler - basically if it's the right address, we always bump up balances if they are zero.

gateway/core/txqueue.go
ale-linux: I dropped all of this code - it was the one the AI generated which I didn't like and ended up re-writing in the v2 file. Now the original, simple queue is used by default, and the v2 one is used for the perf demo only

gateway/core/txqueue.go
ale-linux: dropped

gateway/core/txqueue_helpers.go
ale-linux: technically we don't have approve calls right?

integration/perf/replay_json_dataset_test.go
ale-linux: I agree we should definitely re-add it, but that was dangerous because it was giving us false numbers before... we need to experiment a little and then add some form of safeguard at some point. I've created https://github.com/ale-linux/pzf_demo/issues/38 to track the work

integration/perf/replay_json_dataset_test.go
ale-linux: yep, sorry, undone

kushnireyal (approved):

gateway/core/txqueue_helpers.go
kushnireyal: right

PR #185, opened 2026-05-25
Add namespace-filtered delivery, block DB retention, and USDC demo orchestration

kushnireyal:
Completes the EVM gateway changes for the USDC multi-replica performance demo,
on top of #190 (notification-based tx completion).

- **Namespace-filtered delivery**: each gateway replica skips blocks from other
  namespaces, preventing O(1/N) throughput degradation as replica count grows.
- **Block DB retention**: TruncateBlocks prunes SQLite to a configurable window
  (default 10k blocks), called every 1k delivered blocks to keep storage bounded
  during long replay runs.
- **SQLite scratch-image fix**: PRAGMA temp_store=MEMORY so WAL operations work
  in the release image built from scratch.
- **Demo orchestration** (scripts/run-demo.sh): single command from a developer's
  Mac to sync the Ansible collection on the staging control node, generate testdata,
  and stream the workload in real time; exposed via make run-demo. New
  --reset-fabricx flag does a full backend teardown + start + cert refresh +
  redeploy for reproducible measurements on a clean chain.
- **USDC deployment runbook** (integration/perf/USDC_deployment.md): staging
  topology, testdata steps, run commands, output format, troubleshooting, and a
  "Resetting between runs" guidance section.
- **Perf test extensions**: configurable worker counts via env vars, TPS stability
  stats, latency percentiles, sent/committed/inflight counters.


gateway/app/app.go
```diff
@@ -106,15 +106,17 @@ func buildApp(cfg config.Config, gwSigner sdk.Signer, logger sdk.Logger, endorse
 	case "fabric":
 		submitter, err = nfab.NewSubmitter(orderers, gwSigner, 0, logger)
 	case "fabric-x", "":
-		submitter, err = nfabx.NewSubmitter(orderers, gwSigner, 0, logger)
+		// Use armaSubmitter instead of nfabx.NewSubmitter: the ArMA orderer-router only sends
+		// a response while the gRPC stream is open, so Recv must come before CloseSend.
+		submitter, err = newArmaSubmitter(orderers, gwSigner, logger)
```
ale-linux: 2 points here:

1. I think this was fixed in the sdk [here](https://github.com/hyperledger/fabric-x-sdk/commit/3255417e8fda525ae4e9275f10a55d19d64e9b39)
2. with the new submitter approach proposed by Senthil ([this](https://github.com/hyperledger/fabric-x-sdk/pull/25) I think we should switch back right?

integration/perf/replay_json_dataset_test.go
```diff
@@ -284,11 +289,27 @@ func runReplayTest(t *testing.T, processingWorkerCount int, submittingWorkerCoun
...
+					// can still have status=0x0 if the EVM reverted (e.g. insufficient balance).
+					receipt, err := ec.TransactionReceipt(t.Context(), tx.Hash())
+					if err != nil {
+						t.Logf("Transfer %d: TransactionReceipt error: %v", i, err)
+						panic(err)
+					}
+					if receipt.Status != 1 {
+						t.Logf("Transfer %d: EVM revert (status=0x0), hash=%s", i, tx.Hash().Hex())
+						panic(fmt.Errorf("EVM revert"))
+					}
```
ale-linux: Good idea. But actually on second thought, let's just call `TransactionReceipt` directly. Both internally call the same function to retrieve the same info about the tx. You can just substitute the `TransactionByHash` with `TransactionReceipt`: you can test for

```go
			if err.Error() == "not found" {
				continue
			}
```

or for a nil receipt and keep polling, that's the signal for both "pending" and "never seen this one". If you get another error it means something else was wrong and we must count it as a failure. As soon as we get `err == nil && receipt != nil` it means the tx has committed and we can just stop polling. We should then verify the value of `Status` to decide whether it succeeded or not.

gateway/app/app.go
```diff
@@ -106,15 +106,17 @@ func buildApp(cfg config.Config, gwSigner sdk.Signer, logger sdk.Logger, endorse
...
 	default:
 		return nil, fmt.Errorf("unsupported protocol: %q", cfg.Network.Protocol)
 	}
 	if err != nil {
 		return nil, fmt.Errorf("failed to create submitter: %w", err)
 	}
 
-	chain, err := core.NewChain(cfg.Gateway.Database.ConnString, cfg.Gateway.Database.TriePath, false)
+	chain, err := core.NewChain(cfg.Gateway.Database.ConnString, cfg.Gateway.Database.TriePath, cfg.Network.Namespace, false)
```
ale-linux: also: would any of the changes here impact our tests? They don't use the gateway instance right?

ale-linux: As soon as we get decent numbers with this test, we need to refactor it to

- create the gateway process 
- refactor the loadgen into a process
- as part of the test deployment, spin up pains of loadgen/gw

gateway/core/api.go
```diff
@@ -126,6 +132,15 @@ func (g *Gateway) Start(ctx context.Context) {
...
 	defer g.wg.Done()
+	defer func() {
+		if r := recover(); r != nil {
+			logger.Errorf("worker panic: %v", r)
+			if ctx.Err() == nil {
+				g.wg.Add(1)
+				go g.worker(ctx)
+			}
+		}
+	}()
```
ale-linux: nice

gateway/app/app.go
```diff
@@ -196,8 +198,12 @@ func (a *App) Run(ctx context.Context) error {
 	g.Go(func() error { return a.gwSync.Start(gctx) })
 
 	// Wait for initial sync before serving traffic
+	syncTimeout := a.cfg.Gateway.SyncTimeout
+	if syncTimeout == 0 {
+		syncTimeout = 5 * time.Minute
```
ale-linux: isn't it too long?

gateway/core/api.go
```diff
@@ -151,7 +166,9 @@ func (g *Gateway) processTx(ctx context.Context, tx *types.Transaction) error {
 	if err := g.SubmitFabricTx(ctx, end); err != nil {
 		return err
 	}
-
+	if ct, ok := g.TxQueue.(CommitTracker); ok {
+		ct.RecordSubmit(tx.Hash())
+	}
```
ale-linux: Why?

gateway/core/chain.go
```diff
@@ -90,6 +118,12 @@ func (c *Chain) Handle(ctx context.Context, b blocks.Block) error {
 		return err
 	}
 
+	if c.Store.BlockRetention > 0 && ebl.BlockNumber%1000 == 0 {
```
ale-linux: why hardcode 1000? didn't we have a parameter in the other function?

gateway/core/chain.go
```diff
@@ -129,7 +161,8 @@ func ConvertToDomain(b blocks.Block) domain.Block {
 
 		etx, err := convertTransaction(tx.InputArgs[1], b.Hash, b.Number, tx.Number, tx.ID, status, tx.Status, tx.Events, &logIndex)
 		if err != nil {
-			panic(err) // we surface this for now instead of swallowing it
+			logger.Warnf("skipping malformed tx %s in block %d: %v", tx.ID, b.Number, err)
+			continue
```
ale-linux: let's keep the panic a little longer - I have a story to track this anyway so we won't forget

integration/test_helpers.go
```diff
@@ -153,12 +154,15 @@ func buildTestHarness(t *testing.T, logger sdk.Logger, cfg config.Config, evmCon
...
+		// Create network submitter.
+		// For fabric-x, use ArMASubmitter which uses Send→Recv (not Send→CloseSend→Recv).
+		// The ArMA orderer-router only sends a response while the stream is open; calling
+		// CloseSend before Recv causes the router to close the stream without responding.
 		switch cfg.Network.Protocol {
 		case "fabric":
 			submitter, err1 = nfab.NewSubmitter(orderers, gwSigner, 0, logger)
 		case "fabric-x", "":
-			submitter, err1 = nfabx.NewSubmitter(orderers, gwSigner, 0, logger)
+			submitter, err1 = NewArMASubmitter(orderers, gwSigner, logger)
```
ale-linux: same comment as the other - if the other patch gets merged we shouldn't need this

integration/test_helpers.go
```diff
@@ -405,11 +409,15 @@ func NewFabricXTestHarness(t *testing.T, logger sdk.Logger, evmConfig endorser.E
...
-	// cwd, _ := os.Getwd()
-	// defer os.Chdir(cwd)
-	// _ = os.Chdir("../")
+	configPath := os.Getenv("FABX_CONFIG_PATH")
+	if configPath == "" {
+		cwd, _ := os.Getwd()
+		defer os.Chdir(cwd)
+		_ = os.Chdir("../")
+		configPath = "fabx.yaml"
+	}
```
ale-linux, line 545: I don't think we need this right?

integration/test_helpers.go
```diff
@@ -426,11 +434,18 @@ func NewFabricXTestHarnessWithFactory(t *testing.T, logger sdk.Logger, evmConfig
...
 		return nil, err
 	}
 
+	syncTimeout := cfg.Gateway.SyncTimeout
+	if syncTimeout == 0 {
+		syncTimeout = 60 * time.Second
+	}
+	t.Log("waiting for synchronizer to connect to Fabric-X backend...")
+	waitUntilSynced(t, sync, syncTimeout)
+
```
ale-linux: @arner we did get rid of those right? Why do we still need them now?

ale-linux (changes requested):
Thx @kushnireyal - a few comments

gateway/app/app.go
kushnireyal: Yes, done

integration/test_helpers.go
kushnireyal: It's a fallback for `go test ./integration/perf/...` run from a fresh checkout without `FABX_CONFIG_PATH` set. `os.Chdir("../")` moves cwd from `integration/perf/` to `integration/`, where the in-repo `fabx.yaml` lives (integration/fabx.yaml).
On staging and in our run-demo.sh path, `FABX_CONFIG_PATH` is always set (staging uses /data/fabric-x-evm-test-config-N.yaml per replica), so the whole `if configPath == ""` branch is skipped there.
  
So today the code path serves only "developer runs `go test` directly without setting the env var." Happy to drop it and replace with an error reporting ("FABX_CONFIG_PATH must be set or run from integration/" or sometinh like that) if you prefer

PR #188, opened 2026-05-29
refactor executor config and make maxTxGas configurable

arner:
This moves some test-only features to the endorser `testimpl` wrappers. It simplifies constructors, for instance only passing the blockNumber where before we were passing a BlockInfo around it - or removing other fields which are not used anymore.


ale-linux (approved):
~LGTM - can I just ask you to first merge #190 and then this one? Because I think they'll conflict...~

Forget about it, they are not conflicting so feel free to merge soon as the CI run finishes - thanks!

Issue #108, opened 2026-04-21
Upgrade go-ethereum to latest version


arner:
Staying close to geth helps prevent rework in the future. Our current version (v1.16.7) is from November. Several things have changed in the meantime (besides bug and security fixes), so some changes in our endorser are required as part of the upgrade. Some highlights that may be relevant to us:

1.17

- eth_getTransactionByHash now returns the blockTimestamp as part of the transaction object (https://github.com/ethereum/go-ethereum/pull/33709)
- Error codes for some conditions in eth_simulateV1, eth_getLogs, eth_getStorageAt have been improved to better comply with the execution RPC spec (https://github.com/ethereum/go-ethereum/pull/33007, https://github.com/ethereum/go-ethereum/pull/33320, https://github.com/ethereum/go-ethereum/pull/33282)
- The RPC server now enables plain-text HTTP2 for improved performance (https://github.com/ethereum/go-ethereum/pull/33812)
- OpenTelemetry support for RPC API https://github.com/ethereum/go-ethereum/pull/33599, https://github.com/ethereum/go-ethereum/pull/33452, https://github.com/ethereum/go-ethereum/pull/33780, https://github.com/ethereum/go-ethereum/pull/33521

1.17.1
- New RPC method: eth_getStorageValues (https://github.com/ethereum/go-ethereum/pull/32591)

1.17.2
- Refactored state database in preparation for binary trie integration (https://github.com/ethereum/go-ethereum/pull/33816)

https://github.com/ethereum/go-ethereum/releases

PR #193, opened 2026-06-06
chore: bump go-ethereum to v1.17.3

Storm1289:
Refs #108 

## Summary

Bumps github.com/ethereum/go-ethereum from v1.16.7 to v1.17.3 (released 2026-05-11).

Stays on the Osaka fork; VerkleTime and post-Osaka features remain disabled.
New JSON-RPC methods introduced in v1.17.x (e.g. eth_getStorageValues) are intentionally not added.

## API changes addressed

| Area | Change in v1.17.3 |
|---|---|
| vm.StateDB interface | Added Touch, IsNewContract, LogsForBurnAccounts |
| vm.StateDB interface | PointCache removed; trie/utils package gone |
| vm.StateDB.SelfDestruct | No longer returns the previous balance; EIP-6780 logic folded in |
| vm.StateDB interface | SelfDestruct6780 removed |
| vm.StateDB.Finalise | Now returns *bal.StateAccessList |
| params.ChainConfig | VerkleTime field removed |
| core.Message | Gas/value fields became *uint256.Int |
| core.GasPool | AddGas replaced with core.NewGasPool(amount) |
| core.ErrMaxInitCodeSizeExceeded | Moved to vm.ErrMaxInitCodeSizeExceeded |
| state.CachingDB | Replaced with state.Database interface |

## EIP-6780 SelfDestruct semantics

Adds per-transaction "new contract" tracking in the snapshot StateDB so SelfDestruct matches the EIP-6780 / v1.17.3 behavior:

- Contracts created in the current transaction → fully destroyed
- Pre-existing contracts → only balance is transferred; account, code, and storage survive

Implementation details:

- `newContracts` map added on StateDB
- populated by `CreateAccount` / `CreateContract` via `markNewContract`
- `newContractEntry` journal entry added so `RevertToSnapshot` correctly undoes membership
- `IsNewContract` now reads from the tracked set instead of the previous stub

Also adds 4 new unit tests in:
```text id="b2k8k3"
endorser/state_test.go

### CODE REVIEW

**Comment by Storm1289** [comment]
Hi @ale-linux @arner , this pr doesn't fully close issue #108. During the bump, a set of ethereum/tests fixtures (27 files around SELFDESTRUCT / CREATE2 / extCodeHash) started failing under v1.17.3, and I haven't implemented the deeper StateDB changes needed to make them pass in this PR - they're added to testdata/eth_tests.slow so this PR can stay focused on the version bump.

A few questions on the follow-up:
1. Track under #108, or open a new sub-issue?
2. New v1.17.x RPC methods (eth_getStorageValues, etc.) in the same cycle, or later?

**Comment by arner** [comment]
Hi @Storm1289 , that's great, thank you. You said _"27 ethereum/tests fixtures around SELFDESTRUCT / CREATE2 / extCodeHash now produce post state root mismatch errors. Root cause: v1.17.3 folded the EIP-6780 SelfDestruct6780 semantics into the base SelfDestruct method, which our snapshot-backed StateDB doesn't yet mirror. Mirroring it is non-trivial and deserves its own change."_

Do you know what it would take, would it be a big change? And secondly, do you think it might be possible to already implement that in another PR, even before we merge this one? If possible I would prefer to not disable any compatibility tests on the main branch :)

**Comment by Storm1289** [comment]
Hi @arner ,fair point on not disabling tests, let me think through it properly.

The fix has two main parts. First, implement IsNewContract properly in our snapshot StateDB - track addresses created during the current tx (in CreateAccount / CreateContract), reset on tx boundary, expose via the stub I added. Second, make SelfDestruct EIP-6780-aware - if IsNewContract(addr) returns true, full destroy as today; otherwise just transfer balance to the beneficiary and don't add the account to the destroyed set. Plus the journal/replay side and tests for both branches.
A couple of things I'd like to confirm on the sequencing.
Should I open the SelfDestruct fix as a separate PR that lands before this one (so main never has disabled tests), and then rebase this PR to drop the 27 blacklist entries? Or would you prefer I fold the fix directly into this PR as one combined merge?

**Comment by arner** [comment]
Hard to tell.... It seems to me like it could still be part of this PR. It's quite readable as it is, and might be _more_ understandable rather than less if it includes 6780, considering the many comments it already has about that. But I'll leave the judgement to you!

**Comment by Storm1289** [comment]
@arner take a another look when you have time :)

File: endorser/executor.go [hunk 1]
```diff
@@ -333,18 +334,21 @@ func CallMsgToMessage(msg ethereum.CallMsg, baseFee *big.Int, skipNonceCheck, sk
 		blobGasFeeCap = new(big.Int)
 	}
 
+	// core.Message gas/value fields became *uint256.Int in go-ethereum v1.17.3.
```
arner: We can probably already use the uint256 higher up in this function, instead of creating big ints first and then converting them in the return value?

endorser/statedb.go
```diff
@@ -798,13 +801,24 @@ func (s *StateDB) Prepare(rules params.Rules, sender, coinbase common.Address, d
...
-func (s *StateDB) PointCache() *utils.PointCache { return nil }
-
 func (s *StateDB) Witness() *stateless.Witness { return nil }
 
 func (s *StateDB) AccessEvents() *ethstate.AccessEvents { return nil }
 
-func (s *StateDB) Finalise(deleteEmptyObjects bool) {}
+func (s *StateDB) Finalise(deleteEmptyObjects bool) *bal.StateAccessList { return nil }
+
+// Touch accesses the state without returning anything (EIP-7928 access list tracking).
```
arner: At fist sight, it may look like this should create a read dependency in our read/write set. I've done some research and I think we're safe. Can you to the comment:

// It doesn't affect the control flow, so we don't add a read dependency to the read/write set.

arner (commented):
Looks good, thanks! The comments were helpful for the review, but not really generic enough to stay in the codebase (very specific to this PR).

Example: _SelfDestruct signature changed in go-ethereum v1.17.3: no return value._

Is not relevant outside of the context of this PR. Can you remove the comments that talk about v1.17.3, or change them to be more "timeless"?

endorser/statedb.go
Storm1289: Done

endorser/executor.go
Storm1289: Done

arner:
Sorry for the late reply. The last commit seems to make it more complicated than it was... Maybe the way you did it the first time was actually better! I think you changed it because of a comment I made, but in hindsight, let's change it back :) -- as long as that doesn't have the same issue of the evm changing underlying data of pointers. After that I'll approve!

Storm1289:
Done, reverted CallMsgToMessage back to the original *big.Int + uint256.MustFromBig-at-return style. Each MustFromBig allocates a fresh *uint256.Int, so the resulting Message fields can't alias each other - no risk of the EVM mutating shared pointers.

arner (approved):
Thanks @Storm1289 , very nice!

arner:
@Storm1289 one of the legacy tests is failing, can you take a look?

Storm1289:
Okay

arner:
Ok, I did some research. To be honest I don't think we have to be compatible with Constantinople, so it's justified to skip these particular tests related to selfdestruct (with a comment). Let's discuss separately (out of scope for this PR) with @ale-linux how we want to deal with legacy tests vs the newer ones.

Storm1289:
added the 7 Constantinople SELFDESTRUCT/CREATE2 fixtures to testdata/eth_tests.slow with a one-line note. Will the broader legacy-test policy discussion happen on Discord, or in a separate issue?

arner:
We use the slow tests too. Maybe we should add an eth_tests.skip?

Storm1289:
Added loadSkip + filterSkippedTests in ethereum_test.go, and testdata/eth_tests.skip with the 12 unsupported paths (pre-Cancun legacy + Paris-suffix collision tests).
One small thing: testdata/* is ignored in .gitignore, so the new file needed a one-line exception (!testdata/eth_tests.skip).

arner:
@ale-linux , can you please take a look?

PR #201, opened 2026-06-18
Queue improvements

ale-linux:
This PR improves the throughput achieved by the queue by permitting transitive dependencies to proceed in parallel. We also re-add two makefile targets that have been removed erroneously.


arner (approved):
nice one!

PR #202, opened 2026-06-18
Explicit arma server dependencies in docker compose

tock-ibm:
Orderers now form per-party router->assembler->batcher chains, and every batcher depends on all 4 consenters; sidecar waits for coordinator + all party routers.


ale-linux (approved):

ale-linux:
Thanks @tock-ibm !

PR #204, opened 2026-06-18
Remove in-memory store and txcache

ale-linux:
The perf tests are getting notifications directly from the all-tx notification service so the in-memory store is no longer necessary.


arner (approved):
LGTM

PR #209, opened 2026-06-24
fix(gateway): map eth_call out-of-gas to -32000 instead of -32603

### LINKED ISSUE #208: eth_call out-of-gas returns -32603 (internal error) instead of -32000
## What's wrong

When an `eth_call` runs out of gas, the gateway returns JSON-RPC error code `-32603` (internal error) and leaks the endorser's internal status in the message:

```
error code -32603: query response was not successful, error code 500, msg out of gas
```

A normal `revert` is classified correctly as `-32000 execution reverted`, but out-of-gas — and other EVM gas-related failures such as intrinsic-gas-too-low — fall through to the generic internal-error path.

## Reproduction

Against a local network + gateway (commit `ad46228`):

```
# deploy the Counter sample, note its address
forge create --rpc-url http://localhost:8545 --broadcast solidity/Counter/Counter.sol:Counter ...

# call a view with a gas limit too low to finish -> out of gas
cast call <counter> 'getCount()(uint256)' --gas-limit 21500 --rpc-url http://localhost:8545
# -> error code -32603: query response was not successful, error code 500, msg out of gas

# for contrast, a genuine revert is mapped correctly
cast call <contract> '<reverting fn>' --rpc-url http://localhost:8545
# -> error code -32000: execution reverted: <reason>
```

## Expected vs actual

- **Expected:** out-of-gas is an execution outcome, so it should surface as a `-32000` execution error with a clean message — the same family as a revert (matches geth's behaviour).
- **Actual:** `-32603` internal error, leaking `error code 500` from the endorser query response.

## Root-cause pointer

`gateway/core/endorse.go` (~L129 / L142) wraps a non-OK endorser query response into a generic error: `query response was not successful, error code %d, msg %s`. Reverts carry revert data and are classified as `ExecutionReverted` (`-32000`) by `gateway/api/rpcerr`, but out-of-gas / intrinsic-gas errors aren't recognised and default to `-32603`.

Likely fix: detect EVM execution failures from the endorser response and classify them through the `gateway/api/rpcerr` patterns, consistent with how reverts are handled.

Related prior work: #142 (JSON-RPC error mapping), #166 (error docs).

arunabha003:
## Addressed issue
Closes #208

## Summary
- `CallContract` recognises go-ethereum EVM execution errors (out of gas, code-store OOG, max-code-size, invalid jump, write-protection, ...) and returns a typed `domain.ExecutionError` instead of a generic wrapped error.
- `classifyCallError` maps it to `-32000` (new `rpcerr.ExecutionError`) with a clean message and no revert data, mirroring the existing `domain.RevertError` handling. Detection is anchored to go-ethereum's `vm.Err*` sentinels.
- Out of scope (follow-up): pre-execution gas checks (`intrinsic gas too low` / floor-data-gas) still return `-32603`.

## Validation done
- `make checks` + `make unit-tests` pass; added unit tests for the API mapping and the `CallContract` classification (incl. a generic fault still -> `-32603`).
- End-to-end on a local Fabric-X network: out-of-gas `eth_call` now returns `-32000: out of gas` (was `-32603 ... error code 500`); a normal call still works.

### DIFF
--- a/gateway/api/rpcerrors.go
+++ b/gateway/api/rpcerrors.go
@@ -27,10 +27,9 @@ func classifyValidationError(err error) error {
 	return rpcerr.TxRejected(err)
 }
 
-// classifyCallError maps a Backend.CallContract error to a typed JSON-RPC
-// error. EVM reverts (*domain.RevertError) surface as -32000 with the raw
-// revert payload as ErrorData (geth eth_call contract); everything else is
-// an unexpected backend failure (-32603).
+// classifyCallError maps a Backend.CallContract error to a typed JSON-RPC error:
+// reverts (*domain.RevertError, with payload) and non-revert execution failures
+// (*domain.ExecutionError) surface as -32000; everything else is -32603.
 func classifyCallError(err error) error {
 	if err == nil {
 		return nil
@@ -38,6 +37,10 @@ func classifyCallError(err error) error {
 	var revert *domain.RevertError
 	if errors.As(err, &revert) {
 		return rpcerr.ExecutionReverted(revert.Reason, hexutil.Encode(revert.Data))
+	}
+	var exec *domain.ExecutionError
+	if errors.As(err, &exec) {
+		return rpcerr.ExecutionError(exec.Message)
 	}
 	return rpcerr.Internal(err)
 }
--- a/common/proposal.go
+++ b/common/proposal.go
@@ -20,6 +20,13 @@ const (
 	ProposalTypeState
 )
 
+const (
+	StatusOK             int32 = 200
+	StatusEVMRevert      int32 = 201
+	StatusEVMExecFailure int32 = 400
+	StatusError          int32 = 500
+)
+
 // StateQuery defines what to query from the ledger state, and at which snapshot.
 type StateQuery struct {
 	Account     common.Address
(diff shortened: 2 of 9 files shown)


gateway/api/rpcerrors.go
```diff
@@ -39,5 +38,9 @@ func classifyCallError(err error) error {
 	if errors.As(err, &revert) {
 		return rpcerr.ExecutionReverted(revert.Reason, hexutil.Encode(revert.Data))
 	}
+	var exec *domain.ExecutionError
```
arner, line 41: We can use something new here:

```
	if exec, ok := errors.AsType[*domain.ExecutionError](err); ok {
		return rpcerr.ExecutionError(exec.Message)
	}
```

There are some minor advantages, see https://antonz.org/go-1-26/#errors-astype

arner (changes requested):
Thank you for your contribution @arunabha003 ! 

I agree with the diagnosis, and this will improve our compatibility very nicely. We can make the method more efficient and easy to read, by using the status codes in our responses. The endorser already returns 200 for ok, 201 for reverts, and 500 for errors. We can add another error status code for execution failure:

```
const (
	statusOK             int32 = 200
	statusEVMRevert      int32 = 201 // success so we submit to orderers and get it committed.
	statusEVMExecFailure int32 = 400 // ran and failed (out of gas, etc.); return error to client without committing
	statusError          int32 = 500 // backend error
)
```

arunabha003:
Updated the PR based on your suggestion. All changes made, ready for a review @arner

arner (approved):
Thank you for your contribution, @arunabha003

PR #211, opened 2026-06-24
fix(gateway): populate blockTimestamp on eth_getLogs results

### LINKED ISSUE #210: eth_getLogs returns blockTimestamp: 0x0 (disagrees with eth_getBlockByNumber)
## Description
Every log returned by `eth_getLogs` has `blockTimestamp: 0x0`, even though `eth_getBlockByNumber` returns a real (node wall-clock) timestamp for the same block — so the two RPCs disagree about the same block.

## Reproduction
Against a local network + gateway (commit `ad46228`):

```
# deploy a contract, emit an event, then:
cast logs --from-block 0 --to-block 0xffff --address <contract> --rpc-url http://localhost:8545
#   -> each log shows blockTimestamp 0x0
cast block <that block number> --rpc-url http://localhost:8545 | grep timestamp
#   -> a real, non-zero timestamp
```

## Expected vs actual
- **Expected:** each log's `blockTimestamp` matches the block's timestamp (geth populates it; it should agree with `eth_getBlockByNumber`).
- **Actual:** always `0x0`.

## Root-cause pointer
`domain.Log` (`gateway/domain/models.go`) has no timestamp field, and `domainLogToTypesLog` (`gateway/api/eth.go` ~L474) never sets `types.Log.BlockTimestamp`. The value is available — the `blocks` table has a `timestamp` column — but `storage.GetLogs` doesn't read it. Fix: carry the block timestamp through `domain.Log` (join `logs`↔`blocks` in `GetLogs`) and set `BlockTimestamp` in `domainLogToTypesLog`.

arunabha003:
## Addressed issue
Closes #210

## Summary
- The `blocks` table already stores each block's timestamp, but `GetLogs` never read it. `GetLogs` now joins `logs`↔`blocks` and carries the value on `domain.Log.Timestamp`.
- `domainLogToTypesLog` sets `types.Log.BlockTimestamp` from it, so `eth_getLogs` agrees with `eth_getBlockByNumber`.

## Validation done
- `make checks` + `make unit-tests` pass; added unit tests (storage: `GetLogs` populates the timestamp; api: `domainLogToTypesLog` sets `BlockTimestamp`).
- End-to-end on a local Fabric-X network: a log's `blockTimestamp` now returns the block's real timestamp (`0x6a3c611c`, matching `eth_getBlockByNumber`); was `0x0`.

### DIFF
--- a/gateway/api/eth.go
+++ b/gateway/api/eth.go
@@ -478,14 +478,15 @@ func domainLogToTypesLog(l domain.Log) *types.Log {
 	}
 
 	return &types.Log{
-		Address:     common.BytesToAddress(l.Address),
-		Topics:      topics,
-		Data:        l.Data,
-		BlockNumber: l.BlockNumber,
-		BlockHash:   common.BytesToHash(l.BlockHash),
-		TxHash:      common.BytesToHash(l.TxHash),
-		TxIndex:     uint(l.TxIndex),
-		Index:       uint(l.LogIndex),
+		Address:        common.BytesToAddress(l.Address),
+		Topics:         topics,
+		Data:           l.Data,
+		BlockNumber:    l.BlockNumber,
+		BlockHash:      common.BytesToHash(l.BlockHash),
+		BlockTimestamp: uint64(l.Timestamp),
+		TxHash:         common.BytesToHash(l.TxHash),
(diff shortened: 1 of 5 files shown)


arner (approved):
LGTM, thanks @arunabha003 !

PR #190, opened 2026-06-02
Notification-based transaction completion system

ale-linux:
This PR introduces a notification-based architecture for tracking
transaction completion as an alternative to the traditional block-based
synchronization approach. The new system uses a pending transaction cache and
notification dispatcher to provide lower latency transaction tracking,
particularly beneficial for performance testing scenarios.

The Gateway now uses a BatchSubmitter for asynchronous endorsement submission
with optional notification support, and the integration test infrastructure
supports both traditional synchronization and notification-based modes. The
replay performance test demonstrates the new notification system with a
TxCompletionTracker for monitoring transaction completion, and includes a
hybrid mode that uses the synchronizer for initial catchup before switching to
notifications for ongoing updates.

This architecture maintains backward compatibility while enabling more
efficient transaction processing patterns for high-throughput scenarios.

Issue #136, opened 2026-04-30
Testing and Validation


ale-linux:
**Goal**: Ensure the implementation works correctly through comprehensive unit and integration tests, covering both happy paths and edge cases.

### Task 6.1: Unit Tests for TxQueue
**What to do**: Create comprehensive unit tests for all new TxQueue methods in `gateway/core/txqueue_test.go`.

**Details**:
- Test `IsPending()`:
  - Returns true for transaction in pending queue
  - Returns true for transaction in in-progress map
  - Returns false for transaction in neither
- Test `Dequeue()` moves transaction from pending to in-progress
- Test `MarkCompleted()` removes transaction from in-progress
- Test concurrent access:
  - Multiple goroutines calling `IsPending()` simultaneously
  - Enqueue and Dequeue happening concurrently
  - MarkCompleted called while IsPending is running
- Use Go's race detector: `go test -race`

**Why**: Unit tests catch bugs early and ensure thread-safety, which is critical for concurrent code.

### Task 6.2: Integration Test for Pending Transaction Flow
**What to do**: Create an end-to-end integration test that verifies the complete pending transaction lifecycle.

**Details**:
- Test scenario:
  1. Start Gateway and Chain
  2. Submit a transaction via `SendTransaction()`
  3. Immediately call `TransactionByHash()` - should return `isPending = true`
  4. Wait for block to be committed (may need to trigger block creation in test)
  5. Call `TransactionByHash()` again - should return `isPending = false` with transaction data
- Add assertions at each step
- Use test helpers to create and sign transactions
- Consider adding timing assertions (pending status should last at least X milliseconds)

**Why**: Integration tests verify that all components work together correctly, catching issues that unit tests miss.

### Task 6.3: Test Callback Mechanism
**What to do**: Create tests specifically for the callback notification system.

**Details**:
- Test that callback is invoked when block is committed
- Test that callback receives correct transaction hashes
- Test that transactions are removed from in-progress map after callback
- Test callback with multiple transactions in one block
- Test callback with empty block (no transactions)
- Test that callback errors don't break block processing (use a callback that panics)

**Why**: The callback mechanism is the critical link between Chain and Gateway. Bugs here would break pending status tracking.

### Task 6.4: Test Edge Cases
**What to do**: Create tests for all the edge cases identified in Phase 5.

**Details**:
- Test transaction failure during processing:
  - Mock `processTx()` to return an error
  - Verify transaction is removed from in-progress map
[...]

Issue #135, opened 2026-04-30
Handle Edge Cases and Error Scenarios


ale-linux:
*Goal**: Ensure the system behaves correctly in failure scenarios, prevents data corruption, and handles edge cases gracefully.

### Task 5.1: Handle Transaction Processing Failures
**What to do**: Decide what happens when `processTx()` fails and implement appropriate cleanup.

**Details**:
- In `worker()` method in `gateway/core/api.go`, when `processTx()` returns an error:
  - Here we shouldn't actually do anything yet - we should clearly document in the code using TODO and FIXME comments to highlight the error scenarios.

### Task 5.2: Prevent Duplicate Submissions
**What to do**: Add duplicate detection to `SendTransaction()` to prevent the same transaction from being queued multiple times.

**Details**:
- In `SendTransaction()` in `gateway/core/api.go`, before calling `g.txQueue.Enqueue(tx)`:
  - Check if transaction already exists: `if g.txQueue.IsPending(tx.Hash()) { return fmt.Errorf("transaction already pending") }`
- Consider whether this should be an error or silently succeed
- Add test case for duplicate submission
- Document the behavior in comments

**Why**: Duplicate submissions waste resources and could cause confusion. Ethereum clients typically don't resubmit the same transaction, but defensive programming is good.

### Task 5.3: Document Gateway Restart Behavior
**What to do**: Clearly document what happens to pending and in-progress transactions when the Gateway is restarted.

**Details**:
- Add comments to `TxQueue` explaining that the queue is in-memory only
- Document that on restart:
  - Pending queue is empty (transactions lost)
  - In-progress map is empty (transactions lost)
  - Clients will resubmit transactions if they don't see them confirmed
- Consider whether to add persistence in the future (probably not needed for MVP)
- Add a note in the README or architecture docs about this behavior

**Why**: Operators need to understand the implications of restarting the Gateway. Lost pending transactions are acceptable since clients handle resubmission.

### Task 5.4: Handle Callback Errors Gracefully
**What to do**: Ensure that errors in the callback don't break block processing in the Chain.

**Details**:
- In `Chain.Handle()`, wrap the callback invocation in error handling:
  ```go
  if c.txCompletionCallback != nil {
      defer func() {
          if r := recover(); r != nil {
              log.Printf("Transaction completion callback panicked: %v", r)
          }
      }()
      c.txCompletionCallback(txHashes)
  }
  ```
[...]

PR #212, opened 2026-06-26
gateway/core: reject duplicate tx submissions and document queue life…

Storm1289:
## Summary

Picks up the remaining relevant subtasks from #135 and #136. Parts of the original scope were already superseded by #176, #190, #201, and #204, so this PR implements only the items that still apply.

### 5.1 Drop failed transactions from the queue

`worker()` now drops a transaction from the queue when `processTx` returns an error.

`TxQueueInterface.Complete(common.Hash)` removes the failed transaction from queue tracking so `IsPending` no longer reports it as pending, allowing the client to safely resubmit the transaction.

Per maintainer direction, this PR intentionally **does not** introduce retry logic. Retry policies will be handled in a separate queue-retries issue.

### 5.2 Reject duplicate pending transactions

`Gateway.SendTransaction` now rejects a transaction whose hash is already queued or currently in progress.

Introduces:

- `domain.ErrTransactionAlreadyPending`

The existing `classifyValidationError` mapping surfaces this to JSON-RPC clients as:

```
-32003 Transaction rejected
```

### 5.3 Document queue persistence

`TxQueue` and `TxQueueV2` now explicitly document that the queue is **in-memory only**.

Pending and in-progress state is lost on gateway restart, so clients must resubmit any unconfirmed transactions after restart.

## Interface / API changes

- `TxQueueInterface` now exposes `Complete(common.Hash)`.
- `TxQueueV2.Complete` is promoted from the internal `completeUnlocked` helper to the public interface.
- `docs/JSON_RPC_ERRORS.md` documents the new `ErrTransactionAlreadyPending` sentinel.

## Tests

Added:

- `TestSendTransaction_DuplicateRejected` validates duplicate transaction rejection through `Gateway.SendTransaction`.
- `TestTxQueueV2_Complete_DirectRemovesFromPending` validates the newly public `Complete` method used by the drop-on-failure path.

Existing `TxQueue` and `TxQueueV2` tests continue to pass under `-race`.

## Decisions from review

- **5.4** (`recover` on callback): kept as-is. Not recovering is intentional at this stage of the project—panicking loudly on unexpected handler failures is preferred until v1 hardening. This can be revisited later.
- **6.3** (callback tests): skipped. The original callback model was replaced by the notification handler in #190, so the original subtask is no longer applicable. CI test-coverage tooling will be handled in a separate PR.

## Test plan

- [x] `go build ./...`
- [x] `go test -race -count=1 ./gateway/core/...`
- [x] CI Quick Tests
- [ ] CI Integration Tests

Refs #135, #136


Storm1289:
Hi @ale-linux @arner, please take a look when you have some time.

arner:
Hi, thanks for this @Storm1289 . 

- _5.1 Adds a TODO marker in worker() for the still-open drop-vs-retry question on processTx failure._ -> good point, we should handle that soon. Do you have ideas on how to do it?
- _5.4 (recover on callback)_ not recovering is intentional. In this stage of the project we want to painfully learn unexpected issues so we can resolve them (at least in the tests; when called from the API, the geth code will still recover and return a generic error). We'll remove it when getting closer to v1.
- _6.3 (callback tests): obsolete in the original form._ I'm ok with skipping. Not sure what the notification handler coverage is right now. What do you think of adding a tool to the CI that registers unit test coverage?

Storm1289:
Thanks @arner 

**5.1:** I'll drop a follow-up design comment on this thread with a proposal (rough sketch: classify processTx errors as permanent vs transient, drop permanent immediately, retry transient with bounded backoff + retry counter, dead-letter after N attempts). More detail once I've had a proper look at the current error taxonomy will post before the design discussion so it can feed into that.

**5.4:** Understood, I'll leave it as-is. I'll also add a note in the PR description that the current behavior is intentional.

**6.3:** great, skipping confirmed. On the CI test-coverage tool: happy to take it on. Would you prefer I fold it into this PR, or open a separate one for it?

arner:
5.1: let's keep it simple for now; just return the error without retries first. We have another issue open about retries for the queue.
5.4: ok
6.3: separate one.

arner (approved):
LGTM!

PR #214, opened 2026-06-30
fix(make): retry namespace setup in start-x to tolerate cold-start delays

### LINKED ISSUE #213: make start-x intermittently fails with "context deadline exceeded" on a cold start
## Description
`make start-x` **works almost every time** — this is an occasional, timing-dependent failure, not a consistent one. Once in a while, on a cold start, namespace creation fails with `Error: failed to wait for transaction status event: context deadline exceeded`, which aborts startup. It reproduces on a cold start (freshly pulled images / cold caches / loaded machine) but not on a warm one, so it's more of a reliability / CI-flakiness paper cut than a blocker.

## Reproduction
```
make stop-x
make start-x
#   Waiting for test committer to be ready...
#   Error: failed to wait for transaction status event: context deadline exceeded
#   make: *** [start-x] Error 1
```
Most likely right after `docker rmi` of the tools/orderer images, or under CPU/IO load. A clean retry (`make stop-x && make start-x`) succeeds.

## Expected vs actual
- **Expected:** `start-x` waits until the committer can actually commit, then reliably creates the namespace.
- **Actual:** it proceeds as soon as the TCP port is open and submits the namespace-create tx before the pipeline is fully ready, which occasionally times out.

## Root-cause pointer
`start-x` gates on `while ! nc -z localhost 7001`. But Docker publishes the host port the moment the container is created — before the orderer/committer processes inside are actually listening (7001/4001/7050 all accept connections within ~tens of ms of `compose up`), so it's a false "ready" signal. The subsequent `fxconfig namespace create --submit --wait` retries internally for a few seconds, which usually covers warmup, but a cold/slow start can exceed it. Notes: the container healthcheck is also just `nc -z localhost 7001` (so gating on health wouldn't help), and the sibling `start-full` target waits for both 7001 and 4001 while `start-x` waits only for 7001. Likely fix: replace the port poll with a real readiness check — e.g. retry `fxconfig namespace create` with backoff until it succeeds.

arunabha003:
## Addressed issue
Closes #213

## Summary
- `start-x` gated only on `nc -z localhost 7001`, but Docker publishes that port the moment the container starts — before the committer/orderer pipeline is actually ready. On a cold start the immediate `fxconfig namespace create --wait` could therefore time out (`context deadline exceeded`) and abort startup.
- Wrap the namespace `list-or-create` step in a bounded retry loop (5 attempts, 3s apart). The namespace command itself is unchanged (idempotent `namespace list | grep || namespace create`); the loop just retries on the cold-start timeout and fails loudly (`exit 1`) if all attempts fail, so it never silently succeeds without the namespace.
- No change to the readiness probe or the `start-full` target.

## Validation done
- Makefile-only change; `make -n start-x` parses cleanly and `make checks` passes.
- The namespace list/create invocation (image, env, volumes, flags) is identical to the current target — the change only wraps it in a retry-on-failure loop, leaving the success path unchanged.
- Caveat: the cold-start timeout is timing-dependent and doesn't reproduce reliably on a warm machine, so this is a best-effort robustness change without a before/after reproduction.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,18 +111,26 @@ start-x:
 	@$(COMPOSE) -f compose.fabric-x.yml up -d
 	@echo "Waiting for test committer to be ready..."
 	@while ! nc -z localhost 7001 2>/dev/null; do sleep 1; done
-	@$(DOCKER) run --rm --network $(NETWORK) \
-		--user "$(UID):$(GID)" \
-		--env "FX_NS=$(NS)" \
-		--env "FX_POLICY=$(POLICY)" \
-		-v "$(PWD)/testdata/fxconfig.yaml:/config/fxconfig.yaml:ro,Z" \
-		-v "$(PWD)/testdata/crypto/peerOrganizations/org1.example.com/peers/fxconfig.org1.example.com/tls:/tls:ro,Z" \
-		-v "$(PWD)/testdata/crypto/peerOrganizations/org1.example.com/users/channel_admin@org1.example.com/msp:/msp:ro,Z" \
-		-v "$(PWD)/testdata/crypto/peerOrganizations/org1.example.com/msp/tlscacerts/tlsca.org1.example.com-cert.pem:/org-tls-ca.pem:ro,Z" \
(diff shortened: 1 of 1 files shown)


arner (approved):
Thanks much @arunabha003 ! Good to have this more robust.

Issue #22, opened 2026-03-26
Implement gRPC Communication


ale-linux:
Implement gRPC-based communication between the Gateway and Endorser services to replace the current in-process function calls. Define a protobuf service specification for the endorsement protocol including methods for processing EVM transactions, executing read-only calls, and querying state. Implement the gRPC server in the endorser to handle incoming proposal requests and the gRPC client in the gateway to send proposals and collect responses. Ensure proper error handling, timeout configuration, and connection pooling for production reliability. Update integration tests to validate the gRPC communication layer works correctly across network boundaries.

---

**Quarter:** Q2

ale-linux:
@Storm1289 is working on this

PR #219, opened 2026-07-08
docs: add endorsement API design series

Storm1289:
Design docs for the custom gRPC endorsement API between gateway and endorser (#22).


Storm1289:
Hi @arner @ale-linux, please take a look on this pr 😊

docs/design/endorsement-api/01-api-and-proto.md
```diff
@@ -0,0 +1,220 @@
...
+  messages.
+
+## Service Definition
+
+A single service, `EvmEndorsement`, with one RPC per operation:
+
+- `ProcessTransaction(TransactionRequest) → peer.ProposalResponse`
+- `ProcessCall(CallRequest) → peer.ProposalResponse`
+- `ProcessStateQuery(StateQueryRequest) → peer.ProposalResponse`
+
```
arner, line 79: Maybe we should split them up more. Let's think out of the box a bit more, rather than sticking to the existing interface. We limited ourselves to fit in the strict peer.Proposal format, but after deciding we don't need that, we can make it a clean API the way we want it.

The real functions are more like (please check in real code -- it looks like we don't have an error):

```go
	Execute(tx *types.Transaction) (endorsement.ExecutionResult, error)
	Call(msg ethereum.CallMsg, blockNumber *big.Int) ([]byte, error)
	BalanceAt(account ethcommon.Address, blockNumber *big.Int) *big.Int
	StorageAt(account ethcommon.Address, key ethcommon.Hash, blockNumber *big.Int) []byte
	CodeAt(account ethcommon.Address, blockNumber *big.Int) []byte
	NonceAt(account ethcommon.Address, blockNumber *big.Int) uint64
```

And I think they deserve their own RPC endpoints, if only for type safety.

docs/design/endorsement-api/01-api-and-proto.md
```diff
@@ -0,0 +1,220 @@
...
+The service is **specialized for fabric-x-evm**, not the generic Fabric
+`ProcessProposal` endorsement API, per the design goal.
+
+## Messages
+
+**Response (all three):** reuse `peer.ProposalResponse` directly.
```
arner: Execute is the only one that needs an endorsement signature. See previous message, we can be more radical in the change.

arner (changes requested):
Thanks @Storm1289 ! I have to say I asked to split it up so that we could review and discuss just one part at at time... You pushed the whole thing at once 😅. And the decisions we make here will impact the rest of the documents too. I reviewed until my comments! Shall we keep it at the API and Proto one first? Which means, can you remove the other documents now and write (or re-push) them once we agreed on the API part? Got to go now, let's talk tomorrow!

docs/design/endorsement-api/01-api-and-proto.md
Storm1289: Agreed I've reworked the design to expose the six engine functions as individual typed gRPC methods and dropped the `peer.Proposal` framing entirely.

One thing I noticed while going through the code: all six methods on `EVMEngineInterface` return an error, so failures can map directly to gRPC status codes.

The remaining open question is what the `Execute` signature should look like now that there's no proposal on the wire. I've outlined two options in the design document:
- have the gateway build and sign a proposal hash, or
- sign the execution result directly.

It would be good to settle that before moving on to the implementation.

docs/design/endorsement-api/01-api-and-proto.md
Storm1289: Done only Execute returns an endorsement now; the five read RPCs return plain typed values with no proposal machinery.

Storm1289:
> Thanks @Storm1289 ! I have to say I asked to split it up so that we could review and discuss just one part at at time... You pushed the whole thing at once 😅. And the decisions we make here will impact the rest of the documents too. I reviewed until my comments! Shall we keep it at the API and Proto one first? Which means, can you remove the other documents now and write (or re-push) them once we agreed on the API part? Got to go now, let's talk tomorrow!

My bad 😅 I misunderstood your earlier suggestion. I thought you meant splitting the design into multiple documents and presenting the whole set at once. I've rework it into the format you had in mind.

docs/design/endorsement-api/01-api-and-proto.md
```diff
@@ -0,0 +1,265 @@
...
+(balance bytes, storage word, code bytes, nonce).
+
+## What Execute's Endorsement Covers
+
+`Execute` is the only RPC whose result feeds a Fabric transaction, so it is the
+only one that needs an endorsement signature. What exactly is signed - and how
+the gateway maps it into the envelope the orderer/committer expect - is the key
+open design point:
+
+- **Option A - gateway-built proposal, endorser signs against its hash.** The
```
arner: This is a good question! We should keep compatibility with Fabric-X and classic Fabric. Fabric-X is more relaxed in what must be there. If I remember correctly it does not require the proposal bytes at all for a valid transaction! And if it doesn't require it, we should leave it out (less bytes over the wire, less complexity). This is probably a small change in the SDK `fabricx.TxPackager`.

For Fabric the proposal is required as part of the submitted transaction. And the payload must include `ProposalHash`. So for Fabric, sending the hash over the API should be enough. And IMO we can leave it out for Fabric-X.

docs/design/endorsement-api/01-api-and-proto.md
```diff
@@ -0,0 +1,265 @@
...
+- **Read-write set:** the SDK's existing serialization of
+  `blocks.ReadWriteSet`, kept byte-exact for signing.
+
+## Unary vs Streaming
+
+**Unary RPCs for v1**, laid out so a streaming variant can be added without
+breaking changes. Unary matches the current one-shot call semantics and is
+easy to reason about. If the mempool throughput work (#50) shows per-call
+overhead matters, an `ExecuteStream` can be added alongside the unary method,
+the way the orderer keeps a connection open.
```
arner: Does this hold up if we target 15-20k tps? Either way, I agree with the approach that the first step should be unary, and then we optimize if necessary.

docs/design/endorsement-api/01-api-and-proto.md
```diff
@@ -0,0 +1,265 @@
...
+
+1. **Duplicate** the needed pieces from fabric-x-committer into fabric-x-evm.
+   *Pro:* no new dependency, full control. *Con:* drift over time.
+2. **Depend** on fabric-x-committer directly. *Pro:* single source of truth.
+   *Con:* couples fabric-x-evm to committer internals not meant as public API.
+3. **Upstream** the shared pieces into fabric-x-common, then depend on that.
+   *Pro:* clean shared home. *Con:* needs @senthil / maintainer buy-in and is
+   the slowest path.
+
+> Recommendation pending maintainer input: prefer (3) for anything genuinely
```
arner: Yes. Let's not depend on Fabric-X committer code. First stage, start with just fabric-x-common and basic GRPC packages. But can you already take a look what we miss? Possibly just a single package from the committer (see the custom endorser), to give us a lot of out of the box config for "free"?

docs/design/endorsement-api/01-api-and-proto.md
```diff
@@ -0,0 +1,265 @@
...
+> shared, fall back to (1) for small EVM-specific glue.
+
+## Proto Sketch
```
arner, line 192: Looks good at first sight!

arner (commented):
Thanks @Storm1289 , looking pretty good already! @ale-linux , can you take a look as well?

docs/design/endorsement-api/01-api-and-proto.md
Storm1289: Agreed.
From what I can see, the only pieces we'd still be missing from that base are the gRPC server bootstrap and config loading. From the `custom-endorser` sample you had me build, I already know this is handled by a single package: `fabric-x-committer/utils/serve`. It provides the server setup (listen, TLS, graceful shutdown, health checks) as well as config loading, while everything else remains `fabric-x-common` and standard gRPC.
If it helps, I can also verify that this package stays lightweight and doesn't pull in any heavy transitive committer dependencies.

docs/design/endorsement-api/01-api-and-proto.md
Storm1289: Makes sense ,let's go with that. I'll update the document accordingly once the full doc gets review by @ale-linux .

docs/design/endorsement-api/01-api-and-proto.md
Storm1289: gRPC runs on HTTP/2, so unary calls multiplex , thousands can be in flight over one pooled connection, it's not serial. At 15–20k tps the real cost is per-call overhead, not unary vs streaming; a streaming ExecuteStream would mainly trim that, and it's additive anyway. So I'd keep unary first, benchmark against the target, and add streaming only if it proves to be the bottleneck

docs/design/endorsement-api/02-errors-and-security.md
```diff
@@ -0,0 +1,150 @@
...
+- [Authentication and Authorization](#authentication-and-authorization)
+- [Resilience](#resilience)
+
+## Scope
+
+How the API represents and transports failures, and how the boundary is secured
+and kept resilient. Configuration mechanics (where certs and timeouts are set)
+are in part 3; this part defines *what* the semantics must be.
+
+## Error Model Today
```
arner: If you merge or rebase on main, you'll see that we changed this to be a bit more nuanced. The new code already prepares a bit better for the GRPC API, including returning every error as a response instead of a go error. That reserves real GRPC / connectiviity errors for the go error, and keeps the application errors separate.

docs/design/endorsement-api/02-errors-and-security.md
```diff
@@ -0,0 +1,150 @@
...
+gRPC gives us two independent channels; using both preserves the current
+two-outcome contract (`(response, nil)` vs `(nil, err)`):
+
+- **In-band typed response** (gRPC status `OK`): normal endorsement outcomes the
+  gateway must inspect. `Execute` carries the outcome in its `ExecuteResponse`
+  `status`/`message`/`payload` fields — success (200), revert (201), and
+  execution error (500) — the same triple the endorser produces today, now as
+  typed fields rather than a wrapped `peer.ProposalResponse`. The RPC succeeded;
+  the *result* carries the status.
+- **gRPC error status**: the endorser could not deliver a valid endorsement —
```
arner: See above, it's probably nicer if we return all application errors in the response. What do you think?

docs/design/endorsement-api/02-errors-and-security.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Error Mapping
+
+| Outcome | Endorser today | gRPC channel | Gateway result |
+|---------|----------------|--------------|----------------|
+| Success | status 200 | in-band `ExecuteResponse`, gRPC OK | payload |
+| Revert (`Execute`) | status 201 | in-band `ExecuteResponse`, gRPC OK | committed with `status=0` |
+| Revert (`Call`) | status 201 | gRPC error + revert details | `RevertError` → JSON-RPC `-32000` |
+| Execution error | status 500 | in-band `ExecuteResponse`, gRPC OK | generic error |
+| Pre-execution reject | `nil, err` | `INVALID_ARGUMENT` + details | typed error (nonce/gas/etc.) |
```
arner: See above.

docs/design/endorsement-api/02-errors-and-security.md
```diff
@@ -0,0 +1,150 @@
...
+## Transport Security: mTLS
+
+- **mTLS is required.** Both sides present certificates: the gateway
+  authenticates the endorser, and the endorser authenticates the gateway. No
+  plaintext or server-only-TLS fallback.
+- Reuse the existing TLS material and conventions already used for the endorser
+  ↔ committer connection (`PeerConf.TLSPath` in
+  [`endorser/config`](../../../endorser/config/config.go)); the new API adds its
+  own server-side listener credentials plus client credentials on the gateway.
+- Pin a modern TLS floor (TLS 1.2+/1.3) and support certificate rotation
```
arner: 1.3. Cert rotation without dropping sounds cool, but don't we have to restart the server anyway to rotate certificates?

docs/design/endorsement-api/02-errors-and-security.md
```diff
@@ -0,0 +1,150 @@
...
+
+> **Open question (maintainer):** reuse the endorser↔committer `PeerConf.TLSPath`
+> conventions for the new listener — and is hot certificate rotation required for
+> v1, or is drain-and-restart acceptable?
+
+## Authentication and Authorization
+
+- **Authentication:** the client identity is the mTLS peer certificate. The
+  endorser only accepts connections presenting a trusted client cert.
+- **Authorization:** restrict callers to the trusted gateway(s). Options: a CA
```
arner: For now, we can keep it at only mTLS; the endorser is equipped with a list of CA certificates of trusted organizations. Let's use the exact format of the committer (it's the ca-cert-paths):

```yaml
  tls:
    mode: mtls
    cert-path: /node-tls/server.crt
    key-path: /node-tls/server.key
    ca-cert-paths:
      - /orderer-cas/orderer-org-1/msp/tlscacerts/tlsca.orderer-org-1-cert.pem
      - /orderer-cas/orderer-org-2/msp/tlscacerts/tlsca.orderer-org-2-cert.pem
```

While some work is ongoing for a design with stricter authorizations for the committer (https://github.com/hyperledger/fabric-x-rfcs/pull/7), this can be layered on top later if we decide we need it.

docs/design/endorsement-api/02-errors-and-security.md
```diff
@@ -0,0 +1,150 @@
...
+  at two layers. Non-transient errors (`INVALID_ARGUMENT`, revert, execution
+  error) are not retried.
+- **Connection management:** long-lived, pooled connections with keepalive
+  between gateway and endorser rather than per-call dial cost.
+- **Backpressure:** bound in-flight requests / max concurrent streams so a
+  surge does not exhaust the endorser; surface overload as `RESOURCE_EXHAUSTED`
+  so the caller can back off.
+
+> **Open question (maintainer):** confirm retry lives entirely with the caller
+> (gateway / mempool #50) and the endorser stays stateless — no server-side retry
```
arner: confirmed

docs/design/endorsement-api/02-errors-and-security.md
```diff
@@ -0,0 +1,150 @@
...
+> or idempotency keys for v1? And should the endorser enforce per-client rate
+> limits / quotas, or is bounding in-flight requests + `RESOURCE_EXHAUSTED`
+> enough for v1?
```
arner: yes, enough for v1

docs/design/endorsement-api/02-errors-and-security.md
Storm1289: Agreed, rotating the cert restarts the listener anyway, so it's drain-and-restart now (simplified the line). On the version: we inherit tls.VersionTLS12 from the committer's serve package with 1.3 negotiated when available, so the floor stays 1.2 without us setting a separate policy.

docs/design/endorsement-api/02-errors-and-security.md
Storm1289: Agreed, that's cleaner. Updated the error sections to the #229 model: every application outcome rides in the response with a status code (200 OK, 201 revert, 400 tx-rejected, 460 exec-failure, 500 server-error), and the Go/gRPC error is reserved for transport/connectivity faults. Reworked Error Model, Two Channels, and the mapping table accordingly.

docs/design/endorsement-api/02-errors-and-security.md
Storm1289: Done kept it mTLS-only for v1 with the trusted org CA list in the committer's exact ca-cert-paths format, and linked fabric-x-rfcs#7 as a future layer.

docs/design/endorsement-api/02-errors-and-security.md
Storm1289: Recorded retry is caller-owned with the endorser stateless, and in-flight bounding + RESOURCE_EXHAUSTED for overload, both for v1

docs/design/endorsement-api/02-errors-and-security.md
Storm1289: Agreed done. All application outcomes now carry their status in the response; only transport/connectivity faults surface as a gRPC error. Two Channels section and mapping table updated.

docs/design/endorsement-api/03-client-config-testing.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Embedded vs Split Deployment
+
+- **Embedded (current):** top-level config carries `Endorsers []endorser.Endorser`
+  built in-process. Unchanged; the default.
+- **Split (new):** the gateway carries endorser **client** endpoints
+  (`[]common.ClientConfig`) and dials them; each endorser process runs the gRPC
+  server with its own `serve.Config` (endpoint + mTLS + keep-alive/limits).
+
+The two are mutually exclusive per gateway and chosen by which config block is
```
arner: Maybe we can be a bit more explicit about this. From a client point of view this makes sense, but at the moment we don't support standalone endorsers. So in practice all gateways will also be an endorser themselves.

A common occurrence that we have to support: Org A will get endorsements from Org A, Org B and Org C. Calling other endorsers is straightforward. The way I read your proposal, calling your own endorser would still go over GRPC (localhost) - even though it's running in the same process, did I get that right?

Without having thought through the full implementation, I'm on the fence on this one. In some of the discussions about preventing conflicts and improving performance, we've been thinking of running a first endorsement flow locally to determine the dirty reads and writes. This could help an endorser warm up storage cache like in Ethereums AccessLists, which could even be in the form of creating a first bulk query to the committer (if and when that feature is implemented in the committer). And on the other hand it would be input to build the dependency graph.

So we might like to revisit this part in the future. At the moment I think I'm ok with the GRPC loop to the same process. Its simplicity in implementation is good for now, despite the slight awkwardness. What do you think?

docs/design/endorsement-api/03-client-config-testing.md
```diff
@@ -0,0 +1,163 @@
...
+- **Resilience:** endorser-unavailable and deadline paths surface as the
+  expected retryable gRPC errors; confirm no double-retry against the mempool
+  layer (#50).
+- **Backward compatibility:** the embedded path keeps passing the existing
+  endorser and gateway suites unchanged.
+
+## Decisions and Alternatives
+
+**D3.1 - Implement the gRPC client as another `api.Service`.** Keeps
+`EndorsementClient` and the whole gateway above it untouched; `api.Service` is
```
arner: Yes, but don't be afraid to change the interface definition. It's not used in other places yet so we have the luxury to make it fit well.

docs/design/endorsement-api/03-client-config-testing.md
```diff
@@ -0,0 +1,163 @@
...
+the published contract the #229 restructure created for exactly this.
+*Alternative:* add a gRPC-aware layer in the gateway - rejected, needless churn
+behind an interface that already exists for this.
+
+**D3.2 - Reuse `common.ClientConfig` (client) and the committer `serve.Config`
+(server).** The client endpoints match orderers/committer; the server config
+comes from `fabric-x-committer/utils/serve`, giving mTLS, keep-alive, and
```
arner: This is our preliminary decision, but I'm still not too happy to start to depend on fabric-x-commiter. Let's discuss again before the real implementation. It does not change the rest of the plan so let's just finish the design first.

arner (commented):
Hi @Storm1289 , good additions again. When you create the plan, can you please make sure it's modular and in steps so that the codebase keeps working on every PR, and we don't implement it as a huge big bang? Thanks!

docs/design/endorsement-api/03-client-config-testing.md
arner: https://github.com/hyperledger/fabric-x-committer/issues/675

arner: @Storm1289 , thanks to the fast reply and cooperation of the committer team we will be able to depend on fabric-x-common only! Can you please rewrite the parts about relying on fabric-x-committer to fabric-x-common, anticipating the move of the `utils/serve` package?

Storm1289: I really appreciate that! I had actually already been working on the changes for the last couple of hours 😅, so I'll rework everything once more based on your feedback. I think it'll turn out much better this way.

Storm1289:
Hi @arner, thanks a lot for the detailed review! 🙏 I've reworked the design based on your feedback, updated the relevant parts, and added Part 4 (implementation plan).
I think the design is much clearer now. Whenever you have some time, could you please take another look? Thanks again!

arner (approved):
Great work @Storm1289 , ready to go!

arner:
Merging the plan; @Storm1289 and I agreed that as a last step of the implementation, we will convert the plan format to normal documentation about the endorsement API.

PR #227, opened 2026-07-09
Support out-of-order (future-nonce) transaction submission

### LINKED ISSUE #52: Support Out-of-Order Transaction Submission
Implement support in the mempool for accepting transactions with nonces higher than the current committed nonce, parking them in the queue until prerequisite transactions with lower nonces commit successfully. Ethereum clients often submit multiple transactions in rapid succession with sequential nonces, expecting the network to process them in order. The mempool should track each sender's nonce state, accept transactions with future nonces, and hold them in a pending state. When a transaction commits and advances a sender's nonce, check if any queued transactions for that sender can now be processed. Implement proper ordering within the queue to ensure transactions are processed in nonce order per sender. This enables efficient transaction batching from clients without requiring them to wait for each transaction to commit before submitting the next.

---

**Quarter:** Q2

Storm1289:
## Summary

Closes #52.

Adds per-sender nonce parking so Ethereum clients can submit out-of-order future-nonce transactions. Clients commonly send several transactions back-to-back with sequential nonces and expect them to be processed in order. Previously, the gateway enqueued every validated transaction immediately, so a future-nonce transaction could fail downstream. This PR holds future-nonce transactions until the earlier nonces commit, then releases them in nonce order.

## Design

The gateway owns a per-sender `nonceGate` that caches each sender's next expected nonce and is updated from committed blocks rather than performing a state read for every submission.

- **Admit** (from `SendTransaction`)
  - Steady state: the sender is already cached, so admission performs no state read.
  - First sight: seed the sender's committed nonce from state once, then trust the cache. The seed happens off-lock, followed by a recheck, so a slow store read does not block other senders.
  - `nonce == next` → enqueue
  - `nonce > next` → park
  - `nonce < next` → reject as nonce-too-low

- **Observe** (on each committed block)
  - Advances each committed sender's cached nonce and releases the now-ready parked transaction in nonce order.
  - Only Fabric-valid commits advance a nonce; an invalidated commit does not consume it.

This keeps the hot path free of store I/O once a sender is known.

## Snapshot Reverts / Primed State

The production gate reads a sender's nonce once and then trusts its cache. The test backend, however, can move the ledger nonce out of band through snapshot reverts and primed state.

A test-only `reconcilingGate` re-reads the committed nonce before each admission. It lives in `gateway/testimpl/reconciling_gate.go` and is injected at construction through the `core.WithNonceSequencer` option. Production never wraps the gate.

## Files

- `gateway/core/nonce_parking.go` - new nonce gate.
- `gateway/core/api.go` - routes `SendTransaction` through the gate, calls `Observe` on commit, includes parked transactions in pending lookups, and adds a constructor option for the test wrapper.
- `gateway/core/validate.go` - moves the nonce check out of validation and into the gate, which now owns sequencing.
- `gateway/testimpl/reconciling_gate.go` - new test-only reconcile wrapper.
- `gateway/app`, `integration` - wire the test wrapper for the test backend.

## Tests

`gateway/core/nonce_parking_test.go` covers:

- In-order admission and no-read steady state
- Future-nonce parking and release on commit
- One-release-per-commit chaining
- Multiple independent senders
- Per-sender parked cap
- Reconciliation after an out-of-band nonce change and after a revert
- TTL eviction of orphaned parked transactions
- Persisting an unknown committed sender
- Skipping an invalidated commit
- Seed-error propagation

All tests pass with `-race`, the full `gateway/core` suite is green, and `make checks` is clean.

## Test plan

- [x] `go test -race ./gateway/core/...`
- [x] `make checks`
- [x] CI Quick Tests / Ethereum Tests / OZ Hardhat Compatibility
- [ ] CI Integration Tests


Storm1289:
Hi @ale-linux, whenever you have some time after reviewing #219, could you please take a look at this one as well?

arner:
Hi @Storm1289 , `make checks` is failing, can you take a look please?

Storm1289:
> Hi @Storm1289 , `make checks` is failing, can you take a look please?

fixed, please check again.

Storm1289:
Hi @ale-linux, please review this pr whenever you have sometime.

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,185 @@
...
+
+// nonceGate holds transactions whose nonce is ahead of the sender's committed
+// nonce and admits them to the worker queue in nonce order as earlier nonces
+// commit. It keeps a single transaction per sender in flight (issue #52);
+// cross-sender dependency scheduling is handled separately (issue #59).
+type nonceGate struct {
+	mu     sync.Mutex
+	state  stateReader
+	signer types.Signer
+	admit  func(*types.Transaction) // hands an in-order tx to the worker queue
```
ale-linux: let's please not use function pointers - we can pass in the pointer to the queue (or hide it behind an interface)

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return fmt.Errorf("recover sender: %w", err)
+	}
+
+	committed, err := g.state.NonceAt(ctx, from, nil)
+	if err != nil {
+		return fmt.Errorf("look up nonce: %w", err)
+	}
+
+	g.mu.Lock()
```
ale-linux, line 101: does the lock need to gate in-order submissions handled in lines 90-93?

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,185 @@
...
+		return nil
+	}
+
+	// Future nonce: park until the gap fills.
+	sp := g.senders[from]
+	if sp == nil {
+		sp = &senderParking{parked: make(map[uint64]parkedTx)}
+		g.senders[from] = sp
+	}
+	sp.evictExpired(g.now(), g.ttl)
```
ale-linux: do we need to call this here as well?

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,185 @@
...
+
+	if _, replacing := sp.parked[tx.Nonce()]; !replacing && len(sp.parked) >= g.maxPerSender {
```
ale-linux: this seems buggy - we modify sp.parked and yet return an error..

gateway/core/api.go
```diff
@@ -398,6 +408,10 @@ func (g *Gateway) Stop() error {
...
 	// Convert blocks.Block to domain.Block using the shared conversion function
 	domainBlock := ConvertToDomain(b)
-	err := g.TxQueue.Handle(ctx, &domainBlock)
-	return err
+	if err := g.TxQueue.Handle(ctx, &domainBlock); err != nil {
+		return err
+	}
+	// Release parked transactions whose earlier nonces just committed.
+	g.nonceGate.Released(ctx, domainBlock.Transactions)
+	return nil
```
ale-linux: This code appears backward to me: 

- `TxQueue.Handle` is responsible for feeding workers
- `nonceGate.Released` releases queued work

Wouldn't it make more sense to first release new work and then call the scheduler?

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,185 @@
...
+func (g *nonceGate) Released(ctx context.Context, committed []domain.Transaction) {
+	g.mu.Lock()
+	affected := make(map[common.Address]struct{})
+	for i := range committed {
+		from := common.BytesToAddress(committed[i].FromAddress)
+		if _, tracked := g.senders[from]; tracked {
+			affected[from] = struct{}{}
+		}
+	}
+	g.mu.Unlock()
```
ale-linux: By taking, releasing and retaking the mutex we're exposing the code to races and possible incorrect behaviour... I understand the desire to release the mutex for the slow call to `NonceAt` but I'm afraid it might be dangerous. I'm thinking we should maybe build our local nonce cache by registering to `Handle`, processing txes and building our user/nonce map?

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,185 @@
...
+	defer g.mu.Unlock()
+
+	sp := g.senders[from]
+	if sp == nil {
+		return
+	}
+	sp.evictExpired(g.now(), g.ttl)
+
+	for nonce := range sp.parked {
+		if nonce < committed {
```
ale-linux: How/why would this happen?

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	if len(sp.parked) == 0 {
+		delete(g.senders, from)
+	}
+}
+
+// IsPending returns a parked transaction by hash, or nil if it is not parked.
+func (g *nonceGate) IsPending(hash common.Hash) *types.Transaction {
+	g.mu.Lock()
```
ale-linux: Shouldn't we use a readlock here?

ale-linux (changes requested):
Great start, thanks! I feel there are a few issues, which might be solved if we don't have to access storage to get the nonce - I feel we have to keep our own nonce data structure, at which point we can be the implementation of `stateReader` for validate.go

gateway/core/nonce_parking.go
Storm1289: Done

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,207 @@
...
+	// restart, a missed commit). Re-read once before treating it as a gap.
+	if !seeded && tx.Nonce() > ss.next {
+		committed, err := g.state.NonceAt(ctx, from, nil)
+		if err != nil {
+			return fmt.Errorf("look up nonce: %w", err)
+		}
+		if committed > ss.next {
+			ss.next = committed
+		}
+	}
```
ale-linux: If we hook through `Handle` on all commit events, we can keep a complete nonce map - can we do it that way?

ale-linux: plus, this retry loop is brittle, so even if we go this route and query, let's try to avoid this

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,207 @@
...
+	ss.parked[tx.Nonce()] = parkedTx{tx: tx, parkedAt: g.now()}
+	return nil
+}
+
+// Observe advances each committed sender's next nonce from the block and admits
+// the now-ready parked transaction, if any. It does no state reads.
+func (g *nonceGate) Observe(committed []domain.Transaction) {
+	// Highest committed nonce per sender.
+	highest := make(map[common.Address]uint64)
+	seen := make(map[common.Address]bool)
```
ale-linux: we could get rid of `seen` and go

```go
if cur, ok := highest[from]; !ok || tx.Nonce() > cur {
		highest[from] = tx.Nonce()
	}
```

Storm1289:
@ale-linux, please take a look if you have sometime.

gateway/core/api.go
Storm1289: Agreed, reordered in the rework: Handle now calls nonceGate.Observe(...) (releasing parked work whose earlier nonce just committed) before TxQueue.Handle(...) feeds the workers. Released became Observe and updates the nonce map straight from the committed block, so this path does no state read either.

gateway/core/nonce_parking.go
ale-linux: this is redundant.

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,207 @@
...
+				return p.tx
+			}
+		}
+	}
+	return nil
+}
+
+// evictIdle drops cached senders with no parked txs that have been idle past the
+// TTL, bounding the map for one-shot senders.
+func (g *nonceGate) evictIdle(now time.Time) {
```
ale-linux: Evicting idle nonces is safe (again, under the assumption that gw and endorser are in the same process) for what explained before. I would almost never evict seeds though, because going to the store must happen under the lock, which blocks everything... Let's have - as default - that we cache at least 1M nonces and only evict if we have more than those, LRU

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,207 @@
...
+	// Highest committed nonce per sender.
+	highest := make(map[common.Address]uint64)
+	seen := make(map[common.Address]bool)
+	for i := range committed {
+		tx := committed[i].ToEthTx()
+		if tx == nil {
+			continue
+		}
+		from := common.BytesToAddress(committed[i].FromAddress)
+		if n := tx.Nonce(); !seen[from] || n > highest[from] {
```
ale-linux: We must verify that the transaction is valid before bumping the nonce

ale-linux (changes requested):
To ensure the correctness and simpliticy of the code, I propose that we change the code and implement these two points;

- `Observe` must persist ANY AND ALL nonces it receives (from correct ethereum txes that committed successfully of course) - it must create a `senderState` entry and track the next nonce.
- `Admit` will probe `state.NonceAt` from the endorser ONLY ONCE whenever no `senderState` entry exists in the local cache; this will only happen if
    - the sender has never transacted
    - the sender has transacted at a time when the gateway was down, or before the last gateway restart
    
These two things together will ensrure that the nonce parking is always correct. There are indeed two events that influence the correct processing of an incoming tx:

1) when `Observe` is called to register that the previous nonce for that tx's sender has just committed;
2) when we call `Admit` on that tx

These are the possible cases:

- we receive event 2 but event 1 has never happened, because that's the sender's first tx: then we'll go to storage, find out that the nonce is zero, will cache that and will process the tx immediately
- we receive event 2 and event 1 has happened before we started; then we'll go to storage, prime the local cache and will process the tx immediately
- we receive event 1 first, and then event 2; that's fine because then we are caching the correct nonce locally and we can immeidately process thx tx because we immeidately recognise that it is correct
- we receive event 2 first and then event 1: that's also fine because we'll first queue the tx and then event 1 will unblock it and schedule it

There can never be a race between the two events s.t. we are missing a trigger and the tx is never fired: if Admit is called at the same time of Observe, the lock will serialise the two events: 

- if admit wins and there's nothing cached, we will go to the endorser; the endorser may or may not have already have received notification that the previous nonce has committed
    - if it has, it will tell us, we will cache that info and schedule the tx immediately
    - if it hasn't yet, it will give us a lower nonce, we'll add the tx to pending and then we'll rely on the notification from `Observe` to unblock it
- if Observe wins, it's easy because we'll create the local senderState with the expected nonce and then later admit will schedule immediately without going to storage

Note that this works if the endorser and the gateway are the same process - i.e. if the endorser maintains its nonce store using the same stream that feeds `Observe`. If it's not the case, then we may have a deadlock in the following way:

- gateway processes Alice's tx with nonce 0 (i.e. Alice's very first tx), commits the block ok, crashes; once it restarts, it will only receives updates _after_ that block, i.e. no `Observe` for that transaction will ever be fired on itself; the nonce on the ledger is 1 by now, and Alice will know it and send her next tx with nonce 1
- gateway has no local endorser and so it uses a remote endorser as state provider for the nonce
- the remote endorser is lagging behind and has not yet seen Alice's tx with nonce 0, so it'll report Alice's nonce still to 0
- the gateway admits Alice's tx with nonce 1 which should be scheduled; the gw has no local cache for Alice (and will never get one through Observe now unless we schedule thix tx); so the gateway will contact the endorser; the endorser will report what it knows, which is nonce 0. Eventually it will catch up and reach nonce 1, but it won't tell it to us, and we won't ever learn it from observe. At this point alice's tx is scheduled never to be worked on.

Once again, this deadlock is only present if the gateway doesn't have a built-in endorser, otherwise it can never happen --> @arner , this is a good reason to never break those two up ;)

Storm1289:
Thanks @ale-linux! Pushed the rework addressing your remaining points and I'm happy with the direction:

- **Read-once `Admit`** - the gate seeds a sender's nonce from state once, then trusts the cache (advanced only by `Observe`). The in-gate reconcile-on-mismatch is gone from the production path.
- **Test-only reconcile** - the revert / primed-state re-read moved into a `reconcilingGate` wrapper, switched on via `UseTestNonceReconcile()` for the test backend only (testnode + integration harness), so production stays read-once.
- **LRU eviction** - the sender cache is now bounded by a large-limit LRU (`defaultMaxSenders`) instead of a TTL, and never evicts a sender that still has parked txs.
- **`Observe`** - persists every committed sender (even one it never admitted for) and only advances a nonce on Fabric-valid commits.


Whenever you get a chance, please take a look and let me know if this matches what you had in mind - happy to adjust anything.

gateway/core/api.go
```diff
@@ -127,7 +128,20 @@ func New(ec *EndorsementClient, batchSubmitter *BatchSubmitter, store Store, cha
...
+	return g, nil
+}
+
+// UseTestNonceReconcile makes the nonce gate re-read the ledger before each
+// admit, so it tracks the snapshot reverts and primed state used by the test
+// backend. It must not be called on a production gateway.
+func (g *Gateway) UseTestNonceReconcile() {
+	if ng, ok := g.nonceGate.(*nonceGate); ok {
+		g.nonceGate = reconcilingGate{ng}
+	}
```
ale-linux: Can we inject this in the constructor as opposed to overriding it here?

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,255 @@
...
+	return nil
+}
+
+// resync updates the cache to the sender's committed nonce.
+func (g *nonceGate) resync(ctx context.Context, tx *types.Transaction) error {
+	from, err := types.Sender(g.signer, tx)
+	if err != nil {
+		return fmt.Errorf("recover sender: %w", err)
+	}
+	committed, err := g.state.NonceAt(ctx, from, nil)
```
ale-linux, line 153: we're consulting `state` without acquiring a lock..

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,255 @@
...
+type reconcilingGate struct {
+	*nonceGate
+}
+
+func (g reconcilingGate) Admit(ctx context.Context, tx *types.Transaction) error {
+	if err := g.nonceGate.resync(ctx, tx); err != nil {
+		return err
+	}
+	return g.nonceGate.Admit(ctx, tx)
+}
```
ale-linux: IIUC the reconciling gate is used only for tests - can you park it under ./testimpl/reconciling_gate.go?

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,255 @@
...
+
+	for _, ss := range g.senders {
+		for _, p := range ss.parked {
+			if p.tx.Hash() == hash {
+				return p.tx
+			}
+		}
+	}
+	return nil
+}
```
ale-linux, line 257: This can be potentially very slow. We need an auxiliary map of tx hash -> bool for fast lookups by hash. It needs to be maintained together with the senders/parked..

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,255 @@
...
+
+// evictLRU drops least-recently-seen senders with no parked txs while over the cap.
+func (g *nonceGate) evictLRU() {
```
ale-linux, line 260: we don't evict senders with parked txes, so that's potentially a DoS... we should create an issue to eventually start removing parked transactions after a while

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,255 @@
...
+			return fmt.Errorf("look up nonce: %w", err)
+		}
+		ss = &senderState{next: seed, parked: make(map[uint64]parkedTx)}
+		g.senders[from] = ss
+		seeded = true
+	}
+	ss.lastSeen = g.now()
+	if seeded {
+		g.evictLRU()
+	}
```
ale-linux: Here we have an interesting issue: it's possible that `evictLRU` removes the very `senderState` we just created. We must ensure against this: we evict then insert.

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,255 @@
...
+
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	ss := g.senders[from]
+	if ss == nil {
+		ss = &senderState{parked: make(map[uint64]parkedTx)}
+		g.senders[from] = ss
+	}
+	ss.next = committed
```
ale-linux, line 165: Shouldn't we remove parked txes with lower nonces?

integration/test_helpers.go
```diff
@@ -173,6 +173,8 @@ func defaultHandlerChain(t *testing.T, ctx context.Context, cfg config.Config, e
 	if err != nil {
 		t.Fatalf("build gateway: %v", err)
 	}
+	// Tests prime and revert ledger state out of band; track it before each admit.
+	gw.UseTestNonceReconcile()
```
ale-linux: Can we use this _only_ for the tests that require it?

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,255 @@
...
+	case tx.Nonce() < ss.next:
+		return fmt.Errorf("%w: next nonce %d, tx nonce %d", ethcore.ErrNonceTooLow, ss.next, tx.Nonce())
+	case tx.Nonce() == ss.next:
+		g.queue.Enqueue(tx)
+		return nil
+	}
+
+	// Future nonce: park until the gap fills.
+	ss.evictExpired(g.now(), g.ttl)
+	if _, replacing := ss.parked[tx.Nonce()]; !replacing && len(ss.parked) >= g.maxPerSender {
```
ale-linux: here we're silently replacing but replacing in ethereum should happen under specific rules.. I think we have a story open to handle those, could you please

1. add a comment here referring to the issue
2. log that the replacement took place

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
+	ss := g.senders[from]
+	seeded := false
+	if ss == nil {
+		// First sight: seed the nonce from state.
+		seed, err := g.state.NonceAt(ctx, from, nil)
```
ale-linux, line 108: Sorry I keep circling back to this, I know I okayed this but on second thought I don't think we should seed the nonce from state while holding the lock. Let's rather

1. hold the lock and see whether we have all data cached (that'll be the steady state very quickly)
2. if not, release the lock, seed from store
3. re-acquire the lock, recheck (we might have hit an `Observe` which invalidates our seeding)

gateway/core/nonce_parking.go
ale-linux: it might be fine, pls double-check

ale-linux (changes requested):
Thanks a lot @Storm1289 , great stuff! A couple more comments on the final version

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,255 @@
...
+	return g.nonceGate.Admit(ctx, tx)
+}
+
+// Observe advances each sender's cached nonce from the block and releases any
+// now-ready parked transaction. Only Fabric-valid commits advance a nonce.
+func (g *nonceGate) Observe(committed []domain.Transaction) {
+	// Highest valid nonce per sender.
+	highest := make(map[common.Address]uint64)
+	for i := range committed {
+		if committed[i].FabricTxStatus != 0 {
```
arner: The FabricTxStatus is not reliable as a check. `peer.TxValidationCode_VALID` (classic fabric) = 0, while `committerpb.Status_COMMITTED` = 1. Tbh I don't think our current domain model accurately reflects whether a transaction is committed or not. 

Actually it's worse - while researching this I noticed that the SDK's Transaction.Status (where this value is derived from) is not even set! So it will always be 0. You can't use the ethereum status here either, because a revert still bumps the nonce.

So ConvertToDomain is doing something useless (storing the fabric status which is empty because the SDK doesn't set it), and your check is also not working because we don't know whether it should be 0 or 1. I'm going to look for a good fix for the SDK. Do you have a suggestion for what we should do in fabric-x-evm?

Storm1289: Good catch, and agreed, FabricTxStatus comes from tx.Status, which is never set, so the check is a no-op right now. I think we already have the right signal in ConvertToDomain: tx.Valid. It's what we use to derive the EVM receipt status, and it separates the two cases the nonce cares about - a revert has tx.Valid == true (nonce consumed → advance), an MVCC-invalid tx has tx.Valid == false (nonce not consumed → skip). So for fabric-x-evm I'd switch the check from FabricTxStatus to tx.Valid and drop the dead tx.Status plumbing. This feels like the same family as #312 / #257 - execution/commit outcome not fully surviving into the domain model - so I'll open an issue and link it to those (and to your SDK fix). Does tx.Valid hold up on the committer path from your side?

Storm1289:
Thanks @ale-linux! Pushed `4abecb1` (+ `79b567a`) addressing your review. Quick rundown:

**Done**
- `Admit` now seeds a sender off the lock - hold the lock, on a miss release it, seed from the store, then re-acquire and recheck (an `Observe` may have created the sender meanwhile). Steady state stays lock-free with no store read.
- Evict-then-insert, so LRU can never drop the just-created `senderState`.
- `Resync` drops parked txs whose nonce is below the re-read committed nonce.
- `IsPending` is O(1) via a by-hash index kept in sync by `park`/`unpark` helpers.
- Moved the reconciling gate to `gateway/testimpl/reconciling_gate.go` and injected it at construction via a new `WithNonceSequencer` option - the `UseTestNonceReconcile` override is gone.
- Same-nonce parked replacement now logs, and the comment references the tracking issue #62.
- Opened #347 for the parked-transaction memory follow-up you flagged.

**Two things I'd like your call on**

1. **Scope of the test reconcile** (`test_helpers.go`): I routed it through the constructor option, but kept it applied to the whole integration harness rather than per-test. Reconcile only adds a re-read before admit, so it's never wrong, and scoping it to just the tests that prime/revert risks silently breaking those. Happy to make it a per-test opt-in if you'd prefer.

2. **`Resync` reading state off-lock**: intentional - it reads `NonceAt` off the lock, then takes the lock only to apply the result, the same pattern as the seed path. The store read doesn't touch shared state, so holding the lock across it would just block other senders (and it's a test-only path now). Flagging it so you can confirm you're happy with it.

Storm1289:
Small reminder on the red check: the only failure is `TestFabricX/two_of_two_endorsement_policy` - everything else passes and main is green on it. It's the split-deployment case - the gateway reads the sender's nonce from a remote endorser that can lag, parks the tx, and it never commits (the test times out). That's the deadlock flagged in the review, and my understanding is it's the gateway/endorser sync work @arner is already on - so I'd wait for that to land and re-run.

gateway/core/api.go
ale-linux: A step in the right direction, but can we actually go for simply passing in a `NonceSequencer` argument to the constructor? If you receive nil just create the normal one; the one test that needs the wrapped one will create the normal one and then wrap it and pass it in

integration/test_helpers.go
```diff
@@ -169,7 +170,8 @@ func defaultHandlerChain(t *testing.T, ctx context.Context, cfg config.Config, e
 	if cfg.Network.Namespace == "synthetic" {
 		txPerSec = 10000
 	}
-	gw, err := app.BuildGateway(ctx, ends, gwSigner, cfg.Network, chain, submitters, cfg.Gateway.SubmitterCount, cfg.Gateway.WorkerCount, txQueue, cfg.Gateway.EndorsementChanSize, txPerSec)
+	// Tests prime and revert ledger state out of band; reconcile before each admit.
+	gw, err := app.BuildGateway(ctx, ends, gwSigner, cfg.Network, chain, submitters, cfg.Gateway.SubmitterCount, cfg.Gateway.WorkerCount, txQueue, cfg.Gateway.EndorsementChanSize, txPerSec, core.WithNonceSequencer(testimpl.NewReconcilingGate))
```
ale-linux, line 174: Why can't we just use the standard nonce parking infrastructure?

ale-linux (changes requested):
Thanks @Storm1289 - I left you two more comments. Concerning the test that fails: it shouldn't, at least not for the reasons you state. In a previous PR I added

https://github.com/hyperledger/fabric-x-evm/blob/main/integration/integration_test.go#L226-L236

to explicitly guard against the lack of sync between gateway and endorsers. Can you please debug this in greater depth to find out what is going on exactly?

Storm1289:
Thanks @ale-linux! Digging into the two_of_two failure with your waitForReadEndorser point, here's what I think is happening:

The two transfers are serialized by nonce - the second can't go until the first commits. The first does commit, and the endorser knows it (waitForReadEndorser guarantees that). But the nonce gate no longer reads the endorser per tx - it caches the nonce and advances it only from its own Observe signal, which lags. So when the second tx is admitted, the endorser already says "nonce 1" but the gate's cache still says "0", so it parks the second tx and waits for a commit that already happened -> timeout.

So you're right, it's not the endorser lagging - the endorser is fine. The gap is that the gate's cache and the committed state are updated by two different signals that can drift, so your guard (which fixes the endorser read) no longer covers the cache. On main this couldn't happen because the nonce was read fresh from the endorser every time.

Fix direction I'd propose: the nonce gate should advance from the same "committed" signal the rest of the system uses, so its view and the committed state can't drift apart. I'll dig into the exact sequence with logging to confirm that's the root cause and settle on the cleanest fix.

On your other two comments:
- Standard nonce parking: agreed that's the goal. The reconciling wrapper only exists for the test backend's out-of-band state (snapshot reverts / primed nonces); for normal traffic it isn't needed, and once the drift is fixed, standard parking should work everywhere and the wrapper can go.
- Passing the sequencer into the constructor: one wrinkle - the gate uses the Gateway itself as its state source, so it can't be fully built before New returns, which is why I used a wrapper callback. I'll switch to a plain NonceSequencer arg (nil -> default) if we keep a wrapper at all, which depends on the above.

I'm happy to take this on and figure out exactly what's going on. Since it's a gateway/endorser sync interaction broader than the nonce parking, and this PR is already fairly large, I'd suggest I do the debug + fix in a follow-up PR (with two_of_two marked a known-failure here in the meantime) so this one stays focused - but happy to resolve it here instead if you'd prefer. Let me know.

gateway/core/nonce_parking.go
arner: Valid is set correctly (both for Fabric and Fabric-X). I'll work on the SDK fix separately.

ale-linux: That's actually the issue with the failing test!

```
diff --git a/gateway/core/nonce_parking.go b/gateway/core/nonce_parking.go
index 4eec88c..c56f941 100644
--- a/gateway/core/nonce_parking.go
+++ b/gateway/core/nonce_parking.go
@@ -178,7 +178,7 @@ func (g *nonceGate) Observe(committed []domain.Transaction) {
        // Highest valid nonce per sender.
        highest := make(map[common.Address]uint64)
        for i := range committed {
-               if committed[i].FabricTxStatus != 0 {
+               if committed[i].Status != 1 {
                        continue // invalidated: nonce not consumed
                }
                tx := committed[i].ToEthTx()
```

fixes it for me. We need to add some comments to these fields!!!

- comment on `Status` - we need to test 1 for valid
- comment on `FabricTxStatus` - add the link to the SDK where we detail these values. @arner , you also had a PR to clear those up right?

ale-linux (changes requested):
I think I found the culprit..

Storm1289:
Thanks @ale-linux, you were right, my earlier explanation was off.

The bug is in `Observe`. It skipped invalidated commits with `if committed[i].FabricTxStatus != 0`, but that field comes from the SDK's `Transaction.Status`, which is currently never set - always 0, so the guard never fired. An MVCC-invalidated tx at nonce 5 consumes nothing on-chain, yet the gate counted it and moved the sender to 6. From there the gateway is ahead of the ledger: nonce 6 can never commit, and the client's resubmit at 5 is rejected as `nonce too low`.

Checking the EVM status instead doesn't work either, since it is 0 for both a revert and an invalidation, and a revert does consume the nonce. So I gated on `tx.Valid`, surfaced as `domain.Transaction.FabricValid`.

The unit test missed this because it set `FabricTxStatus = 1` by hand, which production never produces. It now drives `FabricValid`, plus a new test for the revert case. All tests pass locally.

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,278 @@
...
+	// may have created the sender while we were unlocked.
+	ss := g.senders[from]
+	if ss == nil {
+		g.mu.Unlock()
+		seed, err := g.state.NonceAt(ctx, from, nil)
+		if err != nil {
+			return fmt.Errorf("look up nonce: %w", err)
+		}
+		g.mu.Lock()
+		if ss = g.senders[from]; ss == nil {
```
ale-linux: I just noticed this now: while we're executing the `NonceAt` call having released the lock, `Observe` might commit a tx for this sender, then immediately evict it with `evictLRU`. Then we might get the old nonce here and park this tx waiting for the next nonce which has just committed - we would wait forever. The way to go is to create and add an `ss` placeholder instance before we release the lock and call out to `NonceAt` to make sure that `Observe` doesn't remove it under our feet - and then `evictLRU` must check whether it's a placeholder and not evict those.

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,278 @@
...
+		// its own fee-bump rules, tracked in #62; until then we log and overwrite.
+		logger.Infof("nonce gate: replacing parked tx for %s at nonce %d", from, tx.Nonce())
+	}
+	g.park(ss, tx)
+	return nil
+}
+
+// Resync updates the cache to the sender's committed nonce and drops any parked
+// tx that is now too low.
+func (g *nonceGate) Resync(ctx context.Context, tx *types.Transaction) error {
```
ale-linux: This method does three things in a row. First it figures out who sent the transaction. Then it asks the ledger what that sender's committed nonce is. Then it writes that number into its own internal table and throws away any parked transaction that is now too old to matter.

The only reason Resync exists is the test backend. The test backend can rewind the chain with snapshots, so the gate's cached idea of a sender's nonce can go stale, and the test wrapper calls Resync before every admit to catch up. Nothing in production wants this.

My suggestion is to split those three things. The first two steps, working out the sender and asking the ledger, do not need anything private. The test wrapper can do them itself, as long as it is handed a signer and something that can read a nonce, both of which are easy to pass in. Only the third step needs the gate's private table, so only the third step has to stay inside the gate.

So the gate would offer a small method, SetNext, that just says: for this sender, the next nonce is now this number, and drop whatever is parked below it. Nothing more. The test wrapper reads the ledger, gets a number, and calls SetNext with it.

Why this is better is about who decides what. Today the gate itself contains the rule "go read the ledger and believe whatever it says." That rule is wrong for production, because the ledger read can lag behind what the gate has already sent out, and believing a stale answer lets the gate move a sender's nonce backwards and submit the same nonce twice.

ale-linux: This function can be deleted (see comment to `reconciling_gate.go`)

gateway/testimpl/reconciling_gate.go
```diff
@@ -0,0 +1,33 @@
+/*
```
ale-linux: This should change: it's only used in tests, which so far succeeded just fine without it. So I propose that tests just get a no-op nonce parking implementation that passes everything through. This way, we can keep 

```go
func (g *Gateway) SendTransaction(ctx context.Context, tx *types.Transaction) error {
	if err := ValidateTx(tx, g.ChainConfig, g.Signer); err != nil {
		return err
	}
	// Reject a resubmission already in the queue or parked awaiting an earlier nonce.
	if g.TxQueue.IsPending(tx.Hash()) != nil || g.nonceGate.IsPending(tx.Hash()) != nil {
		return domain.ErrTransactionAlreadyPending
	}
	return g.nonceGate.Admit(ctx, tx)
}
```

unchanged - we just have passthrough impls of 

- `IsPending` always returning `nil`
- `Admit` immediately transferring to the queue

ale-linux (changes requested):
Thanks - one more batch of comments..

Storm1289:
Pushed `4ecfc79` with all three.

**The seed race.** `Admit` now inserts the sender entry before it drops the lock, marked as seeding, and `evictLRU` skips those so it cannot be removed while we are away. One thing I added on top: when the store answers, I only take the value if it is higher than what is already there. Otherwise an `Observe` that landed during the read gets overwritten by the older number and we are back to waiting for a nonce that has already committed. Added a test that drives the interleaving, with the old code it fails the way you described.

**Reconciling gate.** Removed, along with `Resync`. Tests now get a passthrough that enqueues everything and always returns nil from `IsPending`, so `SendTransaction` is untouched. Renamed the file to `passthrough_gate.go`, happy to move it back if you would rather keep the path. This also answers your earlier question on `test_helpers.go`, the harness no longer wraps anything.

**On the red CI.** It is not from this PR. The three directive tests fail on main as well, and on #340. I bisected locally with `TestTestNode_HardhatSetBalance`:

    87b073f (#365)  ok    1.1s
    8dbb64b (#355)  FAIL  30s

The directive submits with no error but the value never shows up on read, so the commit poll runs to its timeout. Looks like the write is getting lost in the KVS rather than anything in the directive path. @arner, might be worth a look.

gateway/core/api.go
```diff
@@ -127,7 +128,25 @@ func New(ec *EndorsementClient, batchSubmitter *BatchSubmitter, store Store, cha
 		TxQueue:         txQueue,
 		workerCount:     workerCount,
 		endorsementChan: endorsementChan,
-	}, nil
+	}
+	for _, opt := range opts {
+		opt(g)
+	}
+	// Park future-nonce transactions and release them in nonce order as earlier
+	// nonces commit. An option may have supplied one already.
```
arner: This way, we can't turn it off. Is that what we want? We can also remove this if statement and make it mandatory to use the WithNonceSequencer opt?

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,325 @@
...
+type expectedNonce struct {
+	nonce uint64
+	known bool
+}
+
+func (e expectedNonce) value() (uint64, bool) { return e.nonce, e.known }
+
+func (e *expectedNonce) raise(n uint64) {
+	if !e.known || n > e.nonce {
+		e.nonce, e.known = n, true
```
arner: nit: this is not so common so it takes more effort to read, could we just do:

```go
e.nonce = n
e.known = true
```

gateway/core/nonce_parking.go
```diff
@@ -0,0 +1,325 @@
...
+		g.evictLRU() // cannot drop ss: we still hold it
+	}
+
+	ss.lastSeen = g.now()
+
+	switch {
+	case tx.Nonce() < next:
+		return fmt.Errorf("%w: next nonce %d, tx nonce %d", ethcore.ErrNonceTooLow, next, tx.Nonce())
+	case tx.Nonce() == next:
+		g.queue.Enqueue(tx)
```
arner, line 214: Edge case: the IsPending check is already done at SendTransaction so if two identical transactions are submitted immediately after each other, they will race and it's possible that both will be enqueued. It's not a big issue (second one will fail because of MVCC), but still wasted work. It could be mitigated by checking isPending and inProgess in TxQueue (similar to TxQueueV2) as deduplication.

gateway/core/api.go
```diff
@@ -106,7 +107,7 @@ type Store interface {
 // If txQueue is nil, NewTxQueue() will be used as the default.
 // batchSubmitter handles all endorsement submissions and is owned by the Gateway.
 // endorsementChan is the channel to send endorsements to the BatchSubmitter.
-func New(ec *EndorsementClient, batchSubmitter *BatchSubmitter, store Store, chainID int64, workerCount int, txQueue TxQueueInterface, endorsementChan chan EndorsedTx) (*Gateway, error) {
+func New(ec *EndorsementClient, batchSubmitter *BatchSubmitter, store Store, chainID int64, workerCount int, txQueue TxQueueInterface, endorsementChan chan EndorsedTx, opts ...Option) (*Gateway, error) {
```
arner: The "...Option" pattern is fine in principle, but now we're using it only for the NonceSequencer. That complicates the API of the constructor. txQueue has exactly the same meaning: nillable with a default value. And workerCount as well (zero value is overridden with a default). In this case I would just replicate that pattern for NonceSequencer too. Or make the change to use functional options for all of them, but not mix two patterns in the same constructor.

arner (commented):
Looks good in principle! The queueing part is quite complex but seems well tested, I couldn't find any holes in it (except a small race for the IsPending check).

gateway/core/nonce_parking.go
ale-linux: yup - fixing!

gateway/core/api.go
ale-linux: done

gateway/core/nonce_parking.go
ale-linux: done

arner (approved):
👍

ale-linux (approved):

PR #230, opened 2026-07-13
Refactor test harness creation

arner:
Adds some reusable functions to reduce duplication, and removes unused constructor functions from the test helpers.

### DIFF
--- a/endorser/app/factory.go
+++ b/endorser/app/factory.go
@@ -18,7 +18,6 @@ import (
 	"github.com/hyperledger/fabric-x-sdk/endorsement"
 	efab "github.com/hyperledger/fabric-x-sdk/endorsement/fabric"
 	efabx "github.com/hyperledger/fabric-x-sdk/endorsement/fabricx"
-	"github.com/hyperledger/fabric-x-sdk/identity"
 	sdknet "github.com/hyperledger/fabric-x-sdk/network"
 	nfab "github.com/hyperledger/fabric-x-sdk/network/fabric"
 	nfabx "github.com/hyperledger/fabric-x-sdk/network/fabricx"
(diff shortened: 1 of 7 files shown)


endorser/app/factory.go
```diff
@@ -25,38 +25,73 @@ import (
...
+
+	var builder endorsement.Builder
+	var monotonicVersions bool
+	switch protocol {
+	case "fabric-x":
+		builder = efabx.NewEndorsementBuilder(signer)
+		monotonicVersions = true
+	default: // "fabric" or ""
+		builder = efab.NewEndorsementBuilder(signer)
+	}
```
ale-linux, line 67: afaik we managed so far to keep this package unaware of all things fabric vs. fabric-x. Are we sure we want to change that? Is there any way we can avoid spilling this in this package?

endorser/app/factory.go
```diff
@@ -25,38 +25,73 @@ import (
...
+func NewEndorser(
+	cfg config.Endorser,
+	network common.Network,
+	logger sdk.Logger,
+	testImpl bool,
+) (*core.Endorser, *sdknet.Synchronizer, storage.KVS, error) {
+	// Signer is the identity to connect to the peer for synchronizing, and for signing the endorsement.
+	signer, err := identity.SignerFromMSP(cfg.Identity.MSPDir, cfg.Identity.MspID)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to create signer: %w", err)
```
ale-linux: I think I had noted this at some point: it would probably be best to inject the signer

endorser/app/factory.go
```diff
@@ -65,29 +100,21 @@ func NewEndorser(
...
 
-	// Executing transactions and signing the endorsement.
-	monotonicVersions := network.Protocol == "fabric-x"
-	end, err := core.New(
-		execution.NewEVMEngine(network.Namespace, kvs, evmConfig, monotonicVersions),
-		builder,
-	)
-	if err != nil {
-		return nil, nil, nil, fmt.Errorf("failed to create endorser: %w", err)
-	}
```
ale-linux, line 90: I stand correct - we already did.. any chance we can actually use this refactor to avoid doing this?

ale-linux (changes requested):
Thanks Arne - A few questions/thoughts

endorser/app/factory.go
arner: Hmm good point!

The thing is, it has to live somewhere... We create the application from two places; the `integration` package for testing, and the `cmd` folder. At a certain place the application has to decide what to use. Only the `app` packages know about the switches now, none of the other ones. I don't really like duplicating it in cmd and integration. So the main role of the `app` packages is to wire things together, I think that would be an acceptable place for it. But maybe you have an alternative?

ale-linux (approved):

PR #231, opened 2026-07-13
Standalone testnode for Hardhat tests

arner:
This PR makes the testnode fully standalone, requiring no config file or certificates. It does this by embedding the json with the default testaccounts, and using hardcoded config with `fabrictest` as backend. 

The script that runs the OpenZeppelin tests now uses the fxevm testnode instead of a full gateway talking to Fablo. This makes them very fast to run: 7 seconds to start the node *and* run all the ERC-20 tests. Because of that, this PR also enables the Hardhat tests to run in the CI for every PR instead of only when labeled to request the full suite.

### DIFF
--- a/gateway/app/testnode.go
+++ b/gateway/app/testnode.go
@@ -93,5 +93,12 @@ func NewTestNode(ctx context.Context, tcfg TestNodeConfig) (*App, error) {
 		},
 	}
 
-	return buildApp(ctx, cfg, signer, logger, []eapi.Service{endorser}, nil, endorserKVS, true, tcfg.TestAccountsPath, endorserKVS)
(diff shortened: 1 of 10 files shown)


gateway/app/testnode.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	// HistorySize=128 gives evm_snapshot/evm_revert enough history to rewind through.
+	endorserDB := econfig.DB{Database: "memory", HistorySize: 128}
+	endorser, endorserKVS, _, err := eapp.NewEndorserCore(endorserDB, testNodeChannel, testNodeNamespace, protocol, signer, evmConfig, true)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create endorser: %w", err)
+	}
+
+	// db1 makes fabrictest's MVCC validation read the same DB the endorser reads.
```
kushnireyal: what's `db1`? did you meant endorserKVS?

kushnireyal (approved):
Looks good

PR #232, opened 2026-07-14
Add rate limiter to batch submitter

ale-linux:
Introduce an optional per-second rate limit on transaction submission, shared across all BatchSubmitter workers. Wire a configurable endorsement channel size through BuildGateway so both the production path and test harness respect it. Bump fabric-x-sdk to pick up the nfabx.NewSubmitter signature change and promote golang.org/x/time to a direct dependency.


arner (approved):
Thanks @ale-linux, very nice solution. I added a commit to fix the SDK upgrade. The new version requires the signer (even if it's a fake one) to have a parseable Fabric identity.

Issue #244, opened 2026-07-20
Add the gRPC endorsement client (dormant)


Storm1289:
Part of #22. Design: PR #219 (part 3). Depends on issue #243 .

### Scope
- `grpcEndorser` implementing `api.Service` over gRPC: request marshaling,  response-status/gRPC-error translation, connection lifecycle (dial, keep-alive, pooling, close).
- Unit tests against a mock server covering the full error-mapping table.

### Modularity
Lands **dormant**: the client is constructed only by tests. The gateway's production path still uses the in-process endorser and is not modified.

### Done when
- Client implements `api.Service` and round-trips all six RPCs against a mock.
- Error-mapping table fully covered by unit tests.
- Gateway default path untouched; `main` green.

Issue #243, opened 2026-07-20
Add the gRPC endorser server (dormant)


Storm1289:
## Summary

Add a dormant gRPC server implementing the `EvmEndorsement` service.

Part of #22. Design: PR #219 (parts 1, 2, and 3). Depends on Issue #242 .

## Scope

- Implement a gRPC server that adapts the six `EvmEndorsement` RPCs onto `core.Endorser` via `api.Service`.
- Add server configuration for:
  - Endpoint
  - mTLS (`mode`, `cert-path`, `key-path`, `ca-cert-paths`)
  - Keep-alive
  - Max concurrent streams
  - Rate limiting
- Add a health check endpoint.

## Modularity

The server lands **dormant**: it exists, but nothing starts or dials it, and no existing code path is modified. It must not be observable at runtime before the enablement PR.

## Done when

- The server implements all six RPCs, exercised directly by unit tests.
- Application outcomes are carried as response status (200/201/400/460/500); gRPC errors are reserved for transport failures.
- mTLS, keep-alive, and limits are configurable via the server config (end-to-end mTLS is exercised in the integration issue #245).
- Unit tests cover handler forwarding and status preservation.
- No gateway changes are introduced.
- `main` remains green.

## Dependencies

The server bootstrap (`serve`, along with `connection` and `retry`) is sourced from **fabric-x-committer** for now. Migrating it to `fabric-x-common` once hyperledger/fabric-x-committer#675 publishes `serve` there is tracked in #266.

PR #256, opened 2026-07-20
endorser: refactor api.Service to per-function methods

### LINKED ISSUE #242: Define the endorsement contract: api.Service refactor + EvmEndorsement proto
## Summary

Define the six-function endorsement contract in both Go and protobuf.

Part of #22. Design: PR #219 (parts 1 and 3).

## Scope

- Refactor `endorser/api.Service` from three methods returning `peer.ProposalResponse` to six typed methods:
  - `Execute`
  - `Call`
  - `BalanceAt`
  - `StorageAt`
  - `CodeAt`
  - `NonceAt`
- Update the in-process `core.Endorser` implementation.
- Update `EndorsementClient` and `gateway/app/wiring.go` call sites.
- Add the `.proto` definition for the six-RPC `EvmEndorsement` service.
- Add request/response messages (`Execute`, `Call`, `BalanceAt`, `StorageAt`, `CodeAt`, `NonceAt`) and check in the generated Go code.

## Modularity

The Go refactor is **behavior-preserving**—execution and behavior remain unchanged.

The protobuf definitions and generated code land **dormant**: the files exist, but nothing references them yet. No gRPC server or client is introduced in this issue.

## Done when

- `api.Service` exposes the six typed methods with no `peer.ProposalResponse`.
- `core.Endorser`, `EndorsementClient`, and wiring use the new interface.
- The protobuf definitions compile and generated Go code is checked in.
- Existing endorser and gateway test suites pass unchanged.
- `main` remains green.

Storm1289:
Closes #242.

Defines the six-function endorsement contract in both Go and protobuf.

## Go: `api.Service` refactor

- `Execute` (was `ProcessEVMTransaction`)
- `Call` now returns `([]byte, error)`; a revert or failed execution is returned as a `*common.CallError` carrying the status, message, and revert data, so the gateway continues to map it to `RevertError` / `ExecutionError` exactly as before.
- `BalanceAt`, `StorageAt`, `CodeAt`, and `NonceAt` replace the enum-dispatched `ProcessStateQuery`, allowing the now-unused `StateQuery` and `QueryType` to be removed.

This change is behavior-preserving: `make checks` is clean, and the full test suite (including integration tests) passes.

## Proto

Following the `fabric-x-common` layout:

- Added `api/endorsementpb/endorsement.proto` defining the six-RPC `EvmEndorsement` service.
- Checked in the generated `.pb.go` and `_grpc.pb.go` files alongside the proto (`paths=source_relative`).
- Added a `make proto` target mirroring the `protoc` invocation used by `fabric-x-common`.

`ExecuteRequest.proposal_hash` is left empty for Fabric-X and populated for classic Fabric, following the per-platform decision described in part 1.

This remains **dormant**: nothing outside the generated files references `endorsementpb`. No gRPC server or client is introduced here; those are covered by #243 and #244.


arner:
For protos, we can use the pattern of fabric-x-common or fabric-x-committer (similar).

arner:
Returning *peer.ProposalResponse is fine for now. I think we need the proposal anyway for classic Fabric support? Can you remind me what we discussed about a fabricx.TxPackager change?

Storm1289:
Yes, classic Fabric keeps the proposal - the `fabricx.TxPackager` change was only for the Fabric-X path. It's written up in #219, part 1 ("What Execute's Endorsement Covers"): no proposal on the wire for Fabric-X, proposal hash only for classic Fabric.

That section said "to be confirmed in code" - now confirmed: both packagers take `(end.Proposal, end.Responses...)`, so dropping the proposal for Fabric-X is exactly the small SDK change you expected. Nothing blocks this PR; it lands with the client PR, #244.

On protos - thanks, I'll follow the fabric-x-common layout and add the endorsement proto to this PR.

api/endorsementpb/endorsement.pb.go
arner, line 1: After rebasing, can you add the generated protobuf files to the 'ignore' section of codecov.yml?

api/endorsementpb/endorsement.proto
```diff
@@ -0,0 +1,114 @@
...
+    rpc CodeAt(CodeRequest) returns (CodeResponse) {};
+    rpc NonceAt(NonceRequest) returns (NonceResponse) {};
+}
+
+message ExecuteRequest {
+    // Marshaled Ethereum transaction (types.Transaction.MarshalBinary).
+    bytes ethereum_tx = 1;
+
+    // Proposal hash, required by classic Fabric because the submitted payload
+    // must carry it. Fabric-X does not need the proposal, so this is empty
```
arner, line 33: Removing the proposal requires a change in the SDK, or a workaround on our end. Which is fine; we shouldn't carry and store stuff we don't need. 

Do you have an opinion on whether the EndorsementBuilder belongs here or in the SDK? Iterations would be faster and we would have more freedom to customize it. But on the other hand, having one place to maintain is easier on the long run - especially when we want to support both Fabric and Fabric-X.

The code in the SDK is optimized for generic use cases, and giving existing Fabric features to Fabric-X (storing the request, publishing chaincode events). In the near future we may want to add more metadata (like gasUsed). Perhaps using our own protobuf for metadata would be nice (TBD). Some logic about the ethereum logs (events) already accidentally leaked into the SDK... So I'm leaning towards maintaining our own EndorsementBuilder for now, keep it generic if we can (but don't force it), and contribute back to the SDK if it makes sense.

Anyway, we don't have to decide that for this PR, but it's good to think about already.

arner (commented):
Looks good! The proto definition part is a bit hard to judge without it being connected, but at least the direction is good and if we need tweaks later that's ok too.

arner:
If you have a chance, some unit tests would be good too :)

api/endorsementpb/endorsement.proto
Storm1289: Agreed - let's keep our own EndorsementBuilder here rather than change the shared SDK. That keeps the 
EVM specific bits on our side and lets us iterate freely.

Since carrying the proposal is fine for now, this isn't a blocker - it doesn't hold up this PR or the rest of the endorsement API work. I'll open a separate issue to track it and pick it up as a follow-up once the endorsement API is in place.

api/endorsementpb/endorsement.pb.go
Storm1289: done

Storm1289:
Done - added the codecov ignore for the generated protobuf and unit tests covering the refactored api.Service methods, as you suggested. Please take a look whenever you have some time.

arner (approved):
Good stuff, thanks

Issue #245, opened 2026-07-20
Prove parity and integration across the endorsement boundary


Storm1289:
## Summary

Add parity, integration, and resilience tests for the gRPC endorsement path.

Part of #22. Design: PR #219 (part 3). Depends on Issue #244 .

## Scope

- Add parity tests that run the same `EndorsementClient` test suite against both:
  - the in-process `core.Endorser`
  - the gRPC-backed client
- Add end-to-end integration tests over real mTLS covering:
  - Transaction execution
  - Contract calls
  - State reads
- Add negative and resilience tests covering:
  - Missing client certificate
  - Untrusted client certificate
  - Plaintext connections refused
  - Server unavailable
  - Request deadlines/timeouts

## Modularity

This is a **test-only** change. The gRPC path remains **dormant** and is exercised only by the test suites. No production code paths are modified.

## Done when

- Parity is proven between the in-process and gRPC-backed implementations.
- Integration, security, and resilience test suites pass.
- The gateway's default execution path remains unchanged.
- `main` remains green.

PR #262, opened 2026-07-22
endorser: add dormant gRPC server for the EvmEndorsement service

### LINKED ISSUE #243: Add the gRPC endorser server (dormant)

Storm1289:
Closes #243 

Adds a **dormant** gRPC server that implements the `EvmEndorsement` service by forwarding the six RPCs to the in-process endorser through the `api.Service` interface. Nothing starts or dials it yet; it is exercised only by unit tests.

## What's here (`endorser/server`)

- `server.go`
  - Implements `Server` (embedding `UnimplementedEvmEndorsementServer` and holding an `api.Service`)
  - Adds `RegisterService` (endorsement service + gRPC health)
  - Adds `Serve`
  - Defines `type Config = serve.Config`
- `handlers.go`
  - Implements the six RPC handlers
  - The five state reads and `Call` forward directly
  - `Call` carries application outcomes (revert, execution failure, rejected transaction) in-band through the response status/message/data, reserving gRPC errors for transport failures
- `server_test.go`
  - Adds unit tests using an in-memory `bufconn` connection with a stub `api.Service`
  - Covers handler forwarding, response status mapping, the application-vs-transport error split, and `Execute` forwarding

## Execute

`Execute` reconstructs the `Invocation` server-side from the transaction (arguments `[{proposalType}, ethTxBytes]`, the endorser's namespace, and the caller's proposal hash when present) before forwarding it to the existing builder.

For Fabric-X, the builder only requires the arguments, chaincode name, and transaction ID. Owning our own `EndorsementBuilder` and removing the proposal entirely is tracked separately in #265.

## Bootstrap (`serve`)

Server bootstrap (listener, mTLS, keep-alives, limits, and health) currently comes from `serve` in **fabric-x-committer**, allowing the server to be implemented and reviewed.

Once `fabric-x-committer#675` publishes `serve` into `fabric-x-common`, the import will be switched (tracked in #266).

Because of this, `go mod tidy` promotes `fabric-x-committer` to a direct dependency, bringing in a few transitive dependencies (for example `pgx` and `backoff`) that will disappear once the migration to `fabric-x-common` is complete.

## Tests

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt` (no changes)
- [x] Unit tests pass

### DIFF
--- a/endorser/server/handlers.go
+++ b/endorser/server/handlers.go
@@ -1,0 +1,136 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+*/
+
+package server
+
+import (
+	"context"
+	"errors"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/hyperledger/fabric-protos-go-apiv2/peer"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/hyperledger/fabric-x-evm/api/endorsementpb"
+	"github.com/hyperledger/fabric-x-evm/common"
+	"github.com/hyperledger/fabric-x-sdk/endorsement"
+)
+
+// Execute endorses an Ethereum transaction. Application outcomes ride in the
+// response status; only transport faults surface as a gRPC error.
+func (s *Server) Execute(ctx context.Context, req *endorsementpb.ExecuteRequest) (*endorsementpb.ExecuteResponse, error) {
+	tx := new(types.Transaction)
+	if err := tx.UnmarshalBinary(req.GetEthereumTx()); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "unmarshal tx: %v", err)
+	}
+	// Provisional: today the client builds the invocation (creator identity +
+	// proposal); the wire shape that carries it, and the fabricx.TxPackager
+	// change behind the Fabric-X "no proposal on the wire" decision, are
+	// finalized with the client PR (#244). Until then the server forwards with
+	// an empty invocation; req.proposal_hash is reserved for classic Fabric.
+	resp, err := s.svc.Execute(ctx, endorsement.Invocation{}, tx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "execute: %v", err)
+	}
+	return executeResponse(resp), nil
+}
+
+// Call runs a read-only eth_call. A revert or failed execution is an
+// application outcome carried in the response; only transport faults are a
+// gRPC error.
+func (s *Server) Call(ctx context.Context, req *endorsementpb.CallRequest) (*endorsementpb.CallResponse, error) {
+	out, err := s.svc.Call(ctx, callMsg(req), blockNumber(req.BlockNumber))
+	if err != nil {
+		var ce *common.CallError
+		if errors.As(err, &ce) {
+			return &endorsementpb.CallResponse{ReturnData: ce.Data, Status: ce.Status, Message: ce.Message}, nil
+		}
+		return nil, status.Errorf(codes.Internal, "call: %v", err)
+	}
+	return &endorsementpb.CallResponse{ReturnData: out, Status: common.StatusOK}, nil
+}
+
+func (s *Server) BalanceAt(ctx context.Context, req *endorsementpb.BalanceRequest) (*endorsementpb.BalanceResponse, error) {
+	bal, err := s.svc.BalanceAt(ctx, ethcommon.BytesToAddress(req.GetAccount()), blockNumber(req.BlockNumber))
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "balance: %v", err)
+	}
+	return &endorsementpb.BalanceResponse{Balance: bal.Bytes()}, nil
+}
+
+func (s *Server) StorageAt(ctx context.Context, req *endorsementpb.StorageRequest) (*endorsementpb.StorageResponse, error) {
+	val, err := s.svc.StorageAt(ctx, ethcommon.BytesToAddress(req.GetAccount()), ethcommon.BytesToHash(req.GetKey()), blockNumber(req.BlockNumber))
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "storage: %v", err)
+	}
+	return &endorsementpb.StorageResponse{Value: val}, nil
+}
+
+func (s *Server) CodeAt(ctx context.Context, req *endorsementpb.CodeRequest) (*endorsementpb.CodeResponse, error) {
+	code, err := s.svc.CodeAt(ctx, ethcommon.BytesToAddress(req.GetAccount()), blockNumber(req.BlockNumber))
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "code: %v", err)
+	}
+	return &endorsementpb.CodeResponse{Code: code}, nil
+}
+
+func (s *Server) NonceAt(ctx context.Context, req *endorsementpb.NonceRequest) (*endorsementpb.NonceResponse, error) {
+	nonce, err := s.svc.NonceAt(ctx, ethcommon.BytesToAddress(req.GetAccount()), blockNumber(req.BlockNumber))
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "nonce: %v", err)
+	}
+	return &endorsementpb.NonceResponse{Nonce: nonce}, nil
+}
+
+// blockNumber turns the optional block selector into the *big.Int the engine
+// expects; a nil selector means latest.
+func blockNumber(bn *uint64) *big.Int {
+	if bn == nil {
+		return nil
+	}
+	return new(big.Int).SetUint64(*bn)
+}
+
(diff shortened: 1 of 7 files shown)


endorser/server/handlers.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Call runs a read-only eth_call. A revert or failed execution is an
+// application outcome carried in the response; only transport faults are a
+// gRPC error.
+func (s *Server) Call(ctx context.Context, req *endorsementpb.CallRequest) (*endorsementpb.CallResponse, error) {
+	out, err := s.svc.Call(ctx, callMsg(req), blockNumber(req.BlockNumber))
+	if err != nil {
+		var ce *common.CallError
+		if errors.As(err, &ce) {
```
arner: Bit nicer:

```go
if ce, ok := errors.AsType[*common.CallError](err); ok {
    return &endorsementpb.CallResponse{ReturnData: ce.Data, Status: ce.Status, Message: ce.Message}, nil
}
```

endorser/server/handlers.go
```diff
@@ -0,0 +1,136 @@
...
+	tx := new(types.Transaction)
+	if err := tx.UnmarshalBinary(req.GetEthereumTx()); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "unmarshal tx: %v", err)
+	}
+	// Provisional: today the client builds the invocation (creator identity +
+	// proposal); the wire shape that carries it, and the fabricx.TxPackager
+	// change behind the Fabric-X "no proposal on the wire" decision, are
+	// finalized with the client PR (#244). Until then the server forwards with
+	// an empty invocation; req.proposal_hash is reserved for classic Fabric.
+	resp, err := s.svc.Execute(ctx, endorsement.Invocation{}, tx)
```
arner: This locks us in the approach of owning the builder, which is not planned until https://github.com/hyperledger/fabric-x-evm/issues/265 and there is still a chance that we decide against it eventually.

It might be better to continue with the existing builder first, which requires the Invocation for both Fabric and Fabric-X. I'm aware that passing the Invocation might require a change (temporarily?) in the protos. But then at least we can build against the current way, and optimize later.

Or do you have a different point of view?

endorser/server/server.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/hyperledger/fabric-x-evm/api/endorsementpb"
+	"github.com/hyperledger/fabric-x-evm/endorser/api"
+)
+
+// Config is the gRPC server configuration (endpoint, mTLS, keep-alive,
+// max-concurrent-streams, rate-limit). It is the serve package's own config.
+//
+// serve is sourced from fabric-x-committer for now; once it is published into
+// fabric-x-common (committer#675) this alias moves there, a mechanical change.
```
arner: Before we forget, can you please make a github issue as part of the endorser API epic to track this work? It's important we realign it as soon as it's ready, to keep our dependency graph clean.

I'd consider this one of the few cases where a TODO or FIXME comment is justified 😁

endorser/server/server.go
```diff
@@ -0,0 +1,53 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+*/
+
+// Package server exposes the endorser over gRPC by adapting the EvmEndorsement
+// service onto the in-process endorser through the api.Service seam. It is
+// dormant: nothing starts or dials it until the enable step (#246).
```
arner: There are several comments in this file about the progress of the work related to other issues, but let's just keep the comments clean. We know from the context of the PR that it's not used yet, but this package doesn't know or care about that.

endorser/server/server.go
```diff
@@ -0,0 +1,53 @@
...
+// RegisterService registers the endorsement and health services on the gRPC
+// server. It implements serve.Registerer so the server can be bootstrapped by
+// the serve package.
+func (s *Server) RegisterService(servers serve.Servers) {
+	endorsementpb.RegisterEvmEndorsementServer(servers.GRPC, s)
+	healthpb.RegisterHealthServer(servers.GRPC, health.NewServer())
+}
+
+// Serve bootstraps the gRPC server (listen, mTLS, keep-alive, limits, health)
+// from cfg and serves until ctx is done. Dormant until wired by the enable step.
```
arner: We don't have to mention that it's dormant.

arner (changes requested):

endorser/server/handlers.go
```diff
@@ -0,0 +1,136 @@
...
+	if v := req.GetValue(); len(v) > 0 {
+		msg.Value = new(big.Int).SetBytes(v)
+	}
+	return msg
+}
+
+// executeResponse maps the endorser's ProposalResponse onto the wire response.
+// The read-write set and event live inside the payload today; splitting them
+// into their own fields is finalized with the client PR (#244).
+func executeResponse(pr *peer.ProposalResponse) *endorsementpb.ExecuteResponse {
```
arner, line 127: In the Service interface, you said:

```
	// It returns a *peer.ProposalResponse because the gateway packages it into
	// an sdk.Endorsement, and both SDK packagers require one. Dropping it
	// depends on the fabricx.TxPackager change tracked for the client PR.
```

Maybe I'm still confused but it looks here as if we don't return the *peer.ProposalResponse to the client. If it's just a data transfer object within our own service, we should not depend on a Fabric proto but just have our own struct. However, my confusion is that the builder still needs one... Are we in limbo between two approaches, or am I just missing something here? 🤔

endorser/server/handlers.go
Storm1289: Agreed - I don't want to pre-commit to owning the builder here either. I'll drop the empty-invocation shortcut and use the existing builder, which needs the Invocation. I'll carry the Invocation over the wire as a temporary proto
addition so we build against the current path and optimize later.

endorser/server/handlers.go
Storm1289: The gRPC ExecuteResponse is our own proto - the handler maps the builder's *peer.ProposalResponse onto it, so the Fabric proto stays internal and never crosses the wire. You're right that our own result struct would be cleaner; api.Service still returns *peer.ProposalResponse today, and replacing that is the ater optimization (the own-builder work), not this PR. So: existing builder now, own type later - not two approaches at once.

endorser/server/handlers.go
Storm1289: Done

endorser/server/server.go
Storm1289: Done

endorser/server/server.go
Storm1289: removed

endorser/server/server.go
Storm1289: done

Storm1289:
Hi @arner, please review again.

endorser/server/handlers.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil, status.Errorf(codes.Internal, "execute: %v", err)
+	}
+	return executeResponse(resp), nil
+}
+
+// invocation rebuilds the endorsement invocation from the transaction using the
+// endorser's identity and namespace, keeping the caller's proposal hash when set.
+func (s *Server) invocation(ethTx, proposalHash []byte) (endorsement.Invocation, error) {
+	args := [][]byte{{byte(common.ProposalTypeEVMTx)}, ethTx}
+	inv, err := endorsement.NewInvocation(s.signer, s.channel, s.namespace, s.nsVersion, args)
```
arner: The signer of an invocation is the sender, not the endorser. In other words, the gateway component. So I don't think attaching the signer to the server struct to construct invocations will work... In your other comment you said "I'll carry the Invocation over the wire as a temporary proto", let's just stick with that for now :)

Storm1289: You're right, the invocation's signer is the sender, so the endorser can't build it. Reverted that: the Invocation now travels over the wire as a temporary proto message (tx_id, args, chaincode name/version) and the endorser just maps it onto the builder. Also filled the coverage gaps Codecov flagged - the package is at 98%, with only the serve bootstrap left (exercised in #245).

arner (approved):
Good progress, thanks @Storm1289 !

PR #267, opened 2026-07-22
fix: blocknumber sentinel values

### LINKED ISSUE #192: eth_getLogs with toBlock: "latest" silently returns empty array
### Summary
When `eth_getLogs` is called with `toBlock: "latest"`, the gateway 
returns an empty array even when matching events exist on-chain. 
No error is returned — the call succeeds with HTTP 200 and `[]`.

This also affects Blockscout's token transfer indexer, which uses 
`toBlock: "latest"` internally — causing the Token Transfers tab 
to show 0 transfers for all contracts despite events existing on-chain.

### Environment
- fabric-x-evm: latest main branch (June 2026)
- go-ethereum: v1.16.7
- Chain ID: 4011
- ethers.js: v6
- Network: native Docker Compose stack (compose.fabric-x.yml + compose.yml)

### Steps to Reproduce
1. Deploy any ERC-20 contract
2. Call mint() — creates a Transfer event on-chain
3. Call eth_getLogs with toBlock: "latest"

```javascript
const logs = await provider.send("eth_getLogs", [{
  address: contractAddress,
  topics: [ethers.id("Transfer(address,address,uint256)")],
  fromBlock: "0x1",
  toBlock: "latest",   // ← bug triggered here
}]);
console.log(logs); // prints [] despite Transfer event existing
```

### Expected Result
`eth_getLogs` returns the matching Transfer events.

### Actual Result
Gateway returns `[]` with HTTP 200. No error. Events are silently missed.

### Additional Observation
The same bug is triggered by ethers.js `contract.queryFilter()` 
which passes `toBlock: "latest"` internally — making the bug 
invisible unless you debug the raw RPC call.

Blockscout's token transfer indexer appears to hit the same bug — 
Token Transfers tab shows 0 for all contracts despite logs existing.

### Workaround
Use explicit hex block number instead of "latest":

```javascript
const head = await provider.getBlockNumber();
const headHex = "0x" + head.toString(16);
const logs = await provider.send("eth_getLogs", [{
  address: contractAddress,
  topics: [ethers.id("Transfer(address,address,uint256)")],
  fromBlock: "0x1",
  toBlock: headHex,   // explicit hex — works correctly
}]);
```

### Possible Root Cause
In `gateway/api/eth.go` lines 446-448, `ToBlock` is only set 
when explicitly provided as a number. When `"latest"` is passed, 
`blockNumberOrHashToBlockNumber` appears to return 0 which 
`blockNumberToUint64` maps to "latest" — resulting in no upper 
bound being set on the filter query.

### Note
`fromBlock: "0x0"` also has unexpected behaviour — it maps to 
"latest" rather than genesis block, as documented in 
`docs/COMPATIBILITY.md`. Combined with `toBlock: "latest"`, 
a query with both set to their "default" values returns no 
results at all.

### Additional observation: this bug also affects the Blockscout
integration included in compose.yml.

Blockscout's token transfer indexer appears to use
toBlock: "latest" internally when querying eth_getLogs.
As a result, the Token Transfers tab shows 0 transfers
for all contracts, even when Transfer events exist on-chain
and are visible in the Logs tab.

Tested with:

ERC-20 contract (3 transactions, 2 logs, 0 token transfers shown)
ERC-1404 contract (5 transactions, 4 logs, 0 token transfers shown)
The Logs tab correctly shows events exist — confirming the
issue is in the indexer query, not the events themselves.

arner:
Closes #192, along with 2 related sentinel-resolution bugs.

**gateway/api/eth.go** - filterCriteriaToLogFilter (the original #192 fix) and rpcBlockNumberToBigInt (new: earliest/safe/finalized were leaking through as raw negative numbers instead of resolving correctly. New helper resolveLogFilterBlockNumber.

**endorser/execution/executor.go** - newExecutor and newSnapshotAt both called .Uint64() directly on a possibly-negative *big.Int, corrupting eth_call/eth_estimateGas/eth_getBalance/eth_getCode/eth_getStorageAt/eth_getTransactionCount into reading state from the wrong block when a negative sentinel leaked through.

**gateway/core/api.go** - GetBlockTxCountByNumber and GetTransactionByBlockNumberAndIndex were missing the math.MaxUint64→latest resolution that GetBlockByNumber already had.

### DIFF
--- a/gateway/api/eth.go
+++ b/gateway/api/eth.go
@@ -414,7 +414,11 @@ func (api *EthAPI) FeeHistory(ctx context.Context, blockCount hexutil.Uint, last
 // eth_getLogs
 func (api *EthAPI) GetLogs(ctx context.Context, crit filters.FilterCriteria) ([]*types.Log, error) {
 	logger.Debugf("EthAPI.GetLogs() called with criteria=%+v", crit)
-	query := filterCriteriaToLogFilter(crit)
+	query, err := api.filterCriteriaToLogFilter(ctx, crit)
+	if err != nil {
+		logger.Debugf("EthAPI.GetLogs() returning error: %v", err)
+		return nil, err
+	}
 
 	logs, err := api.b.GetLogs(ctx, query)
 	if err != nil {
@@ -432,15 +436,19 @@ func (api *EthAPI) GetLogs(ctx context.Context, crit filters.FilterCriteria) ([]
 	return result, nil
 }
 
-func filterCriteriaToLogFilter(crit filters.FilterCriteria) domain.LogFilter {
+func (api *EthAPI) filterCriteriaToLogFilter(ctx context.Context, crit filters.FilterCriteria) (domain.LogFilter, error) {
 	filter := domain.LogFilter{}
 
 	if crit.BlockHash != nil {
 		hash := crit.BlockHash.Bytes()
 		filter.BlockHash = &hash
 	} else {
-		filter.FromBlock = resolveLogFilterBlockNumber(crit.FromBlock)
-		filter.ToBlock = resolveLogFilterBlockNumber(crit.ToBlock)
+		from, err := api.resolveLogFilterFromBlock(ctx, crit.FromBlock)
+		if err != nil {
+			return domain.LogFilter{}, err
+		}
+		filter.FromBlock = from
+		filter.ToBlock = resolveLogFilterToBlock(crit.ToBlock)
 	}
 
 	if len(crit.Addresses) > 0 {
@@ -462,7 +470,7 @@ func filterCriteriaToLogFilter(crit filters.FilterCriteria) domain.LogFilter {
 		}
 	}
 
-	return filter
+	return filter, nil
 }
 
 func domainLogToTypesLog(l domain.Log) *types.Log {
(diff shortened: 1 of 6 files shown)


gateway/api/eth.go
```diff
@@ -604,3 +604,21 @@ func blockNumberToUint64(num rpc.BlockNumber) uint64 {
...
+	if n == nil {
+		return nil
+	}
+	if n.Cmp(big.NewInt(int64(rpc.EarliestBlockNumber))) == 0 {
+		zero := uint64(0)
+		return &zero
+	}
+	if n.Sign() < 0 {
+		return nil
+	}
```
kushnireyal, line 649: I think that for the "from" block, returning `nil` here will mean that it will retrieve everything starting from the genesis block, while if the users request, for example, "latest", they actually expect a "from" block of the current head block

arner: Good catch! It's inconsistent. Without this change, a -2 would be converted to block 2, which is wrong. But indeed with this change the backend defaults to just getting all blocks, which is worse...

I pushed a commit that should fix this, by getting the actual latest.

gateway/api/eth.go
```diff
@@ -584,9 +578,15 @@ func (api *EthAPI) blockNumberOrHashToBlockNumber(ctx context.Context, numOrHash
 	return new(big.Int).SetUint64(*num), nil
 }
 
-// rpcBlockNumberToBigInt converts rpc.BlockNumber to *big.Int
+// rpcBlockNumberToBigInt converts rpc.BlockNumber to *big.Int for state queries.
+// "earliest" resolves to block 0; every other negative sentinel (latest, pending, safe,
```
kushnireyal, line 582: 0 is the kvs marker for "latest", not "earliest", is it a problem?

Image

kushnireyal: endorser/storage/lightkvs.go line 165

kushnireyal (approved):
See comments

arner:
Oops, yeah 😵 

Indeed this is a problem for KVS; when asking for earliest you get the latest.... It's actually a separate (but related!) issue. I'll open a new PR for it.

PR #269, opened 2026-07-22
OpenZeppelin/Hardhat compatibility tests ci gate

arner:
This change does a couple of things:

- Adds a CI gate for the OpenZeppelin compatibility tests. Any divergence from the baseline file causes the CI to fail:
  - ❌ Regressions must be fixed, or exceptions manually added to the baseline json
  - 🎉 New tests succeeding? Great, remove the succeeding tests from the baseline json
  - ✅ Only if the outcome matches the baseline json, the CI succeeds.
- Creates or updates a comment on a PR with the ✅/❌/🎉 result of the CI gate and a summary with short instructions.
- Makes the `baseline` tool easier to use with default file names.
- Adds JSON output to the `baseline` tool to make it machine readable.

### DIFF
--- a/.github/workflows/baseline-report-comment.yml
+++ b/.github/workflows/baseline-report-comment.yml
@@ -37,17 +37,29 @@ jobs:
         id: prnum
         run: |
           if [ -f oz-baseline-artifact/pr_number.txt ]; then
-            echo "number=$(cat oz-baseline-artifact/pr_number.txt)" >> "$GITHUB_OUTPUT"
+            n=$(cat oz-baseline-artifact/pr_number.txt)
+            [[ "$n" =~ ^[0-9]+$ ]] || { echo "Non-numeric PR number; skipping."; exit 0; }
+            echo "number=$n" >> "$GITHUB_OUTPUT"
           else
             echo "No PR number found — not a pull_request run, or no artifact was produced. Skipping comment."
           fi
 
       - name: Find-or-update PR comment
+        uses: actions/github-script@v7
+        env:
+          PR_NUMBER: ${{ steps.prnum.outputs.number }}
+        with:
+          script: |
+            const prNumber = Number(process.env.PR_NUMBER);
+
+      - name: Find-or-update PR comment
         if: steps.prnum.outputs.number != ''
         uses: actions/github-script@v7
+        env:
+          PR_NUMBER: ${{ steps.prnum.outputs.number }}
         with:
           # Builds a compact PR comment from cmd/baseline check --json's structured
-          # Summary: an always-visible ✅/❌/🎉 headline, regressions/stale surfaced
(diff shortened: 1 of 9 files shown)


arner:
Note that the comment will not be posted on this PR, but only after the flow is merged to main. That's just a limitation of how github actions works (only flows on main can trigger like that).

.github/workflows/baseline-report-comment.yml
```diff
@@ -0,0 +1,126 @@
...
+                out += `**Expected failures by cause (${total})**\n\n`;
+                for (const c of summary.causes) out += `- ${c.cause}: ${c.count}\n`;
+                out += '\n';
+              }
+              out += `</details>\n\n`;
+            }
+            out += footer;
+
+            const { owner, repo } = context.repo;
+            const prNumber = Number('${{ steps.prnum.outputs.number }}');
```
kushnireyal: There is prompt injection vulnerability here (found with claude):

> 
> `steps.prnum.outputs.number` comes from `cat oz-baseline-artifact/pr_number.txt` — a file inside an artifact produced by the **untrusted** Tests run. For `pull_request` events, the Tests workflow definition itself comes from the PR head, so a fork PR can rewrite the "Record PR number" step to put arbitrary text into `pr_number.txt`.
> 
> Because that value is **template-interpolated into the `github-script` source** (not passed as runtime data), a payload like `1'));/*evil JS*/('` breaks out of `Number('…')` and executes arbitrary JavaScript in the privileged context — which holds `pull-requests: write` and a live `GITHUB_TOKEN`. This is exactly the `workflow_run` + `github-script` interpolation pattern GitHub's hardening docs warn against.
> 
> Note the contrast: `summary.json` is read at runtime via `fs.readFileSync` + `JSON.parse` (data, safe). Only `pr_number.txt` flows into the *script text*, which is what makes it dangerous.
> 
> **Fix** — pass via env and validate as digits:
> ```yaml
> - name: Check for PR number
>   id: prnum
>   run: |
>     if [ -f oz-baseline-artifact/pr_number.txt ]; then
>       n=$(cat oz-baseline-artifact/pr_number.txt)
>       [[ "$n" =~ ^[0-9]+$ ]] || { echo "Non-numeric PR number; skipping."; exit 0; }
>       echo "number=$n" >> "$GITHUB_OUTPUT"
>     fi
> ```
> ```yaml
> - name: Find-or-update PR comment
>   uses: actions/github-script@v7
>   env:
>     PR_NUMBER: ${{ steps.prnum.outputs.number }}
>   with:
>     script: |
>       const prNumber = Number(process.env.PR_NUMBER);
> ```

kushnireyal (approved):
Looks good!

.github/workflows/baseline-report-comment.yml
arner: Good catch! Implemented the fix.

Issue #248, opened 2026-07-20
[EPIC] Hardhat compatibility


arner:
When running as `testnode`, we should support a large portion of hardhat APIs and features. This allows developers to test their smart contracts against real Fabric-X EVM behavior. At the same time it helps us test our compatibility with common smart contract features, like we do with the openzeppelin smart contracts test suite.

Examples:

- eth_signTypedData_v4
- hardhat_impersonateAccount
- hardhat_mine.

We should not blindly aim for 100% compatibility, but rather investigate per feature if it's worth the implementation and maintenance effort.

PR #271, opened 2026-07-23
fix: register hardhat_mine stub in the test RPC

### LINKED ISSUE #251: Register a hardhat_mine stub in the test RPC
Part of #248.

Only `hardhat_setCode` and the `evm_*` helpers are registered (`gateway/testimpl/server.go`); add `hardhat_mine` next to `evm_mine` (`gateway/testimpl/hardhat_helpers.go`). Stub only removes the hard "method not found" — most callers (mostly `RateLimiter`/`ERC3009`/`ERC20TransferAuthorization` setup code) just need the call to not error, not real time control (real block-timestamp/number control is a separate, architecturally harder design question — not in scope here).

125 tests in a local run of the OpenZeppelin Hardhat suite — the single largest test-RPC-only bucket. Test-RPC only, no production analog (Hardhat's own dev-network namespace).

Siddhartha-singh01:
Closes #251
Part of #248

Registers `hardhat_mine` on the test RPC next to the existing Hardhat helpers. This is a stub: the call succeeds and optional `blocks` / `interval` args are accepted for signature compatibility, but chain state is not advanced. Fabric still produces blocks through consensus, so real block or timestamp control stays out of scope (same boundary the issue calls out).

Most OZ callers that hit this method only need it not to return "method not found" (RateLimiter / ERC3009 / ERC20TransferAuthorization setup paths). That matches how `evm_mine` is already stubbed.

## Code

`gateway/testimpl/hardhat_helpers.go` adds `HardhatAPI.Mine`. The hardhat namespace is already registered in the test server, so no wiring change was required beyond the method itself.

`gateway/testimpl/hardhat_helpers_test.go` covers the in process RPC path with zero, one, and two optional hex params (Hardhat docs examples).

## Baseline

After a full compatible set run (`make hardhat-tests-full` / `scripts/run_hardhat_test.sh --full`), `testdata/oz_known_failures.json` was reconciled so the scoreboard matches the stub being present:

* 125 prior failures tagged `hardhat_mine` are gone as a cause
* 52 of those tests now pass and were dropped from the baseline
* remaining failures that used to die on the missing method fail later for other reasons (reverts, funds, timestamps, etc.) and are recorded accordingly

### Full suite summary

```
4755 passed, 896 failed, 1 skipped (5652 total, 84.1% passing)
```

By suite: finance 0/2, metatx 0/12, governance 77/185, access 354/578, crosschain 10/15, proxy 194/252, token 1637/1852, utils 2483/2755.

`cmd/baseline check` against that run is clean (no regressions, no stale entries). Result JSON under `testdata/oz-hardhat-results/` stays gitignored.

## Validation

* `go test ./gateway/testimpl/ ./cmd/baseline/`
* `go test ./... -short`
* `go vet` / `staticcheck` on the tree
* CI style Hardhat path: `ERC20.test.js` (126 passing)
* full OZ compatible set + baseline check as above

Happy to tweak cause tags on the new baseline entries if a different labeling pass is preferred.

### DIFF
--- a/gateway/testimpl/hardhat_helpers.go
+++ b/gateway/testimpl/hardhat_helpers.go
@@ -47,6 +47,15 @@ func (api *HardhatAPI) SetCode(ctx context.Context, address common.Address, code
 	return true, nil
 }
 
+// Mine accepts hardhat_mine for RPC compatibility with Hardhat tests.
+// It does not advance chain state: fabric-evm creates blocks via Fabric
+// consensus, not mining. Optional blocks/interval params match Hardhat's
+// signature and are ignored.
+func (api *HardhatAPI) Mine(ctx context.Context, blocks *hexutil.Uint64, interval *hexutil.Uint64) error {
+	hardhatLogger.Debugf("HardhatAPI.Mine() called blocks=%v interval=%v (stub; no state change)", blocks, interval)
+	return nil
+}
+
 // EvmAPI provides EVM-specific RPC methods for testing, particularly snapshot/revert.
 // Uses LightKVS history mechanism to capture and restore ledger state, and Store
 // for database snapshot/revert.
--- a/gateway/testimpl/hardhat_helpers_test.go
+++ b/gateway/testimpl/hardhat_helpers_test.go
@@ -1,0 +1,54 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+
+WARNING: This package contains test-only/unsafe RPC implementations.
+DO NOT use in production environments.
+*/
+
+package testimpl
+
+import (
+	"context"
+	"testing"
+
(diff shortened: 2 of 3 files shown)


arner (approved):
Great stuff, thanks @Siddhartha-singh01 !

The new "execution reverted" are expected. If we wanted to get rid of those, we'd have to implement the actual mining function. Which will be a good next step (I'll write an issue for it), but for now this is good progress!

arner:
@Siddhartha-singh01 , if you feel like continuing, we have a couple of similar issues in this epic: https://github.com/hyperledger/fabric-x-evm/issues/248. Just reply to one of the sub issues and I'll assign you. Only if you want, of course!

Siddhartha-singh01:
Thanks @arner! Glad the baseline reclassifications look as expected. 
 
 Happy to pick up another issue under #248  Just Assign it to me .

PR #273, opened 2026-07-23
endorser: add dormant gRPC endorsement client

### LINKED ISSUE #244: Add the gRPC endorsement client (dormant)

Storm1289:
Closes #244 

Adds a **dormant** gRPC endorsement client: a new `api.Service` implementation that talks to a remote endorser over gRPC. This is the client-side counterpart to #243. Nothing is wired into the gateway yet; it is exercised only by unit tests.

## What's here (`endorser/client`)

- `client.go`
  - Implements `Client`, wrapping an `endorsementpb.EvmEndorsementClient`
  - Asserts the interface at compile time: `var _ api.Service = (*Client)(nil)`
  - Adds `New(conn)` and `Close()`
  - Implements the six RPCs as the inverse of the server mapping:
    - `Execute` marshals the transaction and invocation into `ExecuteRequest` and maps `ExecuteResponse` back to `*peer.ProposalResponse`
    - `Call` converts non-OK application responses into `*common.CallError` (including revert data), while transport failures remain gRPC errors
    - `BalanceAt`, `StorageAt`, `CodeAt`, and `NonceAt` return their typed values

## Tests

- `client_test.go`
  - Unit tests using a mock `EvmEndorsementServer` over `bufconn`
  - Covers all six RPC round-trips
  - Covers request marshaling
  - Covers application vs. transport error mapping (success, revert, transport failure)
  - Covers `Close()` on a nil connection
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt` (no changes)

### DIFF
--- a/endorser/client/client.go
+++ b/endorser/client/client.go
@@ -1,0 +1,156 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+*/
+
+// Package client is an api.Service backed by a remote endorser over gRPC.
+package client
+
+import (
+	"context"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/hyperledger/fabric-protos-go-apiv2/peer"
+	"google.golang.org/grpc"
+
+	"github.com/hyperledger/fabric-x-evm/api/endorsementpb"
+	"github.com/hyperledger/fabric-x-evm/common"
+	"github.com/hyperledger/fabric-x-evm/endorser/api"
+	"github.com/hyperledger/fabric-x-sdk/endorsement"
+)
+
+var _ api.Service = (*Client)(nil)
+
+// Client adapts a remote EvmEndorsement gRPC endpoint to the api.Service seam.
+type Client struct {
+	rpc  endorsementpb.EvmEndorsementClient
+	conn *grpc.ClientConn
+}
+
+// New returns a Client that talks to the endorser over conn.
+func New(conn *grpc.ClientConn) *Client {
+	return &Client{rpc: endorsementpb.NewEvmEndorsementClient(conn), conn: conn}
+}
+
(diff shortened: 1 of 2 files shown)


endorser/client/client.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/hyperledger/fabric-protos-go-apiv2/peer"
+	"google.golang.org/grpc"
+
+	"github.com/hyperledger/fabric-x-evm/api/endorsementpb"
+	"github.com/hyperledger/fabric-x-evm/common"
+	"github.com/hyperledger/fabric-x-evm/endorser/api"
+	"github.com/hyperledger/fabric-x-sdk/endorsement"
+)
+
+var _ api.Service = (*Client)(nil)
```
arner: We can remove this, we'll notice in other ways if we don't comply to the interface.

endorser/client/client.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	if msg.Value != nil {
+		req.Value = msg.Value.Bytes()
+	}
+	return req
+}
+
+// blockNumberProto turns the block selector into the optional wire field; a nil
+// selector means latest.
+func blockNumberProto(bn *big.Int) *uint64 {
```
arner, line 147: In https://github.com/hyperledger/fabric-x-evm/pull/267/changes, we check the sign. We might need that here too to prevent subtle bugs (see top file in that PR).

endorser/client/client.go
Storm1289: done

arner (approved):
Thanks @Storm1289 ! Looks good

PR #280, opened 2026-07-25
fix: register hardhat_impersonateAccount stub in the test RPC

### LINKED ISSUE #252: Register a hardhat_impersonateAccount stub in the test RPC
Part of #248.

Same registration gap as `hardhat_mine` (see the sibling issue) — not registered in `gateway/testimpl/server.go`/`hardhat_helpers.go`. Unblocks tests acting as an arbitrary address without its key.

29 tests in a local run of the OpenZeppelin Hardhat suite. Test-RPC only.

Siddhartha-singh01:
Closes #252 Part of #248

Same registration gap as hardhat_mine (see the sibling issue) hardhat_impersonateAccount is not registered in gateway/testimpl/server.go / hardhat_helpers.go. The OpenZeppelin Hardhat suite calls this method early in a bunch of test setup paths (AccessManaged, GovernorTimelockAccess, RelayedCall, TransparentUpgradeableProxy, etc.), and those tests immediately die with "method not found" before they can reach the actual assertion.

This PR adds both hardhat_impersonateAccount and hardhat_stopImpersonatingAccount as no-op stubs on the HardhatAPI struct. The calls succeed and return nil, which is enough for the test harness to move past the method-not-found barrier. No actual impersonation logic is added  Fabric doesn't expose an unlocked-account concept the way Hardhat does, so real impersonation is out of scope here (same boundary the hardhat_mine stub follows). @arner flagged hardhat_stopImpersonatingAccount in the issue thread as well, so both are covered.

Code
gateway/testimpl/hardhat_helpers.go  adds two new methods on HardhatAPI:

ImpersonateAccount(ctx, address) accepts the address parameter for signature compatibility, logs at DEBUG level, returns nil. The go-ethereum RPC layer maps hardhat_impersonateAccount to this automatically via the hardhat namespace already registered in server.go through NewHardhatAPI(), so no wiring change was needed.
StopImpersonatingAccount(ctx, address)  mirror companion, same approach.
gateway/testimpl/hardhat_helpers_test.go  refactored and extended:

Extracted the server setup (rpc.NewServer + RegisterName + DialInProc) from the existing TestHardhatAPI_Mine_RPCRegistration into a shared dialHardhat(t) helper with t.Cleanup for client teardown. This keeps the test file DRY now that we have multiple test functions hitting the same namespace.
TestHardhatAPI_ImpersonateAccount_RPCRegistration calls both hardhat_impersonateAccount and hardhat_stopImpersonatingAccount through the in-process RPC client with a valid hex address, asserts the result is nil and no error is returned.
TestHardhatAPI_ImpersonateAccount_MissingAddress  confirms the RPC layer itself rejects calls that omit the required address parameter (the go-ethereum server returns an error for missing args before our handler is even invoked).
Baseline
After a full compatible-set run (make hardhat-tests-full / scripts/run_hardhat_test.sh --full), testdata/oz_known_failures.json was reconciled so the scoreboard matches the stub being present:

All 29 prior failures tagged hardhat_impersonateAccount are reclassified  those tests now get past the missing-method error but still fail downstream on hardhat_setBalance (the next unimplemented Hardhat method in the call chain). The impersonation stub succeeds, then the test tries to set the impersonated account's balance, and that call is what fails now.
Zero remaining hardhat_impersonateAccount cause tags in the baseline.
No tests moved from failing to passing this time (unlike the hardhat_mine stub where 52 tests started passing). That's expected  every test that calls hardhat_impersonateAccount also calls hardhat_setBalance right after, so they just hit the next wall.
cmd/baseline check against that run is clean (no regressions, no stale entries).

Verification
go test ./gateway/testimpl/ new and existing Hardhat tests pass
go test ./... -short
go vet / staticcheck on the tree
full OZ compatible set + cmd/baseline check  no regressions, no stale entries

Happy to adjust cause tags or split the commits differently if preferred.

### DIFF
--- a/gateway/testimpl/hardhat_helpers.go
+++ b/gateway/testimpl/hardhat_helpers.go
@@ -56,6 +56,20 @@ func (api *HardhatAPI) Mine(ctx context.Context, blocks *hexutil.Uint64, interva
 	return nil
 }
 
+// ImpersonateAccount accepts hardhat_impersonateAccount for RPC compatibility.
+// It does not enable signing or sending as the given address.
+func (api *HardhatAPI) ImpersonateAccount(ctx context.Context, address common.Address) error {
+	hardhatLogger.Debugf("HardhatAPI.ImpersonateAccount() called address=%s (stub; no state change)", address.Hex())
+	return nil
+}
+
+// StopImpersonatingAccount accepts hardhat_stopImpersonatingAccount for RPC
+// compatibility. It is a no-op companion to ImpersonateAccount.
+func (api *HardhatAPI) StopImpersonatingAccount(ctx context.Context, address common.Address) error {
+	hardhatLogger.Debugf("HardhatAPI.StopImpersonatingAccount() called address=%s (stub; no state change)", address.Hex())
+	return nil
+}
+
 // EvmAPI provides EVM-specific RPC methods for testing, particularly snapshot/revert.
 // Uses LightKVS history mechanism to capture and restore ledger state, and Store
 // for database snapshot/revert.
--- a/gateway/testimpl/hardhat_helpers_test.go
+++ b/gateway/testimpl/hardhat_helpers_test.go
@@ -18,14 +18,19 @@ import (
 	"github.com/hyperledger/fabric-x-evm/gateway/domain"
 )
 
-func TestHardhatAPI_Mine_RPCRegistration(t *testing.T) {
+func dialHardhat(t *testing.T) *rpc.Client {
+	t.Helper()
 	srv := rpc.NewServer()
 	if err := srv.RegisterName("hardhat", NewHardhatAPI()); err != nil {
 		t.Fatalf("RegisterName hardhat: %v", err)
 	}
-
 	client := rpc.DialInProc(srv)
-	defer client.Close()
+	t.Cleanup(client.Close)
+	return client
+}
+
+func TestHardhatAPI_Mine_RPCRegistration(t *testing.T) {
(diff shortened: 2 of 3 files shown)


Siddhartha-singh01:
@arner If you'd like any changes restructuring commits, tweaking cause tags, adjusting the test coverage  just let me know and I'll get on it. Really enjoying contributing to this project; it aligns well with my interests and I'd be happy to keep picking up issues under #248.

arner:
Happy to hear that @Siddhartha-singh01 , and we appreciate your contributions! Just put a comment on any issue (with status 'Ready') that you'd like to pick up and I'll assign you to it.

I noticed that the baseline tool causes noisy diffs due to character escaping. A PR https://github.com/hyperledger/fabric-x-evm/pull/282 should fix that. After that gets merged, can you please rebase this PR on main and re-generate the json output? Thanks!

Siddhartha-singh01:
Thanks @arner  for the review! Will rebase and re-generate the JSON output once #282 lands.

arner:
Hi @Siddhartha-singh01 , #282 got merged (along with another bugfix for the compatibilty tests). If you rebase and re-generate the json I'll review! Thanks

Siddhartha-singh01:
Hi @arner, rebased on main (which now includes #282 and #285), resolved the merge conflict in `hardhat_helpers_test.go`, and re-generated the baseline JSON with the new canonical encoding. Ready for review 

thanks!

arner (approved):
looking good, thanks @Siddhartha-singh01

PR #282, opened 2026-07-27
ensure canonical encoding for baseline json

arner:
Due to html escaping, diffs for the baseline json were noisy and contained unnecessary ugly escape sequences like \u2014. This change ensures canonical, readable encoding.


kushnireyal (approved):
Looks good

PR #283, opened 2026-07-27
endorser: add pebble-backed persistent KVS as drop-in for LightKVS (#218)

### LINKED ISSUE #218: KVS Persistence
The current implementation for the KVS uses an in-memory database. This is fast but requires persistence. We have 2 options:

1. use pebbleDB from geth to achieve persistence/caching natively
2. implement ad-hoc persistence

This Issue tracks the work required to test pebble and make a decision about how to persist KVS data

kushnireyal:
Closes #218.

Adds `PebbleKVS`, a persistent MVCC store behind the existing `storage.KVS` interface, as a drop-in for the in-memory `LightKVS` (lost on restart). New implementation plus a factory `switch` case; no executor, StateDB or synchronizer changes.

Each block commits as one atomic pebble batch with a persisted block number, so a reopened store resumes from the last committed block. Where it differs from `LightKVS` it follows the sqlite `VersionedDB` the committer validates against: `MAX(version)+1` per key, deletes kept as tombstones. Enable with `database.database: pebble` and `connection-string` as the data directory.

### DIFF
--- a/endorser/storage/kvs_parity_test.go
+++ b/endorser/storage/kvs_parity_test.go
@@ -9,6 +9,7 @@ package storage
 import (
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/hyperledger/fabric-x-common/api/committerpb"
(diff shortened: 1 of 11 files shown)


endorser/storage/pebblekvs.go
```diff
@@ -0,0 +1,517 @@
...
+	return binary.BigEndian.Uint64(raw), nil
+}
+
+// Update atomically applies a batch of writes for a single block. All writes
+// carry the same block number (they come from the same block); the block's data
+// keys and the meta checkpoint are committed in one pebble batch, so a crash
+// either leaves the whole block applied or none of it.
+func (p *PebbleKVS) Update(updates []KeyValueVersion) error {
+	if len(updates) == 0 {
+		return nil
```
arner, line 183: If the block is empty (or it's a config block, or has no ethereum transactions), this skips the update of the metaBlockKey. It's probably fine, but let's think about it.

The main failure mode is a crash after which we start one block earlier and just replay an empty block. In the case of one Synchronizer that feeds both the endorsement KVS and also the blocks database, there is a chance that they briefly disagree about the block height (if the block db logs empty blocks). At the moment, our block database is deliberately not idempotent (but only so that we can catch issues during development). 

Without looking in detail in how we get our block height after a crash and how the other db processes empty blocks, it seems like a potential problem - and even if it's not the case in our current code, it might surface after a change (for instance if we take the KVS as truth for blockheight, we sync one early, and the blocks db crashes because of primary key constraints).

How can we best mitigate this? I think this is a good moment to make block DB idempotent. It doesn't care if an older block is replayed, so that removes at least that part of the failure model. It looks like the world state KVS must never process a transaction twice; it will mess up the versioning permanently. Do you think we should also add a guard to it so that it skips blocks with a lower block number than recorded?

I was also looking into what the Fabric-X blocks give us. For Reads and Read+Writes we actually get the version back! However blind rights don't have that, they only have key and value (see https://github.com/hyperledger/fabric-x-common/blob/main/api/applicationpb/block_tx.proto). Perhaps it would be good to validate our assumed nextVersion with the value from the block for the keys that include a read. And then... log an error? Crash? if it doesn't match... This would require a small change in the SDK, would that be worthwhile or can we rely on our own version accounting?

endorser/storage/pebblekvs.go
```diff
@@ -0,0 +1,517 @@
...
+	}
+
+	p.writeMu.Lock()
+	defer p.writeMu.Unlock()
+
+	blockNum := updates[0].BlockNum
```
arner: This is probably safe due to how we call it, but might be worth a small guard to enforce the blocknumber to be the same for all?

endorser/storage/pebblekvs.go
```diff
@@ -0,0 +1,517 @@
...
+	if len(notifs) == 0 {
+		return nil
+	}
+
+	// Group by block so each block commits as one atomic batch, then apply
+	// blocks in ascending order.
+	byBlock := make(map[uint64][]KeyValueVersion)
+	var order []uint64
+	for _, notif := range notifs {
+		valid := notif.Status == committerpb.Status_COMMITTED
```
arner, line 367: This, and the specific monotonic versioning scheme, make the KVS _only_ work for Fabric-X.

We could add support for Fabric later if we want (no urgency here, as the SQLite store provides persistence at a throughput that's sufficient to keep up with Fabric). However this is not clear to users, and plugging in the Pebble KVS will lead to non-obvious issues. How about adding a validation in the config that checks if the database type matches the blockchain protocol and fails early if it doesn't match?

endorser/storage/kvs_parity_test.go
```diff
@@ -0,0 +1,529 @@
...
+// interface boundary (the white-box tests in lightkvs_test.go exercise
+// LightKVS-internal structure and are not part of this contract).
+type kvsBackend struct {
+	name string
+	// make builds a fresh, empty KVS. historySize is honored by LightKVS and
+	// ignored by PebbleKVS (which keeps full history).
+	make func(t *testing.T, historySize int) KVS
+}
+
+func kvsBackends() []kvsBackend {
```
arner, line 32: This parity check is very valuable!

Do you think it'd be possible to add the VersionedDB with a SQLite backend too? Unless.... We can also get rid of the SQLite backend. Pebble is, as far as I can tell, in every way a better backend for our usecase.

Ok how about this:

1. We add VersionedDB with SQLite backend to this test.
2. We add another dimension to the tests: Fabric vs Fabric-X semantics.
3. We add Fabric support to PebbleKVS.
4. We remove VersionedDB / VersionedDBWrapper.

arner: It also means that some of these tests could be refactored a bit to be more generic. The kvsBackend struct could have explicit markers about what they support so we can skip a test rather than failing (e.g. no persistence in lightkvs).... What do you think?

It could function as documentation and make the behaviour explicit, and be extensible with (resilience) tests quite easily. It would give us stronger guarantees that backends behave identically and prevent regressions. We could do this as a separate followup PR.

arner (commented):
This is really nice @kushnireyal ! Clear performance improvement over SQLite, well tested and the behaviour looks solid. 

I have a few small suggestions for the code and a couple of discussion points, mainly for followups. Let me know what you think of those and if you agree we can make issues for them. @ale-linux your input would be valued here too.

arner (approved):
Looks good! Approved, provided we add a quick check in the config to make sure people don't run Pebble with classic Fabric.

endorser/storage/pebblekvs.go
```diff
@@ -264,21 +327,33 @@ func (p *PebbleKVS) Get(namespace, key string, lastBlock uint64) (*blocks.WriteR
 
 // Handle implements blocks.BlockHandler: it extracts all valid transaction
 // writes from a block and applies them atomically. Called by the synchronizer.
+//
+// A block that yields no writes — empty, config-only, or carrying nothing but
+// invalid transactions — still advances the persisted checkpoint. LightKVS can
```
arner, line 332: IMO we can leave out everything after "LightKVS", it's a bit verbose.

arner (commented):
Can you please amend [c16f9fa](https://github.com/hyperledger/fabric-x-evm/commit/c16f9fac732ef9a2c082a0d8e568e42b5395f7be) to add your signature?

Issue #246, opened 2026-07-20
Enable the gRPC endorsement path (config + wiring)


Storm1289:
Part of #22. Design: PR #219 (parts 3 and 4). 

The **only PR that turns the feature on**, landing after issues #245  is green.

### Scope
- Gateway builds `grpcEndorser` values from `[]common.ClientConfig` endorser endpoints: the co-located same-org endorser over localhost, other orgs remotely.
- Config-driven selection; absent the config, behavior is exactly as today.
- Operator/deployment docs and example wiring (maintainer-gated).

### Modularity
Everything landed dormant in issues #242 - #245  becomes reachable here, and only via config - so the switch is revertible on its own without touching earlier PRs.

### Done when
- The gRPC path is reachable via config; default behavior unchanged without it.
- End-to-end suites green.

PR #284, opened 2026-07-27
integration: prove endorsement gRPC parity, mTLS, and resilience

### LINKED ISSUE #245: Prove parity and integration across the endorsement boundary

Storm1289:
## Summary

Closes #245 

Adds parity, integration, and resilience tests for the gRPC endorsement path introduced in #243 and #244. This PR is **test-only** it does not modify production code. The gRPC endorsement path remains dormant and is exercised only through these test suites.

## What's here (`integration/`)

- `tls_helper_test.go`
  - Adds helpers to generate a self-signed CA and server/client certificates for real mTLS testing.
  - Adds a helper to generate a client certificate signed by an unrelated (untrusted) CA for negative test scenarios.

- `endorsement_grpc_test.go`
  - Adds seven integration tests covering the complete gRPC endorsement path:
    - `TestGRPCEndorsement_Integration`
      - Verifies `Execute`, `Call`, and all four state readers over a real gRPC connection secured with mTLS against a real `core.Endorser` created using the existing `integration.NewEndorser` harness.
    - `TestGRPCEndorsement_Parity`
      - Executes identical scenarios through both the in-process endorser and the gRPC endorser, verifying identical status codes and response messages.
    - `TestGRPCEndorsement_MissingClientCert`
    - `TestGRPCEndorsement_UntrustedClientCert`
    - `TestGRPCEndorsement_PlaintextRefused`
      - Verifies all invalid authentication scenarios are rejected with `codes.Unavailable`.
    - `TestGRPCEndorsement_ServerUnavailable`
    - `TestGRPCEndorsement_DeadlineExceeded`
      - Verifies transport failures surface as retryable gRPC errors rather than hanging indefinitely.

## Notes

- Uses `connection.NewClientTLSCredentials` and `connection.NewClientGRPCTransportCredentials` instead of constructing a custom `tls.Config`, ensuring the client follows the same production TLS path that #246 will eventually use.
- Uses `serve.PreAllocateListener` so the client always connects to the real ephemeral port before the gRPC server starts serving.
- Uses a transaction signed with an incorrect chain ID for deterministic rejection in the parity and integration tests. Since the standalone integration harness does not execute a commit pipeline, nonce reuse cannot reliably produce deterministic failures.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt` (no changes)
- [x] `make checks`
- [x] `make unit-tests`
- [x] `go test -race` (executed 5× with no flakiness)


integration/endorsement_grpc_test.go
```diff
@@ -0,0 +1,430 @@
...
+				KeyPath:     certs.keyPath,
+				CACertPaths: []string{certs.caCertPath},
+			},
+		},
+	}
+	serve.PreAllocateListener(t, &cfg.GRPC)
+	addr := cfg.GRPC.Endpoint.Address()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
```
arner: Would this be handled if we use t.Context() instead of the context with cancel?

Storm1289:
Hi @Arne,
Whenever you have some time, could you please review this pr.

integration/endorsement_grpc_test.go
```diff
@@ -0,0 +1,430 @@
...
+		t.Errorf("NonceAt parity: direct=(%v,%v) wire=(%v,%v)", directNonce, directErr, wireNonce, wireErr)
+	}
+
+	// Same two-step scenario (valid tx accepted, then a wrong-chain-ID tx
+	// rejected) driven independently through both paths; their Invocations
+	// differ (fresh TxID each), but the status/message outcome at each step
+	// must match.
+	directKey, directTx1 := newSignedTx(t)
+	directResp1, directErr := inProcess.Execute(ctx, newInvocation(t, directTx1), directTx1)
+	wireKey, wireTx1 := newSignedTx(t)
```
arner: If we submit _the same_ transaction / invocation, we can also ensure that the endorsed payload is the same byte for byte. That's necessary for successful transactions.

integration/endorsement_grpc_test.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+	creds, err := connection.NewClientGRPCTransportCredentials(tlsCreds)
+	if err != nil {
+		t.Fatalf("NewClientGRPCTransportCredentials: %v", err)
+	}
+	return creds
+}
+
+// dialWithCreds opens a gRPC client connection to addr using creds. Dialing is
+// lazy (grpc.NewClient does not connect eagerly), so failures surface on the
```
arner, line 108: Does this mean that in production, they _do_ surface when starting the service? I think it should, because we want to know right away if something is wrong.

arner (approved):
Nice, thanks! Can you take a look at the comments?

integration/endorsement_grpc_test.go
Storm1289: That's scheduled for the enable step (#246), where the gateway dials real endorsers, we'll make sure connectivity is verified at startup there, not deferred to the first transaction.

integration/endorsement_grpc_test.go
Storm1289: Good point, done

integration/endorsement_grpc_test.go
Storm1289: Done

Storm1289:
Hi @arner, please review the changes

PR #285, opened 2026-07-27
fix: fence hardhat transactions to keep reverted state clean

arner:
A data race is breaking the openzeppelin tests on main. It's caused by old asynchronous transactions committing after a revert.

Gating the sendTransaction API with a mutex and dropping late transactions should resolve the issue.


kushnireyal (approved):
Looks good

PR #286, opened 2026-07-28
Run OZ tests by file or filter, and choose port

arner:
This PR improves the developer experience of working with the OpenZeppelin hardhat tests.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -164,7 +164,7 @@ jobs:
             ${{ runner.os }}-oz-artifacts-
 
       - name: Hardhat OZ compatibility suite (full)
-        run: make hardhat-tests-full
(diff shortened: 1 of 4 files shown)


Storm1289:
Hi @arner, Nice work
One gap in the --port isolation though: RESULTS_DIR (testdata/oz-hardhat-results) is still a fixed path, and run_full_suite does rm -rf on it. So two concurrent full runs on different ports can clobber each other's results, and the baseline check could end up reading a mix of both runs. Narrowed --file/--grep runs are fine since they never touch the results dir. Might be worth either deriving RESULTS_DIR from --port and passing it to baseline check --results, or documenting --port as a narrowed-run aid.
what do you think?

Storm1289:
LGTM

ale-linux (approved):

PR #288, opened 2026-07-29
drain the tx queue on evm_snapshot/evm_revert

arner:
Sorry for the churn; apparently there was still an intermittent race because we were rejecting too strictly. This change uses shorter locks and reads state from the existing queue, so we can be smarter about what we reject and what we let in.

- fence the submit path up to the enqueue rather than the commit: a rewind now either sees a transaction in the queue or turns it away, and one that never commits no longer stalls every later rewind
- drain the gateway's own queue via a new TxQueueInterface.InFlight, instead of a count kept alongside it in the test RPC
- still reject only while a rewind is actually running; queueing there would apply a transaction built against the old state to the new one
- bound the drain and commit waits, and error instead of panicking when a transaction is dropped before it commits

Fixes some of the intermittent "before each" hook failures in the OpenZeppelin tests.

### DIFF
--- a/gateway/app/app.go
+++ b/gateway/app/app.go
@@ -165,7 +165,7 @@ func buildApp(ctx context.Context, cfg config.Config, gwSigner sdk.Signer, logge
 		// Wrap the chain's store with SnapshotStore for snapshot/revert functionality
 		snapshotStore := storage.NewSnapshotStore(chain.Store)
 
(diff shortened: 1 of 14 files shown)


ale-linux (approved):
LGTM - I can't claim to fully understand what's going on but the bulk of the code changes target the testimpl so I think tests passing are enough of a check. If you are unhappy with this, maybe we could have @kushnireyal also take a look? Your call @arner

arner:
Thanks @ale-linux , for me this suffices for the testimpl.

PR #290, opened 2026-07-29
test: run execution-specs transaction_tests through the gateway

### LINKED ISSUE #91: Integrate `TransactionTests` from ethereum/tests
This story tracks the integration of https://github.com/ethereum/tests/tree/v17.0/TransactionTests into fabric-x-evm. Tests should exercise the same aspects as https://github.com/ethereum/go-ethereum/blob/master/tests/transaction_test.go

arunabha003:
### Addressed issue

Closes #91 — integrate `TransactionTests` from the ethereum conformance corpus.

### Summary

- Add `TestTransactionTests`, which walks the execution-specs `transaction_tests`
  fixtures and drives each raw transaction through the gateway's real admission
  path (`UnmarshalBinary` + `core.ValidateTx`), limited to the same fork allowlist
  as the `state_tests` suite (Osaka/BPO1/BPO2 + Prague/Cancun).
- These fixtures are all negative cases, so the runner asserts the **reject/accept
  verdict** rather than the exact EEST exception string — our gateway may reject at
  a coarser level (e.g. an unsupported transaction type) than the fine-grained
  reason the fixture names. This mirrors go-ethereum's own `transaction_test.go`,
  which also checks validity rather than the exception text.
- Extend the `eth-tests` make target to run `TestTransactionTests` alongside
  `TestEthereumTests`; document the new suite in `COMPATIBILITY.md`.
- Add an (empty) `testdata/transaction_tests.skip` mirroring `eth_tests.skip`.

No production/runtime code changes — this adds a test suite plus its build/docs wiring.

### Validation done

- `make checks` passes.
- `TestTransactionTests` passes (113 assertions across the allowlisted forks); a
  throwaway probe confirmed a *valid* transaction is accepted, so the all-reject
  fixtures aren't passing vacuously.
- `make eth-tests` runs both suites green (exit 0).

## Screenshot
image

### DIFF
--- a/integration/transaction_test.go
+++ b/integration/transaction_test.go
@@ -1,0 +1,131 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+*/
+
+package integration
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"math/big"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/tests"
+	"github.com/hyperledger/fabric-x-evm/gateway/core"
+	"google.golang.org/grpc/grpclog"
+)
+
+// transactionTestFile is the execution-specs transaction_test fixture shape:
+//
+//	{ "<name>": { "txbytes": "0x..", "result": { "<Fork>": { "exception": "..", "intrinsicGas": ".." } } } }
+//
+// A non-empty "exception" for a fork means the raw transaction must be rejected
+// under that fork's rules; its absence means the transaction is valid.
+type transactionTestFile map[string]struct {
+	TxBytes hexutil.Bytes `json:"txbytes"`
+	Result  map[string]struct {
+		Exception string `json:"exception"`
+	} `json:"result"`
+}
+
+// txNonceReader is a stub nonce source for the admission check: transaction_tests
+// carry no pre-state, and every fixture that reaches the nonce check is rejected
+// earlier on tx-type grounds, so a constant zero nonce is sufficient.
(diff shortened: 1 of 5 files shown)


testdata/transaction_tests.skip
```diff
@@ -0,0 +1,3 @@
+# execution-specs transaction_tests that are skipped in ALL modes.
+# One test-file path per line; blank lines and lines starting with '#' are ignored.
+# Empty for now — every allowlisted-fork transaction_tests fixture passes.
```
arner: 🥳

integration/transaction_test.go
```diff
@@ -0,0 +1,131 @@
...
+// transaction_test.go, which also checks validity rather than the exception text.
+func TestTransactionTests(t *testing.T) {
+	grpclog.SetLoggerV2(grpclog.NewLoggerV2(io.Discard, os.Stderr, os.Stderr)) // disable grpc logging
+
+	testsDir := filepath.Join("..", "testdata", "execution-specs-tests", "fixtures", "transaction_tests")
+	if _, err := os.Stat(testsDir); os.IsNotExist(err) {
+		t.Skipf("execution-specs fixtures not found at %s; run `make fetch-execution-specs-tests`", testsDir)
+	}
+
+	skip, err := loadSkip(filepath.Join("..", "testdata", "transaction_tests.skip"))
```
arner: I understand that this is here for consistency, but I propose to remove the entire skip mechanics for the TransactionTests. First reason is that we don't need it for TransactionTests (yay!) and likely will never need it. Secondly, I have a PR open (https://github.com/hyperledger/fabric-x-evm/pull/289) where, if approved, we'll move to the same known_failures.json as we use for openzeppelin tests. That will allow for finer grained control over our compatibility scores and causes, and more consistency between the test suites.

arner (commented):
Nice that they all succeed! One suggestion, see above.

integration/transaction_test.go
arunabha003: got it ! addressed in the latest commit .

arner (approved):
Thanks @arunabha003 , good stuff as always

PR #291, opened 2026-07-30
fix: register evm_setAutomine stub in the test RPC

### LINKED ISSUE #253: Register an evm_setAutomine stub in the test RPC
Part of #248.

There's already a real pending-transaction mechanism (`gateway/core/txqueue.go`'s `TxQueue`/`TxQueueV2`, tracking submitted-but-uncommitted txs via `pendingQueue`/`inProgressMap`), so real automine control could in principle be built on it — e.g. gating dispatch on automine state, or driving block production off a schedule of empty transactions when automine is on.

Block cadence today is driven by the Fabric orderer's `BatchTimeout` (`testdata/configtx.yaml:176`), not something the RPC layer owns, and most OZ callers just need the call to not error rather than real manual-mining semantics (`evm_mine` is already a no-op for the same reason, `gateway/testimpl/hardhat_helpers.go:177-183`). Given the small count (6 tests in a local run), real support isn't worth the complexity right now relative to what it'd buy — a stub unblocks the actual failures. Worth revisiting if a real use case needs actual pending/manual-mine semantics later.

Test-RPC only.

Siddhartha-singh01:
Closes #253
Part of #248

Same kind of gap as the hardhat_mine and hardhat_impersonateAccount stubs: the OpenZeppelin suite calls evm_setAutomine, and our test RPC did not register it, so those paths failed with method not found.

This adds SetAutomine on EvmAPI as a no-op. It accepts the bool, returns success, and does not change how blocks are produced. Fabric still owns block timing through the orderer, same reason evm_mine is already a stub. The issue notes only a handful of tests hit this, so a real automine implementation is not worth the complexity right now.

## Code

gateway/testimpl/hardhat_helpers.go adds EvmAPI.SetAutomine. The evm namespace is already registered in the test server, so no server.go change was needed.

gateway/testimpl/hardhat_helpers_test.go covers the in process RPC path for true and false, plus a missing argument case. Shared dialEvm / dialHardhat helpers keep the setup tidy next to the existing Hardhat tests.

## Baseline

Full compatible set run (scripts/run_hardhat_test.sh --full), then reconciled testdata/oz_known_failures.json:

* all 6 prior failures tagged evm_setAutomine no longer fail for method not found
* those same 6 still fail later (pending txs, nonce, reverts, or assertion noise), which is expected when automine is only a stub
* zero remaining evm_setAutomine cause tags

cmd/baseline check against that run is clean (no regressions, no stale entries).

## Verification

* go test ./gateway/testimpl/
* go test ./... -short
* go vet / staticcheck on the touched package
* CI style Hardhat path: ERC20.test.js (126 passing)
* full OZ suite + baseline check

Happy to tweak cause tags if anything looks off. Thanks!

### DIFF
--- a/gateway/testimpl/hardhat_helpers.go
+++ b/gateway/testimpl/hardhat_helpers.go
@@ -223,6 +223,13 @@ func (api *EvmAPI) Mine(ctx context.Context) (string, error) {
 	return "0x0", nil
 }
 
+// SetAutomine accepts evm_setAutomine for RPC compatibility with Hardhat tests.
+// It does not change mining mode: block production is driven by Fabric consensus.
+func (api *EvmAPI) SetAutomine(ctx context.Context, enabled bool) error {
+	hardhatLogger.Debugf("EvmAPI.SetAutomine() called enabled=%v (stub; no state change)", enabled)
+	return nil
+}
+
 // IncreaseTime increases the timestamp of the next block (evm_increaseTime).
 // This is a stub that returns the time increase amount.
 func (api *EvmAPI) IncreaseTime(ctx context.Context, seconds hexutil.Uint64) (hexutil.Uint64, error) {
--- a/gateway/testimpl/hardhat_helpers_test.go
+++ b/gateway/testimpl/hardhat_helpers_test.go
@@ -8,6 +8,7 @@ package testimpl
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


Siddhartha-singh01:
Hi @arner, opened this for #253. Same pattern as the mine and impersonate stubs. Happy to tweak anything if you want changes. Thanks!

arner (approved):
Thanks! It's funny, we have automine by default (we instantly create a block for every transaction). It's probably the SetAutomine(false) that is causing tests to fail. We might add an issue for that later, but it's not the priority now.

arner:
Ah sorry, that last merge caused a conflict, could you please rebase and resolve?

Siddhartha-singh01:
Hi @arner, rebased on latest main (after #280 landed), fixed the merge conflict in hardhat_helpers_test.go, and pushed. CI is running again. Ready to merge when you’re happy with it. Thanks!

PR #294, opened 2026-07-30
refactor: drop HandleTx, unify notification path onto existing Handle…

ale-linux:
… interface

The HandleTx method introduced a second block-handling interface alongside the existing Handle. This change removes it: AllTxBatchDispatcher now assembles a blocks.Block from the streamer events and calls the standard Handle, so LightKVS, VersionedDBWrapper, TxQueue, and TxQueueV2 lose their HandleTx implementations and the TxNotification.NsRWS/Events fields that only served them.

### DIFF
--- a/common/notification.go
+++ b/common/notification.go
@@ -7,10 +7,10 @@ SPDX-License-Identifier: LGPL-3.0-or-later
 package common
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 
-	"github.com/ethereum/go-ethereum/common"
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
 	"github.com/hyperledger/fabric-protos-go-apiv2/peer"
 	"github.com/hyperledger/fabric-x-common/api/applicationpb"
@@ -19,18 +19,6 @@ import (
 	"github.com/hyperledger/fabric-x-sdk/notification"
 	"google.golang.org/protobuf/proto"
 )
-
-// TxNotification contains data about a committed Ethereum transaction, reconstructed
-// from a blocks.Transaction for use by components that need EVM-specific fields
-// (e.g. EthTxHash, EthTxBytes) beyond what blocks.Transaction provides.
-type TxNotification struct {
-	BlockNum   uint64
-	TxNum      uint64
-	FabricTxID string
-	Status     committerpb.Status
-	EthTxBytes []byte
-	EthTxHash  common.Hash
-}
 
 var notifLogger = flogging.MustGetLogger("evm.notification")
 
@@ -71,6 +59,11 @@ func (d *AllTxBatchDispatcher) HandleBatch(ctx context.Context, batch notificati
 		var input peer.ChaincodeInput
 		if err := proto.Unmarshal(event.Metadata[0], &input); err != nil || len(input.Args) < 2 {
 			notifLogger.Debugf("Skipping tx %s: cannot extract eth tx from metadata", event.TxID)
+			continue
+		}
+
+		if !bytes.Equal(input.Args[0], []byte{byte(ProposalTypeEVMTx)}) {
+			notifLogger.Debugf("Skipping tx %s: not an EVM transaction", event.TxID)
 			continue
 		}
 
--- a/common/notification.go
+++ b/common/notification.go
@@ -11,7 +11,6 @@ import (
 	"fmt"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
 	"github.com/hyperledger/fabric-protos-go-apiv2/peer"
 	"github.com/hyperledger/fabric-x-common/api/applicationpb"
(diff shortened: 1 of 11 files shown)


common/notification.go
```diff
@@ -21,46 +20,39 @@ import (
...
 }
 
 var notifLogger = flogging.MustGetLogger("evm.notification")
 
-// TxHandler defines the interface for handlers that process transaction notifications in batches.
-// Both gateway and endorser components implement this to receive committed-transaction
-// notifications over the AllTxStreamer path (an alternative to the block-synchronizer path).
+// TxHandler defines the interface for handlers that process committed blocks delivered
+// via the AllTxStreamer path (an alternative to the block-synchronizer path).
 type TxHandler interface {
```
arner: This is now a block handler instead of a TxHandler. Also defined in `blocks.BlockHandler`

integration/perf/replay_json_dataset_test.go
```diff
@@ -71,15 +72,34 @@ func (t *TxCompletionTracker) Stop() {
 	t.stopped = true
 }
 
-// HandleTx implements common.TxHandler. It receives notifications about completed transactions
-// and forwards them to the completion channel.
-func (t *TxCompletionTracker) HandleTx(ctx context.Context, notifs []fxcommon.TxNotification) error {
+// Handle implements common.TxHandler. It receives committed transactions and forwards
```
arner: Handle implements blocks.BlockHandler

integration/perf/replay_json_dataset_test.go
```diff
@@ -71,15 +72,34 @@ func (t *TxCompletionTracker) Stop() {
...
-	for _, notif := range notifs {
+	for _, tx := range b.Transactions {
+		var ethTx types.Transaction
+		if len(tx.InputArgs) < 2 {
+			continue
+		}
+		if err := ethTx.UnmarshalBinary(tx.InputArgs[1]); err != nil {
+			continue
+		}
+		status := committerpb.Status_COMMITTED
```
arner: This is Fabric-X specific (classic fabric has different codes). A Transaction has a Status field with the original Fabric(x) status, can't we just use that?

integration/state_primer.go
```diff
@@ -333,14 +333,14 @@ func (sp *StatePrimer) commitAndWait(end sdk.Endorsement, tx *types.Transaction,
 		}
 	}
 
-	// Only create the in-proc RPC client when we wait; the no-wait path leaked it.
```
arner, line 336: It looks like we don't need the `ec` outside of the `wait` condition, so we don't have to leak it, right? Before your change I think it was closing correctly?

arner:
Nice, looks like a good refactor already! Left a couple of comments. 

Somehow this seems to have fixed 137 issues in the Hardhat compatibility suite (no idea how 😅)!

Can you please run `go run ./cmd/baseline update --suite oz-hardhat` to update the baseline json?

common/notification.go
ale-linux: done

integration/perf/replay_json_dataset_test.go
ale-linux: done

integration/perf/replay_json_dataset_test.go
ale-linux: done

integration/state_primer.go
ale-linux: yes, that was some refactoring mess-up

common/notification.go
kushnireyal: since this structure already exists in https://github.com/hyperledger/fabric-x-sdk/blob/c4f59746192e7773a584bb85a2cf103cf195eafe/blocks/processor.go#L32 as arne mentioned, can it be re-used from there?

common/notification.go
```diff
@@ -21,116 +20,93 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-// TxNotification contains all data needed to process a transaction notification.
-// EthTxBytes and EthTxHash come from the cache; NsRWS and Events come from the AllTxStreamer event.
+// TxNotification contains data about a committed Ethereum transaction, reconstructed
+// from a blocks.Transaction for use by components that need EVM-specific fields
+// (e.g. EthTxHash, EthTxBytes) beyond what blocks.Transaction provides.
 type TxNotification struct {
```
kushnireyal: nothing uses it now so it can be removed as far as i understand

common/notification.go
```diff
@@ -21,116 +20,93 @@ import (
...
 // AllTxBatchDispatcher implements notification.AllTxHandler. It bridges AllTxStreamer
-// (which delivers every committed transaction) to the internal TxHandler chain.
+// (which delivers every committed transaction) to the internal BlockHandler chain.
 //
-// For each committed transaction event it:
-//  1. Extracts the Ethereum transaction bytes from the event metadata.
-//  2. Converts the event's Namespaces to NsRWS + Events.
-//  3. Builds a TxNotification and dispatches it to all registered TxHandlers.
+// For each committed block it:
+//  1. Filters out non-EVM transactions (those without Ethereum tx bytes in InputArgs).
```
kushnireyal, line 35: i think you removed the ethTxBytes check. I see that the filter is only `if err := proto.Unmarshal(event.Metadata[0], &input); err != nil || len(input.Args) < 2 {` now. Does it filter out non-evm transactions?

kushnireyal (changes requested):
see comments

common/notification.go
ale-linux: I think it's idiomatic Go to define interfaces where needed and not import those defined by other users

common/notification.go
ale-linux: good catch - done!

common/notification.go
ale-linux: Good point. For now I don't think full unmarshalling is necessary, but I've added the same check we have [here](https://github.com/hyperledger/fabric-x-evm/blob/main/gateway/core/chain.go#L147-L150).

integration/perf/replay_json_dataset_test.go
```diff
@@ -48,16 +48,23 @@ var submitters = flag.Int("submitters", 4, "number of goroutines submitting tran
 var orderers = flag.Int("orderers", 8, "number of goroutines submitting transactions to the orderer (BatchSubmitter workers)")
 var outstanding = flag.Int("outstanding", 1000, "maximum number of outstanding transactions")
 
+// txCompletion carries the fields needed by the refill loop after a transaction commits.
+type txCompletion struct {
+	EthTxHash common.Hash
+	Valid     bool
+	Status    committerpb.Status
```
arner, line 55: Is there a reason to have this fabric-x specific or can it also be an int, to cover both platforms? When we assign a value to it, we convert `committerpb.Status(tx.Status),`

arner (approved):
One more comment, otherwise looks good to me!

kushnireyal (approved):

PR #295, opened 2026-07-31
fix: pre-fund known Hardhat test accounts on testnode

### LINKED ISSUE #254: Pre-fund known test accounts (unblock value-transfer tests)
Part of #248.

Test EOAs are unfunded, so any value transfer fails the balance check — 93 tests in a local run of the OpenZeppelin Hardhat suite, the largest "easy" item that isn't a missing-method gap.

Balance-priming infrastructure already exists (`endorser/testimpl/balance_priming_executor.go`); wire it to fund the known Hardhat test accounts on testnode startup.

Test-RPC/testnode only — production accounts correctly start at zero; this is not a production compatibility gap, it's a test-fixture convenience.

Siddhartha-singh01:
Closes #254
Part of #248

Test EOAs on the self-contained testnode started with a zero native balance, so any path that moves value failed the balance check early. That was the largest "easy" bucket in the OZ Hardhat suite that was not a missing-method stub.

The issue mentions the existing balance-priming bits under endorser/testimpl. Those prime ERC-20 storage slots for perf/USDC work; they do not set native ETH. What we needed was real `acc:<addr>:bal` state for the known Hardhat accounts when test RPC is on.

## Code

gateway/testimpl/fund_accounts.go seeds balances into the current endorser KVS snapshot at startup (Hardhat default of 10_000 ETH). Writes are in-place on the current snapshot so we do not burn a history slot that evm_revert could later restore as empty genesis.

gateway/app/app.go calls FundTestAccounts after LoadTestAccounts when enableTestRPC is set. Production paths still start accounts at zero.

Unit coverage for FundTestAccounts plus a NewTestNode check that eth_getBalance / BalanceAt sees the funded amount for every default account.

## Baseline

Full compatible set (scripts/run_hardhat_test.sh --full) with this wired in, then reconciled testdata/oz_known_failures.json:

* insufficient-funds dropped from 94 causes to 21 (leftover cases still need hardhat_setBalance or non-default EOAs)
* about 218 more tests pass overall
* tests that used to die on zero balance now fail further down (timestamps, block number, governor clock, etc.) and are recorded as such
* cmd/baseline check is clean (no regressions, no stale entries)

### Full suite summary

```
4973 passed, 1066 failed, 1 skipped (6040 total, 82.3% passing)
```

By suite: access 354/578, crosschain 12/15, finance 8/16, governance 202/522, metatx 0/12, proxy 257/280, token 1637/1852, utils 2503/2764.

## Verification

* go test ./gateway/testimpl/ ./gateway/app/
* go test ./... -short
* go vet / staticcheck on the touched packages
* full OZ suite + baseline check as above

Happy to adjust cause tags if anything looks off. Thanks!

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,7 +44,9 @@ jobs:
   oz-hardhat-compat:
     name: OZ Hardhat Compatibility
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    # Full compatible set has grown past 15m once more tests get past early
+    # failures (e.g. funded accounts). Keep headroom for CI runner variance.
+    timeout-minutes: 25
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
--- a/gateway/app/app.go
+++ b/gateway/app/app.go
@@ -157,6 +157,13 @@ func buildApp(ctx context.Context, cfg config.Config, gwSigner sdk.Signer, logge
 			return nil, fmt.Errorf("failed to load test accounts: %w", err)
 		}
 
+		// Pre-fund known Hardhat test EOAs so value transfers pass the balance
+		// check (issue #254). Test-RPC / testnode only — production stays at zero.
+		if err := testimpl.FundTestAccounts(lightKVS, cfg.Network.Namespace, testAccountMgr.Addresses, testimpl.DefaultTestAccountBalance); err != nil {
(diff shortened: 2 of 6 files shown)


Siddhartha-singh01:
Closing to rebuild with a clean two-commit history and a single baseline refresh from a full local suite run. Will open a fresh PR for #254 once checks and the OZ baseline gate are green locally.

Issue #293, opened 2026-07-30
state queries for `earliest` return `latest`


arner:
### Summary

State queries at block `0` (`earliest` or literal `0x0`) silently return **current** state instead of genesis state. `eth_call` / `eth_getBalance` / `eth_getCode` / `eth_getStorageAt` / `eth_getTransactionCount` at `earliest` all hit this.

### Steps to Reproduce
Start a self-contained testnode:

```shell
go run ./cmd/fxevm testnode
```

Then, in another terminal,

```shell
# 1. nonce at "earliest" before any transaction: expect 0
curl -s localhost:8545 --json '{"jsonrpc":"2.0","id":1,"method":"eth_getTransactionCount","params":["0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","earliest"]}'

# 2. send a zero-value transaction to bump the nonce
curl -s localhost:8545 --json '{"jsonrpc":"2.0","id":1,"method":"eth_sendTransaction","params":[{"from":"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","to":"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","value":"0x0"}]}'

# 3. nonce at "latest": expect 1 (correct)
curl -s localhost:8545 --json '{"jsonrpc":"2.0","id":1,"method":"eth_getTransactionCount","params":["0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","latest"]}'

# 4. nonce at "earliest" again: expect 0 (genesis), actually returns 1 — bug
curl -s localhost:8545 --json '{"jsonrpc":"2.0","id":1,"method":"eth_getTransactionCount","params":["0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","earliest"]}'
```

### Root Cause

`rpcBlockNumberToBigInt` resolves `earliest` to `big.NewInt(0)` (`gateway/api/eth.go`), which reaches `resolveStateBlockNumber` (`endorser/execution/executor.go:161`) and passes through unchanged as `uint64(0)`. But `uint64(0)` is also the "give me latest" sentinel used by every `KVSSnapshotter` implementation:

- `endorser/storage/lightkvs.go:165`
- `endorser/storage/revertible_lightkvs.go:70`
- `endorser/storage/versioned_db_wrapper.go:38`

So an explicit request for block 0 is indistinguishable from "no block requested" once it reaches the KVS, and silently resolves to latest.

[...]

Issue #292, opened 2026-07-30
Implement `hardhat_mine` for real on testnode


arner:
Part of #248. Follows #251, which registered `hardhat_mine` as a stub (`gateway/testimpl/hardhat_helpers.go:54`) so callers stop hitting "method not found." The stub does not advance chain state at all — it just logs and returns. This issue is the "real block-number control" follow-up that #251 explicitly called out of scope. (Real block-*timestamp* control turns out to be a separable, bigger question — see "On `interval`" below — and stays out of scope here too.)

## Why this is safe to build

`HardhatAPI`/`EvmAPI` are only registered on the test RPC (`gateway/testimpl/server.go`), which is only wired up by `testnode` (`cmd/fxevm`, `gateway/app/testnode.go`). `NewTestNode` builds a fully in-process, ephemeral `fabrictest` network (`fabrictest.Start`, in-memory DB, in-memory `":memory:"` gateway store) — there is no real Fabric channel to pollute and no persistent state to corrupt. Any block-creation trick that would be unacceptable on a production/staging network (synthetic transactions, direct ledger injection) is fair game here.

## What "mine" needs to do

`hardhat_mine([blocks], [interval])` should actually advance the chain by `blocks` (default 1) new blocks, `interval` seconds apart (default 1s), so `eth_blockNumber` / `eth_getBlockByNumber("latest")` reflect it afterward — the way `evm_mine` (`EvmAPI.Mine`, same file, also currently a stub returning `"0x0"`) is supposed to.

## How blocks actually get created today

We use `fabrictest` (`hyperledger/fabric-x-sdk`). From that package:

- `testOrderer.Broadcast` (`fabrictest/orderer.go`) receives transaction envelopes and pushes them into `batchingLoop`.
- `batchingLoop` (`fabrictest/orderer.go`) batches by `Config.BlockCutTime`/`MaxTxPerBlock` and calls `ledger.process`. Testnode uses `fabrictest.Config{}` (all zero), so `BlockCutTime == 0` means **every envelope is flushed immediately as its own block** — one tx in, one block out.
- `ledger.process` (`fabrictest/ledger.go`) is a no-op for an empty batch (`flush()` returns early if `len(batch) == 0`, `orderer.go:87`) — **you cannot cut an empty block through the existing envelope path**, only through a real transaction.
[...]

PR #296, opened 2026-07-31
fix: pre-fund known Hardhat test accounts on testnode

### LINKED ISSUE #254: Pre-fund known test accounts (unblock value-transfer tests)

Siddhartha-singh01:
Closes #254
Part of #248

Test EOAs on the self contained testnode started with a zero native balance, so any path that moves value failed the balance check early. That was the largest "easy" bucket in the OZ Hardhat suite that was not a missing method stub.

The issue mentions the existing balance priming bits under endorser/testimpl. Those prime ERC20 storage slots for perf/USDC work; they do not set native ETH. What we needed was real acc:<addr>:bal state for the known Hardhat accounts when test RPC is on.

## Code

gateway/testimpl/fund_accounts.go seeds balances into the current endorser KVS snapshot at startup (Hardhat default of 10_000 ETH). Writes are in place on the current snapshot so we do not burn a history slot that evm_revert could later restore as empty genesis.

gateway/app/app.go calls FundTestAccounts after LoadTestAccounts when enableTestRPC is set. Production paths still start accounts at zero.

Unit coverage for FundTestAccounts plus a NewTestNode check that BalanceAt sees the funded amount for every default account.

Also raised the OZ hardhat CI job timeout from 15m to 25m. With accounts funded, more of the suite gets past early insufficient funds failures and needs more wall time on the runners.

## Baseline

Full compatible set (scripts/run_hardhat_test.sh --full) with this wired in, then reconciled testdata/oz_known_failures.json:

* insufficient funds dropped from 94 causes to 21 (leftover cases still need hardhat_setBalance or non default EOAs)
* about 218 more tests pass overall
* tests that used to die on zero balance now fail further down (timestamps, block number, governor clock, etc.) and are recorded as such
* cmd/baseline check is clean (no regressions, no stale entries)

### Full suite summary

```
4973 passed, 1066 failed, 1 skipped (6040 total, 82.3% passing)
```

By suite: access 354/578, crosschain 12/15, finance 8/16, governance 202/522, metatx 0/12, proxy 257/280, token 1637/1852, utils 2503/2764.

## Verification

* go test ./gateway/testimpl/ ./gateway/app/
* go test ./... -short
* gofmt / go vet / staticcheck on the touched packages
* full OZ suite + baseline check as above

Happy to tweak cause tags if anything looks off. Thanks!

### DIFF
--- a/gateway/testimpl/fund_accounts.go
+++ b/gateway/testimpl/fund_accounts.go
@@ -1,0 +1,87 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+
+WARNING: This package contains test-only/unsafe RPC implementations.
+DO NOT use in production environments.
+*/
+
+package testimpl
+
+import (
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/params"
+	estorage "github.com/hyperledger/fabric-x-evm/endorser/storage"
+)
+
+// DefaultTestAccountBalance is Hardhat Network's default genesis balance per
+// account: 10_000 ETH. Matches @nomicfoundation/hardhat-network-helpers and
+// testdata/openzeppelin-contracts/test/helpers/account.js.
+var DefaultTestAccountBalance = new(big.Int).Mul(big.NewInt(10_000), big.NewInt(params.Ether))
+
+// FundTestAccounts seeds native ETH balances for the given addresses into the
+// endorser KVS. Intended for test-RPC / testnode startup only: production
+// accounts correctly start at zero.
+//
+// Writes land in the current snapshot in place (no history advance), so the
+// funded balances look like genesis state and survive later Handle/Update
+// clones as well as reverts that restore pre-funding block numbers.
+//
+// kvs must be *estorage.LightKVS or *estorage.RevertibleLightKVS (memory DB).
+func FundTestAccounts(kvs estorage.KVS, namespace string, addresses []common.Address, balance *big.Int) error {
+	if balance == nil || balance.Sign() <= 0 {
+		return nil
+	}
+	if len(addresses) == 0 {
+		return nil
+	}
+
+	snap, err := currentSnapshot(kvs)
+	if err != nil {
+		return err
+	}
+
+	balBytes := balance.Bytes()
+	for _, addr := range addresses {
+		// Key layout matches LightKVS.Reader.Get / collectWrites:
+		// fullKey = namespace + ":" + accKey(addr, "bal")
+		// accKey  = "acc:" + addr.Hex() + ":bal"
+		key := namespace + ":acc:" + addr.Hex() + ":bal"
+		// Copy bytes so callers can reuse the balance buffer later.
+		val := append([]byte(nil), balBytes...)
+		snap.Data[key] = &estorage.ValueVersion{
+			Value:    val,
+			BlockNum: snap.BlockNumber,
+			TxNum:    0,
+			Version:  0,
+			TxID:     "test-account-funding",
+		}
+	}
+	return nil
+}
+
+func currentSnapshot(kvs estorage.KVS) (*estorage.Snapshot, error) {
+	switch k := kvs.(type) {
+	case *estorage.RevertibleLightKVS:
+		if k.LightKVS == nil {
+			return nil, fmt.Errorf("fund test accounts: RevertibleLightKVS has nil LightKVS")
+		}
+		snap := k.Current.Load()
+		if snap == nil {
+			return nil, fmt.Errorf("fund test accounts: KVS has no current snapshot")
+		}
+		return snap, nil
+	case *estorage.LightKVS:
+		snap := k.Current.Load()
+		if snap == nil {
+			return nil, fmt.Errorf("fund test accounts: KVS has no current snapshot")
+		}
+		return snap, nil
+	default:
+		return nil, fmt.Errorf("fund test accounts: KVS type %T does not support in-place funding (need memory LightKVS)", kvs)
+	}
+}
--- a/gateway/testimpl/fund_accounts.go
+++ b/gateway/testimpl/fund_accounts.go
@@ -10,12 +10,17 @@ DO NOT use in production environments.
 package testimpl
 
 import (
+	"context"
 	"fmt"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/tracing"
 	"github.com/ethereum/go-ethereum/params"
+	"github.com/holiman/uint256"
+	"github.com/hyperledger/fabric-x-evm/endorser/execution"
 	estorage "github.com/hyperledger/fabric-x-evm/endorser/storage"
+	"github.com/hyperledger/fabric-x-sdk/blocks"
 )
 
 // DefaultTestAccountBalance is Hardhat Network's default genesis balance per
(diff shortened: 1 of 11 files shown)


Siddhartha-singh01:
@arner Sir review this And if you feel any kind of stuff or any kind of changes then please tell me

gateway/testimpl/fund_accounts.go
```diff
@@ -0,0 +1,87 @@
...
+// kvs must be *estorage.LightKVS or *estorage.RevertibleLightKVS (memory DB).
+func FundTestAccounts(kvs estorage.KVS, namespace string, addresses []common.Address, balance *big.Int) error {
+	if balance == nil || balance.Sign() <= 0 {
+		return nil
+	}
+	if len(addresses) == 0 {
+		return nil
+	}
+
+	snap, err := currentSnapshot(kvs)
```
arner: Instead of getting the current snapshot and then building and "committing" the key/value by hand, maybe we can:

- construct an execution.NewStateDB
- CreateAccount (?), SetBalance for each of the accounts
- get the Result() read write set and wrap it in a synthetic block
- Pass it in the Handle() function of the KVS backend.

This removes the need for the currentSnapshot switch (and makes the code compatible with any backend), and generally is safer because of using the intended interfaces (for instance, the current code bypasses the Update() function and hardcodes version 0).

arner (changes requested):
Thanks @Siddhartha-singh01 , great results! I have a suggestion that could make it a bit more robust, let me know if it works out.

gateway/testimpl/fund_accounts.go
Siddhartha-singh01: Done in 8fbc9a3.

FundTestAccounts now goes through NewStateDB, CreateAccount + AddBalance, Result(), then a synthetic block into KVS.Handle. No more currentSnapshot type switch or hand written keys. Versions come from Update like a normal commit.

Unit tests and TestNewTestNode_PrefundsHardhatAccounts are green.

Siddhartha-singh01:
@arner thanks for the review. I reworked the funding path the way you suggested.

We no longer touch LightKVS snapshot maps directly. The code builds accounts with NewStateDB, CreateAccount and AddBalance, takes Result(), and commits a synthetic block through KVS.Handle. That uses the normal Update versioning and works with any Handle backed KVS.

Pushed in 8fbc9a3. Tests for fund_accounts and NewTestNode prefund pass. Happy to adjust further if something still looks off.

gateway/testimpl/fund_accounts.go
```diff
@@ -0,0 +1,101 @@
...
+var DefaultTestAccountBalance = new(big.Int).Mul(big.NewInt(10_000), big.NewInt(params.Ether))
+
+// FundTestAccounts seeds native ETH balances for the given addresses into the
+// endorser KVS. Intended for test-RPC / testnode startup only: production
+// accounts correctly start at zero.
+//
+// Balances are applied through the normal StateDB write path and committed with
+// KVS.Handle on a synthetic block, so versions and keys match real commits and
+// any KVS backend that implements Handle works (not only LightKVS in place).
+func FundTestAccounts(kvs estorage.KVS, namespace string, addresses []common.Address, balance *big.Int) error {
```
arner: perhaps we can pass the context into this function instead of creating it

gateway/testimpl/fund_accounts_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/hyperledger/fabric-x-evm/endorser/execution"
+	estorage "github.com/hyperledger/fabric-x-evm/endorser/storage"
+)
+
+const testNS = "basic"
+
+func TestDefaultTestAccountBalance_IsHardhatDefault(t *testing.T) {
```
arner: This test is just repeating the actual assignment so it's not really useful, I think we can safely remove it.

gateway/testimpl/fund_accounts_test.go
```diff
@@ -0,0 +1,230 @@
...
+		t.Fatalf("FundTestAccounts: %v", err)
+	}
+
+	reader, err := kvs.NewSnapshot(0)
+	if err != nil {
+		t.Fatalf("NewSnapshot: %v", err)
+	}
+	defer reader.Close()
+
+	stateDB, err := execution.NewStateDB(context.Background(), reader, testNS, 0, true)
```
arner: Using `t.Context()` will help with safe closing.

gateway/testimpl/fund_accounts_test.go
```diff
@@ -0,0 +1,230 @@
...
+		if rec == nil || len(rec.Value) == 0 {
+			t.Fatalf("missing balance for %s", addr.Hex())
+		}
+		got := new(big.Int).SetBytes(rec.Value)
+		if got.Cmp(DefaultTestAccountBalance) != 0 {
+			t.Errorf("%s balance = %s, want %s", addr.Hex(), got, DefaultTestAccountBalance)
+		}
+		// Handle path must assign a real version (not a hardcoded 0 forever).
+		if rec.Version != 0 {
+			// First write is version 0; fine. Just ensure field is populated.
```
arner: If it's fine, let's remove this whole if statement

testdata/oz_known_failures.json
arner, line 1: Any idea why `Governor using $ERC20VotesTimestampMock should revert on vote by signature "before each" hook for "if signature does not match signer"` now succeeds? Is that a real fix or a flakiness issue? If it's a real fix, can you recreate the known failures json once more?

arner (changes requested):
Thanks @Siddhartha-singh01 , left a couple minor comments.

Siddhartha-singh01:
@arner thanks for the notes. All of them are in edb1a42.

FundTestAccounts now takes a context. Removed the constant only test, switched the tests to t.Context, and dropped the empty version check.

On the Governor entry: with accounts funded the before each hook finishes, so that hook title is no longer a failure. The two vote by signature tests then run and fail because eth_signTypedData_v4 is still missing on this branch. I removed the stale hook entry and added those two with the eth_signTypedData_v4 cause. Not flakiness, just the suite getting further past setup.

Unit tests for fund accounts and the testnode prefund check are green. CI should be happy on this push.

Siddhartha-singh01:
@arner GitHub is not starting the Tests workflow on this PR anymore. New pushes only get DCO. The last full run (f1727c2) died before checkout with Actions Service Unavailable. Locally unit tests and baseline check are green after the main merge and baseline refresh.

Could you Approve and run workflows on the latest head, or re run Tests when you have a moment? Thanks.

arner (approved):
Yes, the CI is happy and so am I :)

Thanks for your contribution!

If you can rebase (sorry, an older PR came inbetween), we can merge it!

Siddhartha-singh01:
@arner this is rebased onto current main (includes the pebble KVS work).

CI was green before the main catch up. I also fixed a compile break in gateway/config Config.Validate that came in with #283 (missing for loop brace). Unit tests and baseline check are green on the updated branch.

Ready to merge when you are. Thanks for the reviews.

Siddhartha-singh01:
@arner once this is in, I am free to pick up more work. Any issue you would like me on next? Happy to take an assignment when you have something ready.

Siddhartha-singh01:
@arner sir can we merge this

arner:
Whoops the missing for loop brace was my mistake, thanks for spotting and fixing it! 🙏 

Now, there's one proper issue open on #248, and that's #292 but it depends on an update on the SDK (https://github.com/hyperledger/fabric-x-sdk/issues/51). If you feel like it, you're more than welcome clone the SDK and create a PR for that!

I saw you also replied to #293 which is also a good one, so whichever you prefer. Thanks again for your help! 👍

Issue #276, opened 2026-07-24
EIP-7610 create-collision produces wrong post-state root


arunabha003:
## Description
`CREATE`/`CREATE2` into an address that already holds storage (the EIP-7610
create-collision case) produces a post-state root that diverges from the
reference EVM. Surfaced by the execution-specs suite (#237); four fixtures are
quarantined in `testdata/execution_specs_tests.skip`:
`collision_with_create2_revert_in_initcode`, `create2_collision_storage`,
`init_collision_create_opcode`, `init_collision_create_tx`.

## Reproduction
Remove the `eip7610_create_collision` entries from
`testdata/execution_specs_tests.skip`, then:

    make fetch-execution-specs-tests
    go test ./integration -run 'TestExecutionSpecStateTests/(collision_with_create2_revert_in_initcode|create2_collision_storage|init_collision_create_opcode|init_collision_create_tx)\.json' -v

## Expected vs actual
- Expected: post-state root matches the fixture (a CREATE onto an address with
  pre-existing storage must fail per EIP-7610).
- Actual: `post state root mismatch: got …, want …`.

## Root-cause pointer
`GetStorageRoot` is a stub returning the zero hash, so the EIP-7610 collision
guard is effectively disabled (a contract can be deployed onto an address that
already has storage). Already noted in `docs/COMPATIBILITY.md` (EVM execution
differences).

arner:
I think this is blocked by https://github.com/hyperledger/fabric-x-evm/issues/85, would you agree?

arunabha003:
Hi @arner , yes I checked , it's blocked by #85.  
Once I'm done with #216 , if no one is working on that by then I can take up both #85 and this issue.

arunabha003:
Update : turns out this is not blocked by https://github.com/hyperledger/fabric-x-evm/issues/85 (correcting my last comment). We run go-ethereum v1.17.3's EVM create path unchanged, and geth implements EIP-7610 as a hardcoded list of 28 mainnet addresses, not a general "has storage" check , the guard never calls GetStorageRoot, so https://github.com/hyperledger/fabric-x-evm/issues/85 wouldn't fix it.
The 4 fixtures use a synthetic address with storage that isn't in geth's list, so any geth-based client diverges here, not just us. Fixing it would mean forking geth's create-collision logic , its a big deviation for an upstream divergence, not our bug.

[...]

PR #297, opened 2026-08-03
test: fix nil eth StateDB panic for empty pre-state fixtures

### LINKED ISSUE #275: nil-pointer panic in DualStateDB.Snapshot for a CREATE tx from a non-existent sender
## Description
The execution-specs conformance fixture `no_src_account_create1559` — a CREATE
transaction whose sender account does not exist — triggers a nil-pointer panic
in `DualStateDB.Snapshot`. Surfaced by the execution-specs suite (#237); the
fixture is currently quarantined in `testdata/execution_specs_tests.skip`.

## Reproduction
Remove the `no_src_account_create1559` entries from
`testdata/execution_specs_tests.skip`, then:

    make fetch-execution-specs-tests
    go test ./integration -run 'TestExecutionSpecStateTests/no_src_account_create1559.json' -v

## Expected vs actual
- Expected: the transaction is handled normally (executed/rejected) without crashing.
- Actual: `panic: runtime error: invalid memory address or nil pointer dereference`
  (SIGSEGV) in `DualStateDB.Snapshot` → go-ethereum `state.StateDB.Snapshot`.

## Root-cause pointer
`endorser/execution/dual_statedb.go` — `DualStateDB.Snapshot`: the inner eth
StateDB is nil for this transaction shape (a CREATE from a sender with no
pre-existing account/state).

arunabha003:
### Addressed issue

Closes #275 — nil-pointer panic on the `no_src_account_create1559` conformance
fixture (a CREATE from a sender with no pre-existing account).

### Summary

The panic was a **test-harness** bug, not a production `DualStateDB` issue as the
ticket initially suspected. `primeGenesisAlloc` (in `integration/ethereum_test.go`)
returned early when a fixture had no pre-state accounts — but building the
pre-state is also what creates and sets the eth StateDB the harness later reads.
Skipping it left that StateDB `nil`, so the pre-validation `nonceReader` panicked
in `GetNonce` for a sender that doesn't exist yet.

- Drop the `if len(pre) == 0 { return nil }` early-return so an empty pre-state
  still builds and sets an empty eth StateDB. The priming loop is a no-op for zero
  accounts; the rest of the flow is unchanged.
- With that in place the fixture passes (state-root check included), confirming the
  production path (`core.ValidateTx`, `DualStateDB`, execute/commit) already handled
  this case correctly — so no production code changes were needed.
- Remove the 3 `no_src_account_create1559` entries from `testdata/eth_tests.skip`
  and drop the corresponding note from `COMPATIBILITY.md`.

Test-harness/skip/docs only — no production/runtime code changes.

### Validation done

- The previously-panicking fixture now passes:
  `go test ./integration -run 'TestEthereumTests/no_src_account_create1559.json' -v` → PASS.
- `make checks` passes.
- Full `make eth-tests` is green (both suites, 0 failures) with the fixtures now
  un-skipped — nothing else regressed.
- Skip list drops from 15 → 12 entries (only the EIP-7610 group remains, tracked
  separately in #276).
  
### Screenshot
  
image

### DIFF
--- a/docs/COMPATIBILITY.md
+++ b/docs/COMPATIBILITY.md
@@ -234,8 +234,7 @@ cases. These divergences are deliberate consequences of the modern-only semantic
 regressions.
 
 These known-divergent cases are quarantined in `testdata/eth_tests.skip`: the EIP-7610
-create-collision fixtures above, and a `CREATE` transaction from a non-existent sender (currently a
-nil-pointer panic in `DualStateDB.Snapshot`). All other Osaka-forward state tests pass.
+create-collision fixtures above. All other Osaka-forward state tests pass.
 
 **Native ETH balances not funded**: balances are implemented but unused. Accounts have zero ETH 
 balance by default. Value transfers inside the EVM (`CALL` with value, `SELFDESTRUCT` beneficiary, 
--- a/integration/ethereum_test.go
+++ b/integration/ethereum_test.go
@@ -480,10 +480,10 @@ func newEthereumTestHarness(t *testing.T, evmConfig execution.EVMConfig, pre typ
 
 // primeGenesisAlloc primes the state from ethereum genesis format and sets up ethStateDB for wrappers
 func (eth *ethereumTestHarness) primeGenesisAlloc(ctx context.Context, pre types.GenesisAlloc, wait bool) error {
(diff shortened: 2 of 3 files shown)


arner (approved):
LGTM!

PR #305, opened 2026-08-07
fix: distinguish earliest from latest in KVS snapshots

### LINKED ISSUE #293: state queries for `earliest` return `latest`

Siddhartha-singh01:
Fixes #293

NewSnapshot treated block 0 as latest, so eth_call and related state reads at earliest returned current state. That made earliest and latest the same thing.

Height is a pointer now: nil means latest, any non nil value (including 0) means that exact block. Same idea across LightKVS, RevertibleLightKVS, PebbleKVS, VersionedDBWrapper, and the execution path that resolves block tags.

Integration clients fetch nonce at latest (nil) instead of big.NewInt(0). Zero used to mean latest by accident and would break once 0 meant real genesis.

## Test plan

* go test ./endorser/storage/
* go test ./endorser/execution/
* go test ./gateway/testimpl/
* go build ./cmd/fxevm

### DIFF
--- a/endorser/testimpl/executor_constructors_test.go
+++ b/endorser/testimpl/executor_constructors_test.go
@@ -1,0 +1,82 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+*/
+
+package testimpl
+
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/rawdb"
+	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/triedb"
+	fxcommon "github.com/hyperledger/fabric-x-evm/common"
+	"github.com/hyperledger/fabric-x-evm/endorser/execution"
+	"github.com/hyperledger/fabric-x-evm/endorser/storage"
+)
+
+func testEVMConfig() execution.EVMConfig {
+	return execution.EVMConfig{ChainConfig: fxcommon.BuildChainConfig(4011)}
+}
+
+func emptyEthStateDB(t *testing.T) *state.StateDB {
+	t.Helper()
+	trieDB := triedb.NewDatabase(rawdb.NewMemoryDatabase(), nil)
+	sdb := state.NewDatabase(trieDB, nil)
+	ethStateDB, err := state.New(types.EmptyRootHash, sdb)
+	if err != nil {
+		t.Fatalf("state.New: %v", err)
+	}
+	return ethStateDB
+}
+
+func TestNewBalancePrimingExecutor_UsesLatestSnapshot(t *testing.T) {
+	kvs := storage.NewLightKVS(8)
+	ex, err := NewBalancePrimingExecutor("testns", kvs, testEVMConfig(), true, nil, nil)
+	if err != nil {
+		t.Fatalf("NewBalancePrimingExecutor: %v", err)
+	}
+	if ex == nil || ex.Executor == nil {
+		t.Fatal("expected non-nil executor")
+	}
+	ex.Close()
+}
+
+func TestNewBalancePrimingExecutor_WithPrimingEnabled(t *testing.T) {
+	kvs := storage.NewLightKVS(8)
+	cfg := &BalancePrimingConfig{
+		Enabled:         true,
+		ContractAddress: common.HexToAddress("0x1111111111111111111111111111111111111111"),
+		MappingPosition: 0,
+	}
+	ex, err := NewBalancePrimingExecutor("testns", kvs, testEVMConfig(), true, cfg, nil)
+	if err != nil {
+		t.Fatalf("NewBalancePrimingExecutor: %v", err)
+	}
+	if ex == nil || ex.Executor == nil {
+		t.Fatal("expected non-nil executor")
+	}
+	if _, ok := ex.state.(*BalancePrimingWrapper); !ok {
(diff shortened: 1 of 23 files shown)


endorser/testimpl/executor_constructors_test.go
```diff
@@ -0,0 +1,82 @@
...
+func TestNewBalancePrimingExecutor_UsesLatestSnapshot(t *testing.T) {
+	kvs := storage.NewLightKVS(8)
+	ex, err := NewBalancePrimingExecutor("testns", kvs, testEVMConfig(), true, nil, nil)
+	if err != nil {
+		t.Fatalf("NewBalancePrimingExecutor: %v", err)
+	}
+	if ex == nil || ex.Executor == nil {
+		t.Fatal("expected non-nil executor")
+	}
+	ex.Close()
```
arner: Probably better to use `t.Cleanup` so we're sure it also closes on a fatal error. Also for the other tests in this file. (In the other file you use defer which is also fine, though I have a slight preference for the more idiomatic t.Cleanup).

integration/state_test_util_test.go
```diff
@@ -0,0 +1,35 @@
...
+		types.GenesisAlloc{
+			addr: {Balance: big.NewInt(1), Nonce: 0},
+		},
+		false,
+		rawdb.HashScheme,
+	)
+	if st.StateDB == nil {
+		t.Fatal("expected non-nil StateDB")
+	}
+	st.Close()
```
arner: same as above

integration/ethclient.go
arner: Good find that the bn is only used for MakeSigner. I propose to simplify the code and change the MakeSigner code to:

```go
ethSigner := types.NewEIP155Signer(e.ethChainConfig.ChainID)
```

And then we don't need this function anymore.

Siddhartha-singh01:
@arner thanks for the notes.

Switched the constructor tests over to t.Cleanup, same for the dual state helper test. EthClient now signs with types.NewEIP155Signer on the chain id and GetCtxForSigner is gone.

Should be ready for another look when you have a minute.

Siddhartha-singh01:
@arner  Actually Sir I was thinking that there was some kind of error in CICD

arner (approved):

arner:
Yeah... and ominous that it's about verifying a signature! My first thought was that maybe we should revert the GetCtxForSigner change, but that's not even used in the OZ tests. It turns out to be a flaky test. I'll adapt the test harness soon, for now we can ignore it.

Issue #21, opened 2026-03-26
Split Gateway and Endorser Services


ale-linux:
Refactor the monolithic application into two separate services with distinct responsibilities: the Gateway service handling client-facing Ethereum JSON-RPC APIs and transaction orchestration, and the Endorser service handling EVM execution and proposal endorsement. This separation improves scalability by allowing independent scaling of each component, enhances security by isolating execution from external client access, and enables deployment flexibility where multiple gateways can share a pool of endorsers. Create separate main.go entry points for each service, define clear service boundaries and interfaces, and update deployment documentation to reflect the new architecture.

---

**Quarter:** Q2

Issue #10, opened 2026-03-26
[EPIC] Refactoring and Code Hygiene


ale-linux:
Epic tracking issue for Refactoring and Code Hygiene tasks.

PR #306, opened 2026-08-07
integration: prove a 2-of-2 endorsement policy over a real backend

### LINKED ISSUE #302: integration: prove a 2-of-2 endorsement policy across a distributed deployment
Related to #21 (Split Gateway and Endorser Services). Builds on #246 

## Summary

#245 proved the gRPC endorsement path (server, client, real mTLS) with one endorser. #246 made split deployment (`Gateway.Endorsers`) config-driven and live, but nothing has exercised it with more than one endorser yet. This proves a real 2-of-2 case: two independently-running endorsers, both required to endorse, driven through the actual `gateway/app.App` split-deployment path - not a hand-built `EndorsementClient`.

## What "2-of-2" means here

Endorsement, not commitment. `Gateway.ExecuteEthTx(ctx, tx)` calls `g.endorsers.ExecuteTransaction(ctx, tx)` and returns before anything is submitted to an orderer - so this test proves the *endorsement* fan-out requires and collects a signed response from both endorsers, over real gRPC and real mTLS. It does not need a live orderer or committer, since those connections are dialed lazily and are never required to actually respond for this phase to complete.

## Scope

- Stand up two independent `core.Endorser` instances (separate state, like two orgs), each behind its own real `endorser/server.Server` over real mTLS.
- Build a real `config.Config` with `Gateway.Endorsers` pointing at both, and construct a real `*App` via `app.NewWithSigner` (the actual split-deployment path from #246).
- Call `app.Gateway().ExecuteEthTx(ctx, tx)` and assert exactly two responses come back, both successful, from two distinct endorser identities.
- `app.Shutdown()` at the end, exercising the split-mode connection cleanup path too.

## Modularity

Test-only, extending #245's pattern (self-signed certs, real `core.Endorser`, no stubs) through the real, now-live split-deployment path. No production code changes. Fully self-contained - no docker, no real orderer/committer process needed.

## Done when

- A transaction is endorsed by two independently-running endorsers over real mTLS, driven through the real `gateway/app.App` split-deployment path.
- `main` remains green.

Storm1289:
## Summary

Closes #302. Related to #21 (Split Gateway and Endorser Services).

End-to-end proof that a 2-of-2 endorsement policy is accepted by the real Fabric-X backend, with each organization's endorser reached over real gRPC. Includes the `endorser` / `gateway.endorsers` naming change from review, the Fablo setup moved to the same shape, and two bugs this surfaced in the already-merged split-deployment path.

### Added

| File | Purpose |
| --- | --- |
| `integration/fabx-2of2.yaml` | Gateway config, split mode: dials both endorsers, embeds none |
| `integration/fabx-2of2-org1.yaml` | Standalone org1 endorser (one `endorser`) |
| `integration/fabx-2of2-org2.yaml` | Standalone org2 endorser (one `endorser`) |
| `integration/fablo-org1.yaml` | Standalone org1 endorser for Fablo |
| `integration/fablo-org2.yaml` | Standalone org2 endorser for Fablo |
| `integration/split_endorsers.go` | Shared helpers to serve endorsers over mTLS gRPC and build a split gateway |

### Modified

| File | Change |
| --- | --- |
| `gateway/config/config.go` | `Endorsers []` becomes `Endorser *`; reject both keys set, or neither |
| `gateway/app/app.go` | `newApp` builds one embedded endorser instead of a list |
| `endorser/client/dial.go` | Apply `tls.server-name` when dialing (fix 2) |
| `endorser/client/client.go` | Restore the signed payload from the wire (fix 1) |
| `endorser/server/handlers.go` | Send the signed payload over the wire (fix 1) |
| `integration/fablo.yaml` | Now gateway-only, split across the two org files |
| `integration/fabx.yaml`, `fabx-full.yaml`, `fablo.hardhat.yaml` | `endorsers:` renamed to `endorser:` |
| `integration/test_helpers.go` | `newSplitFileConfigHarness`; build the single embedded endorser |
| `integration/integration_test.go` | `TestFablo` uses the split harness; 2-of-2 subtest |
| `integration/endorsement_grpc_test.go` | Parity test now asserts the signed payload |
| `endorser/client/dial_test.go` | Hostname-only certificate cases for fix 2 |
| `gateway/config/validate_test.go`, `load_test.go` | Cover the singular field and the both-keys error |
| `Makefile` | `start-x` also creates the `basic2of2` namespace |
| `.github/workflows/tests.yml` | Raise job timeouts; create the artifact dir before writing to it |

## Config: `endorser` singular, `gateway.endorsers` plural

`endorsers` previously meant two different things. Now `endorser` (singular, top level) is this process's own embedded endorser, and `gateway.endorsers` (plural) are remote ones dialed over gRPC. Setting both is rejected. Mixed mode is deliberately not implemented, per review.

Since the field is a single object rather than a list, no config can embed more than one endorser. A file per organization is not the same as embedding several: each org file declares exactly one `endorser`, and they run as separate endorsers that a gateway dials.

## Test setup

**Fabric-X** - org1 and org2 each run as a standalone endorser behind its own mTLS gRPC server, using that org's real `endorser.<org>.example.com` material, with a split-deployment gateway dialing both. Drives the real `gateway/app.App`, the same type `fxevm start` uses. `start-x` creates a `basic2of2` namespace with `AND('Org1MSP.member','Org2MSP.member')`, separate from `basic` so the stricter policy doesn't affect the existing single-org cases.

**Fablo** - split the same way. Its chaincode declares no endorsement policy, so Fablo commits it with `AND('Org1MSP.member','Org2MSP.member')` and both orgs must endorse, which is why it embedded two endorsers before and could not stay singular.

`fabx.yaml` deliberately stays embedded with a single `endorser`, so the embedded deployment mode keeps end-to-end coverage against a real backend.

`newSplitFileConfigHarness` keeps each endorser's KVS registered as a block handler and its builder available to the state primer, so `tether_token_replay` still works, while only the gateway's path to them is gRPC - the part under test.

## Fix 1: the signed endorsement payload was dropped over gRPC

`peer.ProposalResponse` has a top-level `Payload` - the signed result `fabricx.CreateTx` assembles the transaction from - which is a different field from `Response.Payload` (EVM return data). Only the latter crossed the wire, so the top-level one arrived empty: 456 bytes in-process, 0 bytes over gRPC.

An empty payload still unmarshals cleanly, so nothing reported an error: the gateway packaged an empty transaction, the orderer accepted it and cut a block, and it was then rejected. Split deployment could endorse and submit but never commit.

Fixed by carrying it in `ExecuteResponse.read_write_set`, the field `docs/design/endorsement-api/01-api-and-proto.md` already specifies for this ("kept byte-exact for signing"), so no proto change was needed. `TestGRPCEndorsement_Parity` had compared only `Response.Payload` - empty on both sides, so trivially equal - and now asserts the signed payload is non-empty and identical across the boundary.

## Fix 2: `tls.server-name` was ignored when dialing an endorser

`ClientConfig.TLS.ServerName` exists and both `ToPeerConf` and `ToOrdererConf` propagate it, but `eclient.Dial` dropped it, so certificates were verified against the dial address:

`x509: cannot validate certificate for 127.0.0.1 because it doesn't contain any IP SANs`

This stayed hidden because the Fabric-X test endorser certs happen to carry `IP:127.0.0.1` SANs. Real peer certificates do not - Fablo's are `DNS:peer0.orgN.example.com` only - so any deployment whose endorser cert is issued for a hostname and reached by IP would fail. Fixed with `grpc.WithAuthority`, covered by a test that serves a hostname-only certificate and fails with that exact error without the fix.

## Verification

Against real backends, not mocks: `TestFabricX` 11/11 (including the new `two_of_two_endorsement_policy`) and `TestFablo` 10/10, both on freshly created networks, re-run after rebasing onto the `main` merges that brought in #305 and #294. `make checks` clean and the full `-short` suite green.

`tether_token_replay` requires a fresh network in both suites - it primes state and replays a historic mainnet transaction. That behaviour is unchanged here and reproduces identically on `main`.

## Test plan

- [x] `go build ./...` / `go vet ./...` / `gofmt -l` (clean)
- [x] `make checks`
- [x] `go test -short ./...`
- [x] `make test-x` (11/11)
- [x] `make test-fablo` (10/10)


Storm1289:
hi @arner, whenever you have sometime please review this pr :)

Storm1289:
Hi @arner, heads up for the OZ Hardhat baseline gate looks flaky around *TimestampMock Governor tests (ERC20 and ERC721 variants) - same tests flipping pass/fail across runs, likely tied to the documented fixed-timestamp limitation. Not related to this PR (Makefile/CI/integration-test only). Let me know how you'd like to handle it.

arner:
Thanks @Storm1289 you're right about the flaky tests (though not sure if the timestamp is the issue - that should be consistent). I think it may have to do with how hardhat tests are forced to be sequential (and build on the previous tx result), even when pushed at almost the same time. That would be exactly the kind of issue that the dependency manager should be able to solve!

In the meantime I'm considering to add a flakiness flag to the tests; right now we can't really deal with flapping tests, we expect either failure or success...

arner:
It looks good, but these are two embedded endorsers. I was hoping that we can make a test over GRPC, with two instances of an endorser (can be the full application including gateway). It would at least mean two config files; one for each organization. That would make it a step closer to a real, multi-org deployment.

We can still run it in one process (easy to start with the integration tests). At some later point we should also add a sample deployment with several organizations, and if we consider it worthwhile we could test against that too.

What do you think?

Storm1289:
Thanks @arner for the review.
Done two standalone endorsers now, each behind its own mTLS gRPC server, with a split-deployment gateway dialing both. One config file per org, driving the real `gateway/app.App`. Still one process, as you suggested.

This surfaced a bug in the merged gRPC path: the endorsed payload wasn't crossing the wire, so split deployment could endorse and submit but never commit. Two lines to fix, plus the parity-test assertion that would have caught it. Kept it here since the new test can't pass without it.

Could you add the `want-integration-tests` label? The 2-of-2 test only runs in the integration job, so without it CI goes green without it executing.

integration/fabx-2of2-org2.yaml
```diff
@@ -0,0 +1,35 @@
...
+  protocol: fabric-x
+  channel: mychannel
+  namespace: basic2of2
+  ns-version: "1.0"
+  chain-id: 4011
+
+logging:
+  spec: info
+
+endorsers:
```
arner: I have to admit that it's still a bit difficult to understand, don't you agree?  It's mainly the fact that "endorsers" can mean two things (embedded or remote). In your example you run two pure endorsers with no gateway, and one gateway without endorsers. I think that's a valid setup. But it is also possible to have one embedded and also connect to several remotely (it is, right?). In a real application we would never embed more than one endorser. 

Maybe it would become more clear if we make the configuration unambiguous: "endorser" (singular) for the configuration of the embedded endorser, and "endorsers" (plural) to point to the remote endorsers. What do you think?

Storm1289: Agreed, singular/plural reads much clearer.

To answer your question: mixing isn't possible today. If `gateway.endorsers` is set, `newApp` returns the split app and the top-level `endorsers` is ignored strictly either/or. Happy to make mixed mode work if you want it, but that's a behaviour change rather than a rename.

One constraint: `integration/fablo.yaml` embeds two endorsers for the Fabric integration test, so strict singular means moving that test to the split setup as well same shape as this PR. Fine by me, just larger than the rename alone.

Unless you'd rather it went another way, I'll keep `gateway.endorsers` where it is: `endorser:` for the embedded one, `gateway.endorsers:` for remote.

arner: Agree with `gateway.endorsers` for remote, and `endorser` for the "own" endorser settings. And yes that would include the Fablo setup. Can you that for this PR? It probably means that some alignment (for instance in namespace name and test setup?) can be done between fablo and your split setup.

Let's leave the mixed mode for later. For now, error if both keys are set. @ale-linux , this is touching on our deployment modes, let's discuss the approach.

Storm1289: okay

Storm1289:
Hi @arner, updated the PR description. All three points are in: `endorser` (singular) for the embedded one, `gateway.endorsers` for remote, and an error if both are set. Fablo moved to the split setup as well. Doing that surfaced a second bug in the merged gRPC path, described in the description.

gateway/config/config.go
```diff
@@ -104,24 +111,30 @@ func (cfg Config) Validate() error {
...
-
-			// Checked here rather than in Endorser.Validate because the backend and
-			// the protocol it has to agree with live at different config levels.
-			// Skipped when the protocol itself is already invalid, so one bad
-			// protocol value is not also reported once per endorser.
-			if protocolErr == nil {
-				if err := endorser.ValidateDatabaseProtocol(cfg.Endorsers[i].Database.Database, cfg.Network.Protocol); err != nil {
-					errs = append(errs, fmt.Errorf("endorsers[%d]: %w", i, err))
-				}
+	// The two deployment modes are mutually exclusive. Mixing an embedded
```
arner, line 114: Propose to change to:

```
// The two deployment modes are mutually exclusive. Mixing an embedded
// endorser with remote ones is not supported yet.
```

arner (approved):
LGTM! @ale-linux can you take a look as well? Maybe we can discuss the setup in our call. The "mixed mode" should not be too hard to add and we can consider going for that for at least org1 (so we have only two config files per org). But that would be a separate PR anyway IMO.

gateway/config/config.go
Storm1289: done

.github/workflows/tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
   quick-tests:
     name: Quick Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 35
```
ale-linux: Hmm - a quick test requiring 35 min... 🤔  can we not have this?

Makefile
```diff
@@ -140,6 +145,26 @@ start-x:
 		sleep 3; \
 	done; \
 	[ "$$ok" = 1 ] || { echo "Error: namespace setup failed after 5 attempts"; exit 1; }
+	@echo "Creating 2-of-2 namespace (retrying until the committer is ready)..."
+	@ok=0; for attempt in 1 2 3 4 5; do \
```
ale-linux: This appears a bit brittle - can we do something a bit more resilient?

ale-linux (changes requested):
Great stuff! Just a couple of small improvements

Makefile
Storm1289: Agreed. The retry was already there for the first namespace and I copied it, which made it two copies of the same shell that's the brittle part. Folded both into one `create-namespace` macro taking the namespace and policy, with a doubling backoff (~1 minute total) instead of five fixed 3s sleeps. One implementation to harden now instead of two.

The root cause is that `start-x` waits on `nc -z localhost 7001`, and the port opens before the committer can serve config changes - so some retry is still needed. Happy to swap it for a real readiness probe if you'd prefer.

.github/workflows/tests.yml
Storm1289: done

Storm1289:
Hi @ale-linux, please review this PR again whenever you have some time

arner (approved):
LGTM!

Makefile
ale-linux: It's ok for now - can you just please create an issue under https://github.com/hyperledger/fabric-x-evm/issues/10 to track the fact that we must use a cleaner way to determine when the committer is ready to handle requests?

ale-linux (approved):

Makefile
Storm1289: opened issue #327 and also please review pr #227

Issue #221, opened 2026-07-08
Synchronisation on startup


ale-linux:
On fabric-x, if updates are received from the notification service, recovery from crash must be performed via delivery. This means that a node must come up, attach to both delivery and notification and stop delivery when the two have caught up

Issue #69, opened 2026-03-26
Optimize Synchronization Architecture


ale-linux:
Reevaluate the Gateway and Endorser synchronization architecture to address performance issues where Fabric-X commits transactions faster than the current delivery client can process them. The current approach using Fabric's Deliver service may not provide sufficient throughput for high-performance Fabric-X deployments. Investigate alternative approaches including subscribing to Fabric's event service for block notifications, using pub/sub mechanisms for asynchronous block delivery, or directly querying the ledger for new blocks. Analyze the performance characteristics, reliability, and complexity of each approach. Implement and benchmark the most promising alternative, measuring synchronization throughput and latency. If the new approach provides significant improvements, migrate the codebase to use it and update documentation. This ensures the synchronization layer can keep pace with Fabric-X's high transaction throughput.

---

**Quarter:** Q3

rootp1:
Hey, @ale-linux. Can I get this assigned and work on this issue?

ale-linux:
There are several points that need to be considered:

- with the fabric backend, we can only use the delivery service
- with fabric-x we can use the far faster notification service
    - this needs to be restructured to ensure that there is no code duplication
    - we still need delivery for after crashes

This is a big one so I'll start on it @rootp1

PR #310, opened 2026-08-11
gateway: HybridSynchronizer for fabric-x startup

### LINKED ISSUE #220: Fabric vs. fabric-x update handler
There should be a single interface to obtain updates from the committer/peer - whether through delivery (for fabric) or through notiication or delivery (for fabric-x)

ale-linux:
On fabric-x, the gateway now starts both the delivery service and the notification service concurrently at startup. The delivery service streams full blocks from the last locally-persisted block, catching up with the network. The notification service tracks the latest committed block number in real time. The switch from delivery to notification is decided inside a single HandleBatch call in the notification gate: once the delivery position meets or exceeds the notification position, the gate atomically flips to live mode and forwards the batch, eliminating any gap between the two phases. After the switch, all delivery resources are cancelled and released for GC.

Closes #220 #221 #69

### DIFF
--- a/integration/test_helpers.go
+++ b/integration/test_helpers.go
@@ -50,8 +50,6 @@ import (
 	"github.com/hyperledger/fabric-x-sdk/local"
 	"github.com/hyperledger/fabric-x-sdk/network"
 	nfab "github.com/hyperledger/fabric-x-sdk/network/fabric"
-	nfabx "github.com/hyperledger/fabric-x-sdk/network/fabricx"
-	"github.com/hyperledger/fabric-x-sdk/notification"
 	"google.golang.org/protobuf/proto"
 )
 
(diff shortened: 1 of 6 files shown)


integration/test_helpers.go
```diff
@@ -507,12 +462,30 @@ func NewFabricXTestHarnessWithNotifications(t *testing.T, logger sdk.Logger, evm
...
 
+	// Wait for the hybrid synchronizer to finish delivery catch-up.
+	if err := app.WaitUntilSynced(t.Context(), sync, 60*time.Second); err != nil {
+		return nil, fmt.Errorf("timed out waiting for sync: %w", err)
+	}
+
+	// Now that catch-up is complete, reconfigure the live handler chain:
+	// remove the chain store (block persistence is not needed for live perf
+	// testing and would just add latency) and add the extra handler.
+	if hs, ok := sync.(interface {
```
arner: AFAICT AddHandler and RemoveHandler are only here to satisfy this test. Proposal: don't wire in Chain at all for the perf tests. We can create an in memory height tracker and add that as the last handler and as the BlockHeightReader of the notification service.

Then we can remove the AddHandler and RemoveHandler, and the code will be safer and easier to reason about.

```
type HeightTracker struct {
  last atomic.Uint64
}

func (h *HeightTracker) Handle(ctx context.Context, b blocks.Block) error {
  if b.Number >= 0 { 
    h.last.Store(uint64(b.Number))
  }
  return nil
}

func (h *HeightTracker) BlockHeight(ctx context.Context) (uint64, error) {
  return h.last.Load() + 1, nil
}
```

arner (commented):
Looks good, one comment

integration/test_helpers.go
ale-linux: Really good point - done. Much cleaner now.

arner (approved):
Nice!

PR #311, opened 2026-08-11
fix: complete leaked TxQueue entries on submission failure

arner:
A submission failure after endorsement (e.g. ExecFailure, or a real orderer/network error) was silently dropped in BatchSubmitter — no TxQueue.Complete() call, so it leaked forever, permanently inflating in-flight transaction counts which caused issues with txFence too.

Thread the tx hash through BatchSubmitter and call Complete() on failure so it's cleaned up instead of leaking.


ale-linux (approved):
Nice catch - very subtle issue which would have been hard to debug!

Issue #148, opened 2026-05-03
Non atomic block ingest can permanently fork gateway state from persisted chain


rootp1:
it is happening because state trie is committed and prevHash is advanced before SQL persistence, so a later DB failure leaves parent state advanced while block rows are missing (chain.go (line 73), chain.go (line 79), chain.go (line 81)).

We can repoduce it by calling Chain.Handle with a block that causes InsertBlock/tx insert failure 
(for example malformed lengths violating DB checks); 
Handle returns error, but root and prevHash are already changed.

arner:
We continue in https://github.com/hyperledger/fabric-x-evm/issues/304.

Issue #241, opened 2026-07-20
Make eth_estimateGas return a real value


arner:
The real gas is already computed on the endorser (endorser/execution/executor.go's ApplyMessage, ~line 433: result.UsedGas is already there, just discarded).

Getting it back to EstimateGas on the gateway means threading it through the endorser's query protocol (changing the ProcessCall signature which returns a standard Fabric *peer.ProposalResponse with no gas field today) and changing the CallContract signature on the Backend interface.

This will improve our compatibility with tooling and tests, even though we don't charge for gas.

PR #313, opened 2026-08-11
fix: advance chain tip only after SQL commit

### LINKED ISSUE #304: Chain.Handle advances tip before SQL commit (non atomic block ingest)
### Summary

`Chain.Handle` advances gateway chain state before the block is durable in SQL. If `InsertBlock` fails after that, the process keeps a parent hash (and optionally a trie root) that the database never stored. Later blocks and restarts can see a broken parent chain or a lasting split between the MPT and the block store.

This is the same class of problem called out in #148. I am writing it up again with a bit more detail after re reading the current code, including the case where the trie is off (which is what production uses today via `NewChain(..., false)`).

### Where

`gateway/core/chain.go`, `Chain.Handle`

Rough order today:

1. Convert the Fabric block to the domain model
2. If the trie is enabled, `ts.Commit` runs (can flush to disk)
3. `prevHash` is always updated in memory
4. `Store.InsertBlock` writes the block and txs to SQLite

There is no rollback of steps 2 or 3 if step 4 returns an error.

### Steps to Reproduce

1. Build a `Chain` with the trie enabled (as in `chain_resilience_test.go`)
2. Call `Handle` with a block that will fail inside `InsertBlock` (for example data that violates a DB constraint or length check)
3. Observe that `Handle` returns an error
4. Check that `prevHash` (and the trie root when the trie is on) already moved
5. Feed a later block or restart and inspect parent hash / state root vs SQL

With the trie disabled the same `prevHash` advance still happens before SQL, so parent links on later blocks can still be wrong even without an MPT.

### Expected Result

Either the block is fully committed (SQL, parent tip, and trie tip if used) or none of those side effects stick. A failed `InsertBlock` should leave the chain tip and state root as they were before `Handle` started.

### Actual Result

On SQL failure after the early updates:

- `prevHash` already points at the failed block
- with the trie on, the MPT may already be committed and flushed
- SQL has no matching block row
- the next block gets a parent that is not in the store
- restart seeds from SQL latest, so the durable trie tip and the SQL tip can disagree

### Why it matters

As soon as someone turns the trie on for real (for example for proofs or any root based tooling), a single failed insert can leave the gateway with a permanent split between state and block history. Even with the trie off, advancing `prevHash` before SQL is enough to break the parent chain the RPC layer reports.

### Possible fix direction

Keep provisional work off the published tip until SQL succeeds. For example:

- compute parent and domain block without mutating `prevHash`
- commit the trie only after SQL succeeds, or hold a provisional root and discard it on SQL failure
- set `prevHash` only after `InsertBlock` returns nil

A regression test that forces `InsertBlock` to fail after a successful trie commit (or after a `prevHash` update) and asserts the tip and root are unchanged would lock this in.

@arner would be good if you could take a look when you have a moment. Happy to follow up with a fix PR if this direction sounds right.

Siddhartha-singh01:
Fixes #304

Chain.Handle used to set prevHash before InsertBlock finished. If SQL failed after that, the in memory tip already pointed at a block the store never held, so later blocks could pick a parent that was not on disk. Same class of issue as #148, with more detail on the withTrie false path production uses today.

Agreed direction from the issue: advance prevHash only after InsertBlock returns nil. Trie commit and flush stay before SQL so a crash cannot leave SQL ahead of the MPT. Replay on resubscribe is still fine because the trie path is idempotent. Retryable vs hard fail handling is left as a follow up.

## Change

gateway/core/chain.go moves the prevHash update to after a successful InsertBlock. ParentHash for the new block still comes from the current tip before the write.

## Tests

* go test ./gateway/core/
* go test ./gateway/... short
* go vet ./gateway/core/

New coverage in chain_resilience_test.go for the happy path, insert failure without the trie, and insert failure with the trie on.

### DIFF
--- a/gateway/core/chain.go
+++ b/gateway/core/chain.go
@@ -69,8 +69,14 @@ func NewChain(dbConnStr, triePath string, withTrie bool) (*Chain, error) {
 	return &Chain{Store: blockStore, db: db, ts: ts, prevHash: prevHash}, nil
 }
 
-// Handle implements blocks.BlockHandler. It commits the block's write sets to the trie,
-// then persists the block and its transactions to the database.
+// Handle implements blocks.BlockHandler. Order matters for tip safety (#304):
+//  1. Build the domain block and stamp ParentHash from the current tip.
+//  2. Commit write sets to the trie when enabled (flush stays here so a crash
+//     after SQL but before a deferred flush cannot leave SQL ahead of the MPT;
+//     trie apply is idempotent on resubscribe replay).
+//  3. Persist the block via InsertBlock.
+//  4. Advance prevHash only after SQL returns nil, so a failed insert never
+//     publishes a tip that the store does not hold.
 func (c *Chain) Handle(ctx context.Context, b blocks.Block) error {
 	ebl := ConvertToDomain(b)
 
@@ -84,12 +90,12 @@ func (c *Chain) Handle(ctx context.Context, b blocks.Block) error {
 	} else {
 		ebl.StateRoot = types.EmptyRootHash[:]
 	}
-	c.prevHash = common.BytesToHash(ebl.BlockHash)
 
 	if err := c.Store.InsertBlock(ctx, ebl); err != nil {
 		return err
 	}
 
+	c.prevHash = common.BytesToHash(ebl.BlockHash)
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


Siddhartha-singh01:
@arner hey, Quick Tests DCO and Codecov are green on this one. OZ Hardhat keeps bouncing though, same kind of baseline noise we saw on earlier PRs, nothing pointing at the prevHash change itself.

If you have a preferred way to re kick that job or if you want me to leave it for a later rerun that is fine with me. Happy to tweak anything else you spot.

arner (approved):
Thanks. Don't worry about the OZ tests for now. I'm hopeful that they'll be fixed after  https://github.com/hyperledger/fabric-x-evm/pull/311 and https://github.com/hyperledger/fabric-x-evm/pull/313 are merged. As far as I'm aware the code in this PR doesn't touch their path anyway.

Siddhartha-singh01:
@arner Sir is there is anything else on which i can work on .

arner:
Hi, maybe https://github.com/hyperledger/fabric-x-evm/issues/241?

PR #299, opened 2026-08-05
gateway: enable the gRPC endorsement path via config 

Storm1289:
## Summary

Closes #247. Design: #219 (part 4).

Enables the gRPC endorsement path through configuration the final PR in the #22/#247 series. Everything introduced dormant in #242–#245 becomes reachable here, and only when configured. Without the new configuration, gateway behavior remains exactly as it is today.

### Dial the endorser over gRPC

- Added `endorser/client/dial.go` with:
  - `Dial(cfg common.ClientConfig) (*Client, error)`
- Builds mTLS-enabled gRPC transport credentials from `cfg.TLS`.
- Dials `cfg.Endpoint` and returns a `Client` implementing `api.Service`.
- Uses the same credential-building path already exercised end-to-end in #245.

### Config-driven split deployment

- Added `Gateway.Endorsers []common.ClientConfig` to `gateway/config/config.go`.
- When configured, the gateway dials all endorsers over gRPC (localhost for the co-located endorser, remote endpoints for other organizations) instead of constructing them in-process from the top-level `Endorsers` configuration.
- `Validate()` now branches by deployment mode:
  - Split mode validates gRPC dial targets and no longer requires the embedded endorser configuration.
  - Combined mode remains unchanged.

### Gateway wiring

- `gateway/app/app.go` introduces `newSplitApp`, selected automatically when `Gateway.Endorsers` is configured.
- Dialed gRPC clients are tracked on `App` and closed during `Shutdown`.
- Test RPC is explicitly disabled in split mode since it requires a local KVS for snapshot/revert support.

## Interface / API changes

- New exported `endorser/client.Dial`.
- New configuration field:
  - `config.Gateway.Endorsers []common.ClientConfig`
- `App` now owns an internal `endorserConns []*eclient.Client`, closed during shutdown.

## Tests

Added:

- `endorser/client/dial_test.go`
  - No-TLS success
  - Missing certificate file
  - Invalid certificate contents
- `gateway/config/validate_test.go`
  - Split mode accepts an empty top-level `Endorsers`
  - Invalid `Gateway.Endorsers` entries are rejected
- `gateway/app/app_test.go`
  - Split-mode routing
  - Dial failure
  - Cleanup of previously opened connections after a later dial failure
  - Successful split-mode application construction and shutdown
  - `buildApp` failure after successful dialing
  - Shutdown tolerating connection close errors

## Decisions from review
[...]

PR #317, opened 2026-08-13
fix: use gateway-supplied timestamp as EVM block.timestamp

### LINKED ISSUE #277: Gateway-supplied, bounded TIMESTAMP for EVM execution
## Problem

`block.timestamp` as seen by contracts (`vm.BlockContext.Time`) is currently a hardcoded constant, `defaultBlockTime = 1_000_000` (`endorser/execution/executor.go`), for every single transaction execution. It never varies with wall time, block number, or anything else. (`BLOCKNUMBER` has the same problem — `blockCtx.BlockNumber` is also always `0` on this path, see Non-goals.)

The issue surfaces in a number of OpenZeppelin tests. It's the reason `VestingWallet`, `TimelockController`, `AccessManager`, and any other time-gated contract logic can't behave correctly today — already documented as a known limitation (`docs/COMPATIBILITY.md`, `TIMESTAMP always 1_000_000`) and tracked as its own bucket in the baseline: 68 entries carry `"cause": "fixed-timestamp"` in `testdata/oz_known_failures.json`.

## Why this isn't just "call time.Now()"

Endorsement here is pre-order execution: the read/write set is computed once, by several independent endorsers, *before* ordering — and those RWS results must match exactly for endorsement policy / MVCC validation to succeed at commit. Any value each endorser derives independently (its own local clock, however NTP-synced) is a source of non-deterministic RWS divergence, not just "occasional skew." It's the same failure class regardless of how the value is rounded, because rounding only reduces disagreement between independently-computed values, it can't eliminate it.

Also explicitly ruled out: clamping each endorser's own timestamp to `max(candidate, lastSeenByThisEndorser)` to force monotonicity. This breaks under concurrency.

## Proposed design

A single deterministic input, supplied once by the party constructing the transaction request, sanity-checked (not re-derived) by the endorser. This is the same shape Fabric already uses for `ChannelHeader.Timestamp` — "the local time when the message was created by the sender," set once by the proposal's originator.

1. **Gateway sets a timestamp when it builds the request to the endorsers** (`EndorsementClient.ExecuteTransaction`) — `time.Now()` at request-construction time, passed alongside `inv`/`tx`, not folded into `endorsement.Invocation` itself. Keeps this entirely within fabric-x-evm's own types (`endorser/api.Service`, `EVMEngineInterface`, `endorsementpb`) — no `fabric-x-sdk` change needed.
2. **Endorser validates it against its own clock with an asymmetric bounded window**, then uses the value as-is — no averaging, no clamping against other transactions, no cross-transaction state. Pure function of that one request.
3. **That validated value becomes `blockCtx.Time`**, replacing `defaultBlockTime` in `NewExecutor` (`endorser/execution/executor.go`).

Deterministic by construction: every endorser evaluating the same request reads the same field out of the same request, not something it computes itself. No new moving parts, no dependency on another team's infrastructure (raised and rejected a KVS-backed shared-clock alternative for the same reason — it would need buy-in and operational ownership we can't get, and it still costs live contention on every transaction that reads `TIMESTAMP`, for no better a guarantee than this).

**No monotonicity enforcement.** Given the clamp is unsafe under concurrency, accept that `block.timestamp` can occasionally move backward by a small amount between two closely-spaced, differently-ordered transactions (bounded by real clock skew + queueing jitter — realistically tens to low hundreds of ms, not seconds). Worth documenting plainly: this will spuriously revert the rare contract that treats `block.timestamp` as a strict monotonic counter (`require(block.timestamp > lastAction)`) rather than a coarse clock. That's a contained, deterministic, per-transaction revert — a categorically safer failure mode than a consensus-breaking RWS mismatch, which is the alternative.

## Trust model

The gateway↔endorser channel is mTLS-authenticated, so this isn't a defense against an anonymous attacker — the bound exists to catch a *misbehaving-but-authenticated* gateway (bug, clock drift, compromised-but-still-trusted process) before it can materially skew contract-visible time. Content-level signature verification (so an endorser could trust a request's fields independent of transport auth) is tracked separately as future work and isn't a prerequisite for this — mTLS is the trust boundary this design relies on today.

## Bounds should be asymmetric, and configurable

The two directions aren't equally risky:
- **Too far in the past** — harmless given no clamping: it just makes that one transaction's `block.timestamp` stale, doesn't let anyone jump ahead. Replay is already covered independently of this bound, on two layers: the Ethereum tx's own sender nonce (checked against ledger state in `PrepareMessage`, `executor.go`), and Fabric's TxID uniqueness, itself derived from a random per-proposal nonce (`protoutil.ComputeTxID(nonce, creator)`, `endorsement/proposal.go`) rather than from the timestamp — so a replayed proposal is already rejected on TxID collision regardless of how old its claimed timestamp is. That means this bound carries no replay-prevention weight at all; it's purely hygiene (reject obviously-stale garbage before spending endorsement cycles on it). Can be generous — tens of seconds to low minutes — without weakening anything.
- **Too far in the future** — this is the actual exploitable direction: a gateway (buggy or compromised) claiming more elapsed time than real lets a caller early-unlock a timelock, close a Dutch auction favorably, skip a cooldown, or slip into an eligibility window that should've closed. Textbook [SWC-116](https://swcregistry.io/docs/SWC-116) "timestamp dependence" shape. Should be tight — on the order of single-digit-to-low-double-digit seconds, enough to cover realistic NTP drift plus gateway→endorser latency, nothing more. For further reference: post-merge Ethereum gives block proposers ~zero timestamp discretion; the much-criticized pre-merge PoW ~900s miner leeway is specifically what SWC-116 cites as too loose, and it's the ceiling we should be nowhere near.

Make both bounds config (`endorser/config/config.go`, alongside the existing `Endorser` struct), not hardcoded — different deployments (co-located low-latency consortium vs. cross-region multi-org) legitimately need different values, and there's no single constant that's right everywhere. But we should provide sane defaults: proposing **10s future / 60s past** as the starting point (out of the box, works for both testnode and a typical single-region deployment without tuning) — narrow enough on the future side to keep the SWC-116 exposure negligible, generous enough on the past side to absorb realistic retry/queueing delay given it's carrying no security weight. Open to adjustment once we have real latency numbers from an actual multi-node deployment rather than reasoning from first principles.

## Concrete changes

All of this stays inside fabric-x-evm — no `fabric-x-sdk` release/bump needed, `endorsement.Invocation` is untouched:

- `endorser/api.Service.Execute` (`endorser/api/service.go`): add a `timestamp time.Time` parameter alongside `inv`/`ethTx`.
- `gateway/core/endorse.go` (`EndorsementClient.ExecuteTransaction`): compute `time.Now()` once per call, pass it to each endorser alongside `inv`/`tx`.
- `endorser/core/endorser.go` (`Endorser.Execute`): accept the new parameter, validate against the configured bounds before calling `f.Engine.Execute`; reject (as an application-level endorsement failure, not a crash) outside the window.
- `EVMEngineInterface.Execute`/`EVMEngine.Execute` (`endorser/core/endorser.go`, `endorser/execution/executor.go`): accept the validated timestamp, thread to `NewExecutor`.
- `endorser/config/config.go`: new bounded-window config fields.
- `api/endorsementpb/endorsement.proto`: add a `timestamp` field to `ExecuteRequest` itself (sibling to `ethereum_tx`, not nested in `Invocation` — same reasoning as above, keep it separate from the invocation concept). Wire carrier for the gRPC path — worth checking whether that client is wired up yet at all; grepping today, `endorsementpb` is only referenced server-side (`endorser/server/`), nothing in `gateway/` calls it, so this transport may still be dormant. Include the field regardless since it's cheap and keeps the proto in sync with the in-process path.

## Non-goals (separate issues)

- **`BLOCKNUMBER`** will be tackled in another issue.
- **Ledger/API-level block timestamp** (`eth_getBlockByNumber`'s `.timestamp`, currently `time.Now()` in `fabric-x-sdk/blocks/{fabric,fabricx}/parser.go`) is a separate, `fabric-x-sdk`-owned concern from the EVM-visible one this issue fixes. Could plausibly reuse the same per-tx value once it exists, but that's its own follow-up.
- Real monotonicity guarantees — deliberately not attempting, see above.
- **Hardhat's `evm_setNextBlockTimestamp`/`evm_increaseTime`** (stubs in `gateway/testimpl/hardhat_helpers.go`) stay stubs. This issue is a prerequisite for that.

## Validation

- Unit test: two simulated endorsers given the identical request produce identical `blockCtx.Time` (and therefore identical RWS) regardless of when each happens to process it.
- Unit test: request timestamp outside the configured future/past bounds is rejected; inside bounds is accepted unmodified (no clamping).
- Concurrency test: many concurrent requests, timestamps not required to be monotonic across them, no shared state contended.
- Re-run baseline against `testdata/oz_known_failures.json` — the 68 `fixed-timestamp` entries are the primary target; expect most, not necessarily all, to flip to passing (some may land on a different known cause, e.g. non-monotonicity edge cases accepted above).

Siddhartha-singh01:
Fixes #277

block.timestamp inside EVM execution was stuck at 1000000 for every transaction. Time gated contracts (timelock, vesting, AccessManager, and friends) could not behave correctly, and a large OZ fixed timestamp bucket followed from that.

## Design

Gateway stamps time.Now once per ExecuteTransaction and sends the same value to every endorser. Each endorser checks that the request time is within a configurable window of its local clock (defaults: 10s future, 60s past), then uses that Unix second as blockCtx.Time with no clamping. Multi org RWsets stay deterministic because they share one input.

This is per request time for endorsement, not a single per block clock. Endorsement is pre order and is not re executed at commit, so EVM visible time has to be fixed at endorse. JSON RPC eth_getBlockByNumber timestamps remain a separate ledger or API path (fabric x sdk parse time today) and may not match the endorsement time contracts saw. That gap is documented in COMPATIBILITY.md and can be a follow up.

## Change

* gateway/core/endorse.go: one reqTime for all endorsers
* endorser/core: validate bounds, reject pre epoch and unset times
* endorser/execution: Execute and NewExecutor take blockTime
* endorser/config: max timestamp future and past with defaults
* endorsementpb ExecuteRequest.timestamp for the gRPC path
* COMPATIBILITY.md updated for real TIMESTAMP on execute

## Test plan

* go test ./endorser/core/ ./endorser/execution/ ./gateway/core/ ./endorser/client/ ./endorser/server/
* go test ./... short
* go vet and staticcheck on touched packages
* CI Quick Tests
* OZ Hardhat: expect many fixed timestamp entries to move; refresh oz_known_failures.json if the baseline gate asks for it


Siddhartha-singh01:
@arner Sir can you review this once

docs/COMPATIBILITY.md
```diff
@@ -254,19 +254,27 @@ network values.
...
+`block.number` inside an executed transaction always reads `0` until a separate fix lands.
 
 **`eth_call` with a block number**: the state DB is correctly snapshotted at the requested height,
-and the EVM `NUMBER` opcode is set to that block-number argument (`0` for `latest`). `TIMESTAMP`,
-however, is always `1_000_000` regardless of the requested block. Contracts that read
-`block.timestamp` inside a view function therefore see a fixed, non-historical value (and
-`block.number` reads `0` for the common `latest` call).
+and the EVM `NUMBER` opcode is set to that block-number argument (`0` for `latest`). `TIMESTAMP` on
+the call path is the endorser's wall-clock Unix second (not historical and not the legacy
+`1_000_000` stub), so view functions that check role grant schedules / delays see a time close to
```
arner: no need to mention the historical stub

api/endorsementpb/endorsement.proto
```diff
@@ -36,6 +36,10 @@ message ExecuteRequest {
 
     // Invocation built by the sender, which the endorser passes to the builder.
     Invocation invocation = 3;
+
+    // Unix seconds for EVM block.timestamp. Set once by the gateway at request
+    // construction so all endorsers share the same value (issue #277).
```
arner: Let's not mention issue numbers (also in other places).

Storm1289:
Hi @Siddhartha-singh01, Design looks sound. Three things I think are worth addressing:

* **Constructor safety:** Consider moving timestamp bounds into `New(engine, builder, cfg)` so the security defaults are applied at construction rather than relying on a separate `SetTimestampBounds` call.
* **Partial-endorser rejection:** Now that #299 is runtime-enabled, clock drift between remote endorsers is a real deployment concern. It would be useful to document the impact on N-of-N policies and how `max-timestamp-future` / `max-timestamp-past` should be tuned.
* **Single source of truth:** Timestamp skew defaults are currently handled in `config`, `SetTimestampBounds`, and `skewBounds`. Consider routing these through the config methods to avoid having multiple fallback paths for the same security-sensitive setting.

The OZ Hardhat CI failure looks like baseline drift and is probably better handled separately with @arner in the context of PR #286.

what do you think?

endorser/core/endorser.go
```diff
@@ -26,37 +28,72 @@ import (
...
 	NonceAt(ctx context.Context, account ethcommon.Address, blockNumber *big.Int) (uint64, error)
 }
 
-// New returns a new Endorser.
+// New returns a new Endorser with default timestamp skew bounds (#277).
 //
 // Arguments:
 //   - `engine`:  Manages EVM execution and state reads.
 //   - `builder`: Creates the signed ProposalResponse.
 func New(engine *execution.EVMEngine, builder endorsement.Builder) (*Endorser, error) {
```
Storm1289: Consider New(engine, builder, cfg config.Endorser) so timestamp bounds are set at construction and SetTimestampBounds can go away. If applyTimestampBounds isn't called (or a test builds &Endorser{} directly), maxFuture/maxPast are zero and only skewBounds()'s guard rescues that security policy shouldn't live behind a helper that a future refactor could quietly remove.

endorser/core/endorser.go
```diff
@@ -70,6 +107,45 @@ func (f *Endorser) Execute(ctx context.Context, inv endorsement.Invocation, ethT
...
+func (f *Endorser) clock() time.Time {
+	if f.now != nil {
+		return f.now()
+	}
+	return time.Now()
+}
+
+// skewBounds returns future/past windows, applying the same defaults as
+// SetTimestampBounds when fields are zero (e.g. Endorser literals in tests).
+func (f *Endorser) skewBounds() (future, past time.Duration) {
```
Storm1289: Same zero→default fallback also exists in SetTimestampBounds (L69) and in config.TimestampFutureSkew/PastSkew. Route through the config methods so there's one authority three copies for a security-relevant setting is a drift bug waiting to happen.

endorser/core/endorser.go
```diff
@@ -26,37 +28,72 @@ import (
...
+	f.maxFuture = future
+	f.maxPast = past
+}
+
 // Execute endorses an Ethereum transaction and returns a signed proposal response.
 // Reverts are endorsed and submitted (so the receipt records status=0); client-caused failures
 // (invalid tx or failed execution) surface as a non-2xx status that CreateSignedTx won't submit.
-func (f *Endorser) Execute(ctx context.Context, inv endorsement.Invocation, ethTx *types.Transaction) (*peer.ProposalResponse, error) {
+//
+// timestamp is the gateway-supplied wall time used as EVM block.timestamp (#277).
```
Storm1289: please do not mention the issue no

Siddhartha-singh01:
@Storm1289 thanks again, and thanks for the inline notes too. That sharpened it for me.

On construction: fully with you. Putting cfg into New(engine, builder, cfg) so bounds are set at create time, and dropping SetTimestampBounds, is the right move. Relying on skewBounds as a late rescue when someone builds &Endorser{} or skips applyTimestampBounds is not a place for a security policy to live. I will make construction the only place that installs the bounds.

On single source of truth: agreed. The zero to default path should go through the config helpers only, not a copy in SetTimestampBounds and another in skewBounds. One authority so it cannot drift.

On issue numbers in code comments: got it, I will strip those and keep the comments about behavior only.

Docs for multi endorser clock skew and N of N still on the list, and OZ baseline with @arner separately as you suggested.

Really appreciate the precision here.

endorser/execution/executor.go
```diff
@@ -241,18 +256,22 @@ type Executor struct {
 }
 
 // NewExecutor creates an Executor with the provided StateDB and reader.
-// blockNumber sets the EVM block context (nil = 0). evmConfig.ChainConfig must be set.
+// blockNumber sets the EVM block number context (nil = 0).
+// blockTime is the Unix second for block.timestamp (0 → DefaultBlockTime).
+// evmConfig.ChainConfig must be set.
 // The caller is responsible for closing the reader when done with the Executor.
 // The stateDB parameter accepts ExtendedStateDB interface to allow DualStateDB for testing.
-func NewExecutor(stateDB ExtendedStateDB, reader ReadStore, blockNumber *big.Int, evmConfig EVMConfig) (*Executor, error) {
```
arner, line 247: I wonder if this default blocktime is still necessary?

In production I assume we will always have one. If the gateway for some reason didn't supply one, we could default to the endorser's time.Now(). In by far the most cases this will not lead to conflicts. But on the other hand, why would we allow that? We control the code of the gateways and I think it's reasonable to make it mandatory (actually I just noticed the GRPC handler returns an error if timestamp is 0).

arner (changes requested):
wow that's a big one. I need some more time to review on Monday.

testdata/oz_known_failures.json
```diff
@@ -1895,7 +1598,7 @@
   },
   {
     "id": "GovernorCountingFractional using $ERC20Votes \"before each\" hook for \"deployment check\"",
-    "note": "Predicted, not yet directly observed failing — this file's fixture uses the identical vulnerable pattern already confirmed racy in 8 other Governor-family entries (GovernorHelper.delegate()'s Promise.all of dependent writes, no confirmation wait, endorsement-time read snapshot goes stale before commit).",
+    "note": "Predicted, not yet directly observed failing \u2014 this file's fixture uses the identical vulnerable pattern already confirmed racy in 8 other Governor-family entries (GovernorHelper.delegate()'s Promise.all of dependent writes, no confirmation wait, endorsement-time read snapshot goes stale before commit).",
```
arner: Is it possible that you're running an older build of the baseline tool? If everything works as it should, the current version wouldn't do this html escaping.

arner:
About the hardhat tests. You may have been too optimistic to remove the tests that are now listed as "Regressions". The 39 fixed are real though, so congrats!

The real reason for many of these tests to fail is that `evm_increaseTime`, `evm_setNextBlockTimestamp`, and `evm_mine` are no-ops. We have an issue open for evm_mine (it depends on a small addition to fabrictest in the fabric-x-sdk). There is no issue for evm_increaseTime and evm_setNextBlockTimestamp yet. After the changes in this PR have solidified, we can create that issue as a followup and it should hopefully fix the others!

Siddhartha-singh01:
@arner thank you for the thorough review, this was really useful.

On the Hardhat side you were right: I was too aggressive removing known failures. The 39 fixed cases are real wins and stay out of the baseline. The ones that still fail because time helpers are no ops are back in known failures with cause hardhat time rpc, and COMPATIBILITY now notes that evm_increaseTime, evm_setNextBlockTimestamp, and evm_mine still need a follow up once this PR settles, as you suggested.

On DefaultBlockTime: agreed it should not paper over a missing stamp. Execute now requires a non zero blockTime, NewExecutor rejects zero, and the gateway path already treats an unset proto timestamp as an application error. Call still uses wall clock for views.

On docs and comments: removed the historical stub wording, dropped issue numbers from code and proto comments, and added a short multi endorser clock skew section so operators know how max timestamp past and future interact with N of N.

On the baseline file escaping: cleaned the known failures list with the current tool shape (plain JSON, no HTML entities). If anything still looks off on the next CI run I will re check against main.

Pushed as cafe325. Happy to adjust further if anything still feels off.

Storm1289 (commented):
LGTM

arner (approved):
Good stuff @Siddhartha-singh01 , thanks!

Issue #247, opened 2026-07-20
[EPIC] Endorser API


arner:
Add a GRPC API for endorsement.

As discussed in https://github.com/hyperledger/fabric-x-evm/pull/219.

Issue #312, opened 2026-08-11
Commit EVM execution failures (out-of-gas, invalid opcode) instead of silently dropping them


arner:
## Summary

A transaction whose EVM execution fails with `ExecFailure` (out-of-gas, `INVALID` opcode, ...) is
endorsed as if it will be committed, but is then silently discarded before ever reaching an orderer.
The client that submitted it never gets a receipt — `eth_getTransactionReceipt` returns `null`
forever, so `tx.wait()`-style polling hangs indefinitely. On real Ethereum, the equivalent
transaction is mined with a failed (`status: 0`) receipt, just like a `revert()`.

This is not a corner case: any contract path that can run out of gas or hit `assert()`/`INVALID` hits
it, which real dApps and every serious test suite exercise routinely.

Depends on [#311](https://github.com/hyperledger/fabric-x-evm/pull/311) — that fix already stops a
failed submission (this case included) from leaking `TxQueue` state forever, so the operational
symptom (a stuck queue) is handled separately. What's left, and what this issue is about, is that the
transaction still never gets a real receipt.

## How this was found

Traced from a gateway log line during an OZ Hardhat suite run:

```
ERRO [gateway.core.batch_submitter] worker -> Worker 1: submit failed: package proposal: proposal
response was not successful, error code 460, msg invalid opcode: INVALID
```

A transaction had been endorsed as if it would commit, then silently rejected at packaging.

## Current behavior (traced end to end)

1. `classify()` in [`endorser/core/endorser.go`](../endorser/core/endorser.go) maps an
   `*execution.ExecFailure` engine error to `StatusExecFailure = 460`
   ([`common/proposal.go:21`](../common/proposal.go#L21), comment: *"valid tx whose EVM execution
   failed ...; should be mined (not yet)"*).
2. [`gateway/core/endorse.go`](../gateway/core/endorse.go)'s `ExecuteTransaction` treats
   `StatusExecFailure` as an endorsable/successful outcome (same branch as `StatusOK` and
   `StatusEVMRevert`), so no error is raised at endorsement time — the transaction is queued for
   submission.
3. `fabric-x-sdk`'s packager (`network/fabricx/packager.go:63`, `CreateTx`, adapted from
   `protoutil.CreateSignedTx`) rejects any response status outside `[200, 400)`. `460` fails this
   check, so `PackageTx` returns an error *before* the transaction is ever broadcast to an orderer.
[...]

Issue #257, opened 2026-07-20
Report accurate gasUsed in transaction receipts (currently hardcoded to 0)


arner:
Part of #7.

`gasUsed`, `cumulativeGasUsed`, and `effectiveGasPrice` are all hardcoded to `0` in `eth_getTransactionReceipt` (`gateway/api/models.go:69-89`, the `receipt()` function). Block-level `gasUsed` is also hardcoded to `0` (`gateway/api/models.go:254`) — likely derivable as a sum of per-tx values once this is fixed, so treating that as a small follow-on rather than separate scope.

This is a different, harder plumbing problem than #241 (`eth_estimateGas`): #241 only needs the gas value for the duration of one synchronous RPC call, but a receipt is served long after execution, from committed block data. Today `domain.Transaction` (`gateway/domain/models.go`) has no `GasUsed` field at all, and the extractor that builds it purely from committed Fabric block data (`convertTransaction`, `gateway/core/chain.go:167-233`) has no gas information available to it — just the raw tx bytes, Fabric validation code, and events.

The root value is the same as #241's: `ApplyMessage`'s `result.UsedGas` in `endorser/execution/executor.go`, currently discarded. Getting it into the receipt means it has to survive from endorsement through ordering and commit, not just return synchronously from a call. Gas usage is deterministic (a pure function of code + state + gas limit), so independent endorsers should compute the same value — the same property that already lets per-tx logs flow through commit today via `common.MarshalLogs`/`UnmarshalLogs` (`common/events.go`, wraps log data in a `peer.ChaincodeEvent` that rides along with the transaction and gets decoded back out by the extractor). That's a plausible model to follow for gas too, though the concrete plumbing (new field on `domain.Transaction`, whatever persists it in `gateway/storage`, decoding it back in `convertTransaction`) needs to be worked out as part of this issue rather than assumed up front.

**Validation:** a real ERC20 `transfer` receipt's `gasUsed` should land in a sane range (tens of thousands), not `0`; cross-check against `eth_estimateGas` once #241 lands (should be close, not identical, since one is exact and one is an estimate).

PR #318, opened 2026-08-14
endorser: own the fabric-x endorsement builder and drop the proposal

### LINKED ISSUE #265: endorser: own the EndorsementBuilder and drop the proposal on the Fabric-X path
Part of #22 / #247. Follow-up - to be done **after** the enable PR (#246). Context: #219 part 1 ("What Execute's Endorsement Covers") and the discussion on #256 (endorsement.proto:33) with @arner.

## Background

`Execute` returns a `*peer.ProposalResponse` and the transaction is assembled by the SDK's packager, which requires the proposal. For Fabric-X the proposal is not needed for a valid transaction, so today we carry and store data we don't need. Dropping it requires either a change in `fabric-x-sdk` or a workaround on our side.

## Decision (with @arner)

Maintain our **own EndorsementBuilder in fabric-x-evm** rather than change the shared SDK:
- the SDK's builder is generic, and EVM-specific logic (ethereum logs/events) has already leaked into it;
- owning it keeps the EVM-specific bits on our side and lets us iterate freely (e.g. future metadata like `gasUsed`, possibly our own metadata protobuf);
- keep it generic where natural, and contribute back to the SDK if it makes sense.

## Scope

- Add our own EndorsementBuilder in fabric-x-evm.
- Drop the proposal on the **Fabric-X** path so `Execute` no longer carries or stores it; classic Fabric keeps sending only the proposal hash.
- Finalize the `ExecuteRequest` / `ExecuteResponse` wire shape, replacing the provisional `Execute` handling left in the server (#243) and client (#244).

## Sequencing

**Not a blocker** - carrying the proposal is fine for now, so this waits until the gRPC endorsement path is live (after #246) and can be validated end to end.

## Done when

- Fabric-X transactions are built without the proposal, via our own builder.
- Classic Fabric still works (proposal hash only).
- Existing endorser and gateway suites stay green.

Storm1289:
## Summary

Closes #265. Part of #22 / #247.

On the Fabric-X path, `Execute`'s result is assembled into a transaction by the SDK's packager, which requires the full `peer.Proposal`. Fabric-X doesn't need the proposal for a valid transaction, so we were carrying and storing bytes we never use. Per the decision with @arner (#265), fabric-x-evm now owns its endorsement builder and drops the proposal on the Fabric-X path. Classic Fabric is unchanged.

## What changed

**`endorsement/fabricx/` (new)**

- `builder.go` - our own endorsement builder, a byte-identical port of the SDK's, so the endorsement bytes stay exactly the same.
- `invocation.go` - a Fabric-X invocation that builds only the header a Fabric-X envelope actually keeps, instead of a whole proposal.
- `builder_test.go` / `invocation_test.go` - unit tests, including a byte-for-byte equivalence check against the SDK's output.

**Gateway / endorser wiring**

- `gateway/core/endorse.go` - `EndorsementClient` gains a `protocol`; `createInvocation` selects by it: classic Fabric still signs over the whole proposal and keeps the SDK's builder, while Fabric-X uses the header-only builder. An empty protocol means Fabric-X, as everywhere else.
- `endorser/api/service.go`, `endorser/app/factory.go` (+ `factory_test.go`), `gateway/app/wiring.go` - thread the protocol/builder selection through.

**Tests**

- `integration/test_helpers.go` - a Fabric-X proposal now carries only a header, so the eth tx can no longer be extracted from it; `submit` and `getEndorsedTxForSmartContractCall` pass the tx explicitly, and the now-dead `extractEthTxFromProposal` helper is removed.

## Behavior

- **Classic Fabric:** unchanged - full proposal, SDK builder.
- **Fabric-X:** the proposal is dropped; endorsement bytes are byte-identical to before via the ported builder, so nothing downstream needs to change.

## Testing

Unit tests cover the new builder/invocation, including byte-equivalence with the SDK, and the protocol selection in `endorse.go`.

This changes what the Fabric-X path signs and stores, so it is worth one end-to-end run. **Suggest running this with the `want-integration-tests` label.**

#306's 2-of-2 endorsement test over a real gRPC backend (`integration/endorsement_grpc_test.go`) exercises exactly this path and isn't modified here, so it's the strongest confirmation that a real Fabric-X backend still accepts the endorsement once the proposal is dropped.


Storm1289:
Hi @arner, please review this pr, whenever you have sometime.

arner:
Will review soon.

endorsement/fabricx/builder.go
```diff
@@ -0,0 +1,171 @@
...
+			Message: res.Message,
+			Payload: res.Payload,
+		},
+	}, nil
+}
+
+// marshalMetadata returns the signed-over metadata: input args, then the event.
+// Each is left out when empty.
+func marshalMetadata(inv endorsement.Invocation, res endorsement.ExecutionResult) ([][]byte, error) {
+	var metadata [][]byte
```
arner, line 80: There are some opportunities here! Originally I kept it as close to classic Fabric as possible, but that was when we didn't have the metadata field yet. As a followup we can think of a nice way to populate the metadata slice. We don't need the ChaincodeInput and ChaincodeEvent formats. And we have some additional fields that we may want to surface (revert/failed, gasUsed, timestamp...).

Anyway, not for this PR, but let's keep it in mind.

endorsement/fabricx/invocation.go
```diff
@@ -0,0 +1,72 @@
...
+	ccHdrExt, err := protoutil.Marshal(&peer.ChaincodeHeaderExtension{ChaincodeId: ccid})
+	if err != nil {
+		return endorsement.Invocation{}, fmt.Errorf("marshal chaincode header extension: %w", err)
+	}
+	chdr, err := protoutil.Marshal(&commonpb.ChannelHeader{
+		Type:      int32(commonpb.HeaderType_ENDORSER_TRANSACTION),
+		TxId:      txID,
+		Timestamp: timestamppb.Now(),
+		ChannelId: channel,
+		Extension: ccHdrExt,
```
arner: There's a good chance we don't need this. Can you try?

endorsement/fabricx/invocation_test.go
```diff
@@ -0,0 +1,196 @@
...
+		Header: protoutil.MarshalOrPanic(&commonpb.Header{
+			ChannelHeader:   protoutil.MarshalOrPanic(chdr),
+			SignatureHeader: protoutil.MarshalOrPanic(shdr),
+		}),
+		Payload: payload,
+	}
+}
+
+// identitySigner serializes a real msp identity, which the packager unmarshals.
+type identitySigner struct{}
```
arner, line 192: we can get this from the sdk now: `fabrictest.MockSigner{}`

endorsement/fabricx/builder.go
```diff
@@ -0,0 +1,171 @@
...
+	// Whatever is left was read but not written.
+	for _, r := range readByKey {
+		read := &applicationpb.Read{Key: []byte(r.Key)}
+		if r.Version != nil {
+			read.Version = &r.Version.BlockNum
+		}
+		readsOnly = append(readsOnly, read)
+	}
+
+	sort.Slice(readsOnly, func(i, j int) bool {
```
arner: I know this is a copy/paste, but apparently nowadays the faster option is slices.SortFunc. It's even in the comment of the function in the stdlib :).

https://pkg.go.dev/sort?utm_source=chatgpt.com#Sort

arner (changes requested):
Looks good! Some small nitpicks / minor improvements. Having it here gives us more flexibility to easily iterate; later we can see what we feed back to the SDK.

endorsement/fabricx/invocation_test.go
Storm1289: Happy to use it, but it is only on SDK main, not our pinned version, so it needs a `go.mod` bump of about six weeks of SDK changes. I would rather do that as its own commit than bury it in this PR, so it is easy to bisect if something
breaks. Say the word and I will do it separately.

endorsement/fabricx/invocation.go
Storm1289: You were right, nothing reads it. `CreateTx` unmarshals the channel header, changes the type to MESSAGE and re-marshals it, and never touches the extension. Removed.

It changes the bytes that go on chain, so I checked against a real network rather than unit tests: `TestFabricX` 13/13 on a fresh one. Classic Fabric is unaffected, it still uses the SDK's `NewInvocation`.

endorsement/fabricx/builder.go
Storm1289: Done, switched all three sorts to `slices.SortFunc`.

endorsement/fabricx/builder.go
Storm1289: Agreed. Opened #363 so it does not get lost, linked to #257 and #312 since the gasUsed and revert fields land in the same place

Storm1289:
One thing I noticed while working through these, separate from the review points. `make checks` runs gofmt, go vet and staticcheck, and none of them flag old code that newer Go has a simpler replacement for. I ran `modernize` from `golang.org/x/tools`, and it found 16 places in our own code (the rest are in generated files). None of them are bugs, they are all "there is a cleaner way to write this now". It would not have caught the `sort.Slice` one though, so it is no replacement for review.

I am not suggesting we do it in this PR, and I do not think it is quick: some of it is in shared code, so it would need proper testing rather than a blind auto-fix. But the repo has grown a lot and it seemed worth raising. Want me to open an issue so we can pick it up later, or leave it?

arner:
Yes, let's integrate `go fix` in `make checks`. 

As for the rest of the PR, now that we have the option for a generic SDK implementation that satisfies all our fabric-x-evm needs (see https://github.com/hyperledger/fabric-x-evm/issues/363#issuecomment-5569450773, plus the removal of the unnecessary marshaling), I don't see the need anymore to include this code in our own repo. Then it's much cleaner to keep it at the source.

Sorry for the churn, but I propose that you copy your changes into the SDK and make a PR there directly, so that fabric-x-evm can keep depending on the package and we can centralize the improvements. Thanks!

PR #319, opened 2026-08-14
Add perf smoke test

ale-linux:
This PR adds a quick perf test run as part of CI to detect the current status of the perf in the system and spot a) breakages for the perf tests and b) massive drops.


.github/workflows/tests.yml
```diff
@@ -11,6 +11,7 @@ on:
 
 permissions:
   contents: read
+  pull-requests: write
```
arner: Granting writes on pull requests can be abused by a malicious PR that modifies the workflow file in the PR branch. Because GitHub runs the workflow definition from the PR head, it can elevate its own token to the GITHUB_TOKEN of the repo and perform repo writes (comments, labels, edits, etc.). That's why in the oz tests in this file we work around that (it's a bit elaborate but alas) - see `baseline-report-comment.yml`.

Something like (thanks Copilot):

1. Remove `pull-requests: write` from `.github/workflows/tests.yml` so the Tests workflow only has read-only permissions (e.g. `contents: read`).
2. Make the perf smoke job produce a small, plain artifact (or write a step-summary file) that includes:
   - the machine-readable PERF RESULT line (or a /tmp/perf-smoke-result.txt)
   - the PR number (pr_number.txt)
   Do not post comments directly from the Tests workflow.
3. Add a separate workflow (run from the default branch) that posts/updates the PR comment. Use the exact pattern in `.github/workflows/baseline-report-comment.yml`:
   - Trigger on `workflow_run` for the Tests workflow (types: [completed]).
   - That workflow definition lives on the default branch, so fork/PR edits can't change it.
   - Give that workflow only the minimal write scope it needs (for example in baseline-report-comment.yml it uses `contents: read`, `actions: read`, `pull-requests: write`).
   - The workflow downloads the artifact from the completed Tests run, reads the result/pr_number, and then creates or updates the PR comment (find-or-update comment logic like baseline-report-comment.yml).
4. Replace the current "Post PR comment" step in tests.yml with an artifact upload (actions/upload-artifact) of the result files, or with writing to the job summary and uploading that if you prefer.

arner (changes requested):
Very cool and useful!

arner (approved):
🥳

PR #325, opened 2026-08-17
fix revert funding

arner:
the testnode writes balances to a block 0. The hardhat tests revert to block 0 several times. The revertible LightKVS reverted to the genesis block, pre funding. That is now changed to the last written block 0.

This change is necessary to enable parallelization of the hardhat tests.


ale-linux (approved):

ale-linux (changes requested):
Actually sorry, I approved too quickly - shouldn't we then remove the tests from the flaky ones?

arner:
I looked into it. These bugs don't show up in a current, normal end to end run. It only happens when trying to run the openzeppelin test suites in a custom order (which will become necessary only when we do parallel runs, for which I'll open another PR later. So nothing to change in the baseline json for the current PR!

ale-linux (approved):

PR #326, opened 2026-08-17
truncate block store on revert

arner:
Simplifies the revert action on the block store. It fixes a beforeEach bug (snapshot for block N not found) which was caused by the % 128 ring buffer. OpenZeppelin baseline goes from 6094 → 6247 tests, 5144 → 5296 passing. It also removes the need to store databases in the /tmp directory.

### DIFF
--- a/gateway/storage/snapshot_store.go
+++ b/gateway/storage/snapshot_store.go
@@ -9,22 +9,11 @@ package storage
 import (
 	"context"
 	"fmt"
-	"os"
-	"path/filepath"
-	"sync"
-	"sync/atomic"
 
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
 )
 
 var storeLogger = flogging.MustGetLogger("gateway.storage.snapshot_store")
-
-const snapshotRingBufferSize = 128
-
-type snapshot struct {
-	BlockNumber uint64
-	DbPath      string // path to the snapshot database file
-}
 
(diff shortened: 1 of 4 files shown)


gateway/storage/snapshot_store.go
```diff
@@ -35,241 +24,70 @@ type Revertible interface {
...
+	}
+	defer tx.Rollback() //nolint:errcheck // no-op once committed
+
+	// Children first: transactions and logs both reference blocks(block_number)
+	// and the connection runs with foreign_keys=ON.
+	for _, stmt := range []string{
+		"DELETE FROM logs WHERE block_number > ?",
+		"DELETE FROM transactions WHERE block_number > ?",
+		"DELETE FROM blocks WHERE block_number > ?",
+	} {
```
kushnireyal, line 78: in production we have the `PRAGMA foreign_keys=ON` which makes the order you wrote here critical to not get an error, but the tests don't turn this flag on, so if i change the order here (as i tried) all tests pass (with the flag on as in production it does throw an error). Consider either turning this flag on for the tests (which breaks some existing tests that needs to be fixed) or to add a dedicated test that will enforce this order in the case someone changes it accidently in the future. Also, it'll be good to add to the tests log writings (currently they do only blocks and transactions)

kushnireyal (approved):
Looks good. See comment

PR #328, opened 2026-08-18
EIP-6780: add selfdestruct marker

arner:
plus finalise to prune empty addresses, and a decoding bug.

### DIFF
--- a/endorser/execution/executor.go
+++ b/endorser/execution/executor.go
@@ -89,6 +89,7 @@ func (e *EVMEngine) Execute(ctx context.Context, tx *types.Transaction, blockTim
 			return endorsement.ExecutionResult{}, err
 		}
 		// Revert: a committed outcome, endorsed as Status 201 with a revert event.
+		ex.state.Finalise(true)
 		event, mErr := fxcommon.MarshalRevert(ret, "", tx.Hash().Hex())
 		if mErr != nil {
 			return endorsement.ExecutionResult{}, fmt.Errorf("marshal revert event: %w", mErr)
@@ -101,6 +102,8 @@ func (e *EVMEngine) Execute(ctx context.Context, tx *types.Transaction, blockTim
 			Payload: ret,
 		}, nil
 	}
+
+	ex.state.Finalise(true)
 
 	var logs []byte
 	if l := ex.state.Logs(); len(l) > 0 {
--- a/endorser/execution/statedb.go
+++ b/endorser/execution/statedb.go
@@ -209,6 +209,7 @@ type StateDB struct {
 	refund           uint64
 	selfDestructed   map[common.Address]struct{}
 	newContracts     map[common.Address]struct{}                    // EIP-6780: contracts created in the current transaction
+	touched          map[common.Address]struct{}                    // EIP-161: addresses touched by a mutating op this transaction
 	accessList       *accessList                                    // EIP-2929/2930 access list (not persisted, reset per transaction)
 	transientStorage map[common.Address]map[common.Hash]common.Hash // EIP-1153 transient storage (not persisted, reset per transaction)
 
@@ -229,6 +230,7 @@ func NewStateDB(ctx context.Context, store ReadStore, namespace string, blockNum
 		monotonicVersions: monotonicVersions,
 		selfDestructed:    make(map[common.Address]struct{}),
 		newContracts:      make(map[common.Address]struct{}),
+		touched:           make(map[common.Address]struct{}),
 		accessList:        newAccessList(),
 		transientStorage:  make(map[common.Address]map[common.Hash]common.Hash),
 		journal:           make([]any, 0),
@@ -351,6 +353,7 @@ func (s *StateDB) putState(key string, value []byte) {
 
 // CreateAccount creates an account with zero balance and nonce.
 func (s *StateDB) CreateAccount(addr common.Address) {
+	s.touched[addr] = struct{}{}
 	s.putState(accKey(addr, "bal"), uint256ToBytes(uint256.NewInt(0)))
 	s.putState(accKey(addr, "nonce"), uint64ToBytes(0))
 	s.markNewContract(addr)
@@ -358,6 +361,7 @@ func (s *StateDB) CreateAccount(addr common.Address) {
 
 // CreateContract creates a contract account with empty code.
 func (s *StateDB) CreateContract(addr common.Address) {
+	s.touched[addr] = struct{}{}
 	s.putState(accKey(addr, "code"), []byte{})
 	s.markNewContract(addr)
 }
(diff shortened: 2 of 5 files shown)


endorser/execution/statedb.go
```diff
@@ -413,6 +421,7 @@ func (s *StateDB) GetNonce(addr common.Address) uint64 {
 
 // SetNonce sets the nonce of an account.
 func (s *StateDB) SetNonce(addr common.Address, nonce uint64, reason tracing.NonceChangeReason) {
+	s.touched[addr] = struct{}{}
```
ale-linux, line 424: We don't have this marker when `SetState`, `SelfDestruct`, `AddRefund`, `SubRefund`, `AddLog`,  are called.. why not?

endorser/execution/statedb.go
```diff
@@ -805,7 +826,22 @@ func (s *StateDB) Witness() *stateless.Witness { return nil }
...
+// transaction that ends up empty (zero balance, zero nonce, no code) gets its
+// account fields explicitly deleted, so the pruning is durable in the RWS
+// itself rather than relying on a replayer's own state object to infer it.
+func (s *StateDB) Finalise(deleteEmptyObjects bool) *bal.StateAccessList {
+	if deleteEmptyObjects {
+		for addr := range s.touched {
+			if s.Empty(addr) {
+				s.putState(accKey(addr, "bal"), nil)
+				s.putState(accKey(addr, "nonce"), nil)
+				s.putState(accKey(addr, "code"), nil)
```
ale-linux, line 839: what about state?

endorser/execution/executor.go
```diff
@@ -85,6 +85,7 @@ func (e *EVMEngine) Execute(ctx context.Context, tx *types.Transaction) (endorse
 			return endorsement.ExecutionResult{}, err
 		}
 		// Revert: a committed outcome, endorsed as Status 201 with a revert event.
+		ex.state.Finalise(true)
```
ale-linux, line 92: defer?

ale-linux (changes requested):
Thanks @arner, a couple of comments

endorser/execution/executor.go
arner: No defer because it has an effect on the read/write set, and if we defer it would be called after `ex.state.Result()` and we would miss its writes.

endorser/execution/statedb.go
arner: EIP-161 states that "An account is considered empty when it has no code and zero nonce and zero balance." (https://github.com/ethereum/EIPs/blob/master/EIPS/eip-161.md). So we only set "touched" for functions that can change any of these 3 values (and with that, the outcome of the Empty() call).

endorser/execution/statedb.go
arner: Yeah good question. We can't really enumerate the full state in the read-write set. On the one hand we don't have the interface for it, and on the other hand it would potentially make the read-write set extremely large. You could argue that it's not necessary: we can only call SSTORE (the opcode that leads to a PutState in str:contractaddr:slot) by executing the contract code (so "code" is not nil, and s.Empty is false).

The only way "code" can become nil is the same-tx-created selfdestruct. For the selfdestruct we create a marker (also in this PR) and don't even call Finalise. However, at the moment we only prune the *Chain triedb* contract state on selfdestruct (which we only enable in specific tests), but not the *statedb*. We have no pruning or compaction for the statedb at all...

That's probably the real issue; at the moment in this PR selfdestruct is not correctly implemented. It doesn't even clear the code... Solving it would mean something like:

1. On selfdestruct, not just null out the balance but also the code and nonce
2. For each of the state backends, delete the keys (or add a new tombstone)

(because if we don't clear the state, a CREATE2 metamorphic contract can read old state, which is bad...)

It would become a bit messy with the upcoming Query Service datastore though....

Open to ideas 😅

ale-linux: Hmm - so besu/geth etc. leak state on self destruct? If so, happy to do it too - but maybe can you quickly check whether that's the case?

endorser/execution/statedb.go
```diff
@@ -805,7 +826,22 @@ func (s *StateDB) Witness() *stateless.Witness { return nil }
...
+// Finalise implements EIP-161: any address touched by a mutating op this
+// transaction that ends up empty (zero balance, zero nonce, no code) gets its
+// account fields explicitly deleted, so the pruning is durable in the RWS
+// itself rather than relying on a replayer's own state object to infer it.
+func (s *StateDB) Finalise(deleteEmptyObjects bool) *bal.StateAccessList {
+	if deleteEmptyObjects {
+		for addr := range s.touched {
+			if s.Empty(addr) {
+				s.putState(accKey(addr, "bal"), nil)
+				s.putState(accKey(addr, "nonce"), nil)
```
ale-linux, line 838: one other thing... can we pls check that `putState(..., nil)` acts as a real delete, and that on fabric and fabric-x the account will really be detected as deleted?

gateway/storage/trie/store.go
```diff
@@ -129,6 +148,15 @@ func applyWrite(sdb *state.StateDB, w blocks.KVWrite) {
...
 
+// applySelfDestruct handles a "sd:<hexAddr>" marker: a same-tx-created account
+// self-destructed (EIP-6780). Prior-block storage slots this transaction never
+// touched can't be enumerated from the write set, so this delegates to the real
+// trie's own SelfDestruct rather than trying to delete matching "str:" keys by hand.
+func applySelfDestruct(sdb *state.StateDB, w blocks.KVWrite) {
+	addr := common.HexToAddress(strings.TrimPrefix(w.Key, "sd:"))
+	sdb.SelfDestruct(addr)
+}
+
```
ale-linux, line 159: so we self-destruct from the trie, but the endorser's own state does not do the same... are we okay with this?

gateway/storage/trie/store.go
```diff
@@ -67,6 +67,9 @@ func New(dbPath string, initialRoot common.Hash) (*Store, error) {
 	return &Store{diskdb: diskdb, db: db, root: initialRoot, persistent: persistent}, nil
 }
 
+// selfDestructMarkerPrefix flags a same-tx-created account that self-destructed (EIP-6780).
+const selfDestructMarkerPrefix = "sd:"
```
ale-linux, line 71: we appear to define this as a constant in two different places and yet we do

```
	addr := common.HexToAddress(strings.TrimPrefix(w.Key, "sd:"))
```

without using the constant. A rename anywhere is asking for trouble...

ale-linux (changes requested):

endorser/execution/statedb.go
arner: They don't... Because they use the TrieDB it's easy to remove the full state for a contract. So yeah it's a gap on our end!

arner:
You make very good points @ale-linux and it seems that my approach to EIP-6780 is wrong. Closing this, will create a new one with just the Finalise and decoding fix.

PR #337, opened 2026-08-20
fix: estimateGas requires padding

arner:
The EVM requires some buffer of gas to be sent at all times, so our original implementation of estimateGas, which just returned the gas used, was too tight. For a delegateCall, there is even more gas required to be carried across to the next call. Geth does a binary search to reach a specific, low number. Since we don't charge for gas, we don't have to be that precise so we just double until we have enough. This strikes a balance between finding the right answer fast (often in 1 go) and to still cover the tail of expensive multi-hop calls. Empty reverts are still returned correctly.

This commit also updates the oz_known_failures.json to be more precise about which tests are flaky and which are not, and adding a few known causes. Quarantined (known flaky) results are now reported with their cause and whether they succeeded this run or not: this should help to show which tests become non-flaky over time.

### DIFF
--- a/gateway/core/endorse.go
+++ b/gateway/core/endorse.go
@@ -138,14 +138,59 @@ func (e *EndorsementClient) CallContract(ctx context.Context, args ethereum.Call
 	return payload, err
 }
 
-// EstimateGas simulates the call and returns EVM usedGas.
-// Reverts and other call failures surface the same errors as CallContract.
+// estimateGasCeiling bounds the search. The endorser never charges for gas,
+// so there's no cost to simulating generously.
+const estimateGasCeiling = uint64(10_000_000)
+
+// EstimateGas returns a gas limit verified to work if resubmitted, not a raw
+// usedGas value: EVM rules like EIP-2200's SSTORE sentry and EIP-150's
+// 63/64 call-forwarding need gas in reserve during execution, not just
+// enough in total, so exact usedGas can sometimes still run out (view calls
+// and other simple cases usually don't hit this). Since gas isn't charged,
+// we generously double on each failure, until we hit the ceiling or succeed.
 func (e *EndorsementClient) EstimateGas(ctx context.Context, args ethereum.CallMsg, blockNumber *big.Int) (uint64, error) {
-	_, gas, err := e.call(ctx, args, blockNumber)
-	return gas, err
+	call := args
+
+	probeGas := func(gas uint64) (usedGas uint64, ok bool, err error) {
+		call.Gas = gas
+		_, usedGas, err = e.endorsers[0].Call(ctx, &call, blockNumber)
+		if err == nil {
+			return usedGas, true, nil
+		}
+		callErr, isCallErr := errors.AsType[*common.CallError](err)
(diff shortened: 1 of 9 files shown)


gateway/core/endorse.go
```diff
@@ -138,14 +138,59 @@ func (e *EndorsementClient) CallContract(ctx context.Context, args ethereum.Call
...
+// 63/64 call-forwarding need gas in reserve during execution, not just
+// enough in total, so exact usedGas can sometimes still run out (view calls
+// and other simple cases usually don't hit this). Since gas isn't charged,
+// we generously double on each failure, until we hit the ceiling or succeed.
 func (e *EndorsementClient) EstimateGas(ctx context.Context, args ethereum.CallMsg, blockNumber *big.Int) (uint64, error) {
-	_, gas, err := e.call(ctx, args, blockNumber)
-	return gas, err
+	call := args
+
+	probeGas := func(gas uint64) (usedGas uint64, ok bool, err error) {
```
ale-linux, line 154: Why not just add a configurable buffer on top of whatever gas we got from `e.call`?

ale-linux (changes requested):
Thanks Arne - just a small Q

gateway/core/endorse.go
arner: The usedGas return value of a Call is always too low, because it only returns the actual gas used. At first I thought, why not just add the 2300 (or 2301) for the Sentry check of EIP-2200. But the tricky part is from EIP-150, which handles nested calls (CALL and DELEGATECALL). It requires the parent frame to always keep 1/64th of the total budget. And it compounds with more deeply nested subcalls. Extra complication is that the code path can change based on `gasLeft()` from within the contract. And there are some other dynamic gas pricing rules that make it unpredictable.

Basically, not even geth can calculate gas reliably; it does a binary search between very high and very low gas limits until the transaction succeeds.... The best we could do was either return a crazy high static value (like we did at first), or get closer to what geth does.

- Returning the static value was not ideal because it made many openzeppelin tests fail ("assert gas cost < 30k") and it's possible that other tooling will not support it either (they don't always understand the implication of free gas so well because they're built for ethereum originally, so it's better to return a realistic number).
- The binary search seemed unnecessarily complex, so I went for a simpler exponential try window. We can overshoot by up to 2x the required gas, but we don't charge for it and this should still fall within the "realistic" window for tooling.

ale-linux: right but then again, why not double (or treble or whatever) the number we get from the EVM and return? that's the bit I don't get..

arner: Usually that's enough, but we can't guarantee it. The more subcalls, the less reliable the result. Even a cheap call can become expensive if it's deeply nested and every layer has to hold back 1/64th of the gas. We'll catch most of the situations by just doubling once (cheap), but the loop guarantees that we also support edge cases.

ale-linux (approved):
LGTM - after chatting

arner:
As discussed, will follow up with using MaxUsedGas + 2301 as the first guess.

Issue #329, opened 2026-08-18
[EPIC] WebSocket & filter RPC


arner:
Add support for WebSocket & filter RPC.

## Why

fxevm's gateway serves standard Ethereum JSON-RPC over plain HTTP only. Two closely related parts
of that standard surface are missing entirely, and both are documented as known, deliberate gaps
rather than oversights (`docs/COMPATIBILITY.md:95-96`, `:464-466`):

- **WebSocket transport.** No `ws://` (or `wss://`) endpoint at all — only HTTP.
- **Server-side filters and subscriptions.** No `eth_newBlockFilter` / `eth_getFilterChanges` /
  `eth_uninstallFilter`, no `eth_subscribe` / `eth_unsubscribe`. The documented workaround is "poll
  `eth_blockNumber` and `eth_getLogs` instead."

That workaround is fine for a one-off script, but not for the class of software this gap actually
blocks: anything that needs to react to new blocks or events rather than busy-poll for them —
indexers, monitoring/alerting, relayers, and any Ethereum client library that opens a WS connection
by default instead of falling back to HTTP polling. (Paladin, an EVM privacy-execution engine, is
one concrete example: its core blockchain client refuses to start without a working WS connection,
and its block indexer relies specifically on `eth_newBlockFilter`/`eth_getFilterChanges`.)

This epic closes both gaps.

## Foundational decision: WS shares the HTTP port

- fxevm serves WS on the **same** listener/port as HTTP (`gateway.Listen`), dispatched by the `Upgrade: websocket` header — exactly how go-ethereum's own `node` package runs HTTP and WS together (`node/rpcstack.go`'s `isWebsocket` / `httpServer.ServeHTTP`).
- This also matches the common Ethereum-node convention of one RPC port serving both transports
  simpler to document, firewall, and deploy than a second port only some callers would ever use.

Issue #331, opened 2026-08-18
Server-side filters (block + log)


arner:
Part II of https://github.com/hyperledger/fabric-x-evm/issues/329.

`eth_newBlockFilter` / `eth_newFilter` / `eth_getFilterChanges` / `eth_getFilterLogs` /
`eth_uninstallFilter` — block filters are the mechanism most block-indexing tooling relies on for
change detection; log filters are the standard way dApps/indexers watch for specific contract
events without hand-rolling a block-range cursor on top of `eth_getLogs`. (`eth_newPendingTransactionFilter` is explicitly not part of this — see decision 5.)

**Depends on:** https://github.com/hyperledger/fabric-x-evm/issues/330 — 
this is the transport most WS-oriented Ethereum tooling expects these
methods to be reachable over, even though the methods themselves are transport-agnostic and would
work over plain HTTP too.

### Decisions

1. **New `gateway/api/filters` package.** Houses `BlockFeed` (implements the SDK's
   `blocks.BlockHandler` — `Handle(ctx, blocks.Block) error`) and `FilterAPI` (the RPC-facing
   struct). Kept as a leaf package under `gateway/api` rather than `gateway/core`: it's an
   RPC-surface concern (filter/subscription bookkeeping), not chain/state logic, and unlike
   `gateway/core.Chain` it never needs `domain.*` types — only block hashes — so it doesn't
   inherit the `blocks.Block → domain.Block` conversion boundary the rest of `gateway/api`
   observes.
2. **Wiring.** `gateway/app.buildApp` constructs one `filters.NewBlockFeed()`, passes it as one of
   `extraHandlers ...blocks.BlockHandler` (`app.go:160,184` — already exactly the extension point
   used for this purpose, per `docs/design/multi-node-topology`) *and* into a widened
   `api.NewServer(b Backend, feed *filters.BlockFeed)`, which registers both `EthAPI` and
   `filters.NewFilterAPI(feed)` under the `"eth"` name. Confirmed safe: go-ethereum's
   `rpc.Server.RegisterName` (`rpc/service.go:60-92`) merges callbacks into the existing named
   service by method name rather than overwriting — geth itself does exactly this (its own
   `ethapi` and `filters.FilterAPI` both register under `"eth"`) — so `EthAPI` itself doesn't
   change.
3. **The hot path must never block on a slow RPC subscriber.** 
   `blocks.BlockHandler.Handle()` runs synchronously, in order, on the synchronizer's block-commit
[...]

PR #338, opened 2026-08-20
feat: serve JSON-RPC over WebSocket on the HTTP listener

### LINKED ISSUE #330: WebSocket transport
Part I of #329.

**Scope:**

- Extend `NewHTTPServer` (or the handler it returns) to check the `Upgrade: websocket` header
  (mirroring go-ethereum's own `isWebsocket`) and, when present, serve
  `srv.WebsocketHandler(nil)`; otherwise fall through to the existing `node.NewHTTPHandlerStack`
  chain unchanged. Both dispatch off the one `*rpc.Server` already built in `api.NewServer`.
- **Do not pass `[]string{"*"}` as the allowed-origins list** — that's an explicit opt-out of
  origin checking, and it's what turns this into Cross-Site WebSocket Hijacking (CSWSH) surface
  (browsers don't apply same-origin protection to WS handshakes the way they do to `fetch`/XHR;
  the server-side origin check is the only guard). Passing `nil`/empty instead activates
  go-ethereum's own built-in default (`rpc/websocket.go`'s `wsHandshakeValidator`, already vendored
  at v1.17.3, no new code required): requests with **no** `Origin` header pass through
  unconditionally (this covers every real client here — Paladin, ethers.js/web3.js in Node, curl —
  none of which send a browser-style `Origin` header), while a browser-originated cross-origin
  handshake is rejected unless it's from `http://localhost` or the host's own name.
- No change to `EthAPI`, `Backend`, or any existing RPC method beyond this — this issue is
  transport only. Plain (non-subscription) JSON-RPC calls over WS work immediately once this
  lands, since `rpc.Server` handles WS framing generically.

**Out of scope:** filters, subscriptions (part of other issues in #329).

**Acceptance criteria / testing:**

- New test dials the running server over `ws://` (go-ethereum's `rpc.DialContext` dispatches
  transparently by URL scheme, no separate WS-specific client needed) and calls an existing method
  (e.g. `eth_chainId`), asserts the correct response.
- Existing HTTP-path tests (`gateway/api/eth_test.go`, `eth_handlers_test.go`) continue to pass
  unmodified, plus one explicit test asserting HTTP and WS both work concurrently against a single
  running server instance — proves the dispatch doesn't regress the existing path.
- A request without the `Upgrade` header still falls through to the HTTP handler correctly
  (dispatch-correctness regression test, not just a happy-path WS test).
- **Origin checks (CSWSH):** a WS handshake with an `Origin: https://evil.example` header is
  rejected; a handshake with no `Origin` header at all (matching every real client's actual
  behavior) succeeds; a handshake with `Origin: http://localhost` succeeds. These three cases are
  what actually prove the `nil`-vs-`"*"` choice above, not just that WS works at all.
- No `docs/COMPATIBILITY.md` change here: "no WebSocket support" becomes literally false, but
  there's nothing behind it yet (no subscribe/filter methods) until the epic is completed.

Siddhartha-singh01:
## Description

Closes #330

Adds WebSocket JSON-RPC on the same listener as HTTP. Upgrade requests go to `rpc.Server.WebsocketHandler` with nil origins (geth default handshake checks, not `*`). Everything else keeps the existing HTTP stack. Logging wraps the HTTP path only so WS can Hijack. Also drops wildcard CORS and vhosts on HTTP as discussed.

Transport only. No EthAPI or Backend changes. Filters and subscriptions stay for later issues under #329.

## Type of Change
- [x] New feature

## Related Issue
#330

## Testing
- [x] WS dial via `rpc.DialContext` and `eth_chainId`
- [x] HTTP and WS concurrent on one server
- [x] HTTP without Upgrade still works
- [x] Origin: evil rejected, no Origin ok, localhost ok
- [x] `go test ./... -short` locally

## Checklist
- [x] Code builds
- [x] Unit tests pass
- [x] No COMPATIBILITY.md change (subscribe/filters not in yet)

### DIFF
--- a/gateway/api/server.go
+++ b/gateway/api/server.go
@@ -11,6 +11,7 @@ import (
 	"context"
 	"io"
 	"net/http"
+	"strings"
 
 	"github.com/ethereum/go-ethereum/node"
 	"github.com/ethereum/go-ethereum/rpc"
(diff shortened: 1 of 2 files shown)


gateway/api/server.go
```diff
@@ -41,14 +42,44 @@ func NewServer(b Backend) (*rpc.Server, error) {
 }
 
 // NewHTTPServer creates and configures an HTTP server without starting it.
+// The same listener serves JSON-RPC over HTTP and WebSocket: Upgrade requests
+// go to srv.WebsocketHandler (nil origins = geth default handshake checks);
+// everything else uses the HTTP stack. Logging wraps the HTTP path only so WS
+// upgrades can Hijack the connection.
```
arner: what do you mean WS upgrades can Hijack the connection?

arner (approved):
Looks good! Would you like to continue with https://github.com/hyperledger/fabric-x-evm/issues/331? If so, please reply to that issue and I'll assign you!

Siddhartha-singh01:
@arner Sure Sir ! 
Thanks !

PR #339, opened 2026-08-21
fix: off-by-one in delivery to notifier switch

arner:
- the hybridx synchronizer required delivery to reach the batch's own block number after the handlers had all processed the block. In practice this meant that the switch to notifier never happened.

- At the switch moment, there was some overlap between the two inputs, which causes some blocks to be processed twice. This commit changes the switch condition to only have covered everything _before_ it, and ensures only one identical block passes at the switch.

- AllTxStreamer does not support block hashes yet. Keeping them nil caused insert conflicts in the database (which we didn't notice because the switch never happened before). This commit temporarily uses block numbers as fake hashes, until https://github.com/hyperledger/fabric-x-committer/issues/773 is implemented.

### DIFF
--- a/common/notification.go
+++ b/common/notification.go
@@ -9,6 +9,7 @@ package common
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
 	"fmt"
 
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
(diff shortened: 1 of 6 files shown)


ale-linux (changes requested):
I'm not a great fan of the mutex-based solution, here's an alternative which imho is a bit cleaner and lock-free.

We use two integers:

1) keeps track of the currently claimed block
2) only set by delivery to signal when it is done dispatching that block

# Delivery 
- just tries a blind compare-and-swap to set the currently claimed block to the block it is about to dispatch:
    - if it wins the race it'll dispatch it; after dispatch is complete, it writes the block number it just handled in an atomic integer to signal it's done with its dispatch
    - if it loses the race, it disables itself forever

# Notification
- reads the currently claimed block first:
    - if it is its block number minus one it too tries a compare-and-swap to set it to its block number:
        - if it wins, it polls the other integer to see when delivery has finished, then dispatches this block (and all subsequent ones)
        - it if loses, it means delivery won again so it just drops the current block and will retry at the next
        
A couple of additional points:
- We need to seed integer 1 from store before we start, because delivery will automatically deliver from the next block and we can't break it
- we must have hard panics for failures, and probably need to open PRs:
    - technically we could survive a panic on delivery by re-asking for delivery of that block, but that'll likely panic again (unless it was some kind of transient I/O error or something)
    - notification panics are basically terminal and can only be handled by restarting delivery and the whole shebang

arner:
Right, that seems like a more elegant solution. Shall we close this one, so that you can make a new PR?

ale-linux:
done: https://github.com/hyperledger/fabric-x-evm/pull/343. If we merge that, we can then rebase this one on top to handle blocknum/hash, which I don't do in the other

arner:
Superseded by #343

PR #340, opened 2026-08-21
feat: add server-side block and log filters

Siddhartha-singh01:
## Summary
Follow up to #330 / #338 for issue #331.

Adds server side polling filters so indexers and dApps can watch new blocks and contract logs without rolling their own eth_getLogs cursor.

What this does:
1. New `gateway/api/filters` package with `BlockFeed` (implements `blocks.BlockHandler`) and `FilterAPI`
2. `Handle` never blocks the synchronizer commit path; a small buffer plus drain goroutine does the fan out
3. RPC methods: `eth_newBlockFilter`, `eth_newFilter`, `eth_getFilterChanges`, `eth_getFilterLogs`, `eth_uninstallFilter`
4. Wired in `buildApp` (feed in the handler chain) and registered under `eth` next to EthAPI for both prod and test servers
5. `App.Shutdown` closes the feed so the drain goroutine does not leak across testnode restarts
6. `docs/COMPATIBILITY.md` flipped for block and log filters; `eth_newPendingTransactionFilter` stays unsupported on purpose

Stayed close to the decisions on #331 (geth style IDs and 5 minute expiry, log matching off the hot path, panic isolation per filter). Pending tx filters are explicitly out of scope.

@arner ready for a look whenever you have a moment.

## Test plan
- [x] `go test -race ./gateway/api/filters/`
- [x] `go test -short -race ./gateway/...`
- [x] `go test -short ./...`


arner (dismissed):
Nice, seems to be close to the geth implementation! Thanks @Siddhartha-singh01 , no notes.

Siddhartha-singh01:
@arner Hello sir Can I start working on the another issue inside this Parent issue can you assign it to me

gateway/api/filters/feed.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	go f.drain()
+	return f
+}
+
+// SetSink registers the consumer that receives blocks from the drain loop.
+// Typically the FilterAPI. Safe to call once during wiring before traffic.
+func (f *BlockFeed) SetSink(s blockSink) {
+	f.sink.Store(s)
+}
```
ale-linux: The wiring is quite static, why can't we just set the sink pointer at construction time?

gateway/app/app.go
```diff
@@ -174,10 +176,14 @@ func buildApp(ctx context.Context, cfg config.Config, gwSigner sdk.Signer, logge
 		return nil, err
 	}
 
+	blockFeed := filters.NewBlockFeed()
+
 	// Chain must be called before gateway, to persist blocks before marking transactions complete.
-	handlers := append(extraHandlers, chain, gateway)
+	// blockFeed observes commits without blocking the hot path (see filters.BlockFeed.Handle).
+	handlers := append(extraHandlers, blockFeed, chain, gateway)
```
ale-linux, line 189: if we only wire this here and not under `./integration/` it means we'll never test this in the integration tests - is that okay?

gateway/api/filters/api.go
```diff
@@ -0,0 +1,294 @@
...
+	defer api.mu.Unlock()
+	id := rpc.NewID()
+	api.filters[id] = &filter{
+		typ:      LogsSubscription,
+		deadline: time.NewTimer(api.timeout),
+		crit:     crit,
+		logs:     make([]*types.Log, 0),
+	}
+	return id, nil
+}
```
ale-linux: this code can be refactored to avoid duplication

gateway/api/filters/feed.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+func (f *BlockFeed) deliver(b blocks.Block) {
+	v := f.sink.Load()
+	if v == nil {
+		return
+	}
+	v.(blockSink).onBlock(b)
+}
```
ale-linux: I think you can get rid of this if we simplify the creation of sink

gateway/api/filters/feed.go
```diff
@@ -0,0 +1,127 @@
...
+// consumers: it does a non-blocking send into an internal buffer and returns.
+// A dedicated drain goroutine fans blocks out to the registered sink.
+type BlockFeed struct {
+	in   chan blocks.Block
+	quit chan struct{}
+	done chan struct{}
+
+	sink atomic.Value // blockSink
+
+	dropMu   sync.Mutex
```
ale-linux: this isn't necessary: currently `Handle` can only be called by one goroutine at a time. At most I'd say we can use an atomic boolean but the mutex is too much imho

gateway/api/filters/api.go
```diff
@@ -0,0 +1,294 @@
...
+
+// onBlock is called from BlockFeed's drain goroutine.
+func (api *FilterAPI) onBlock(b blocks.Block) {
+	api.mu.Lock()
+	ids := make([]rpc.ID, 0, len(api.filters))
+	for id := range api.filters {
+		ids = append(ids, id)
+	}
+	api.mu.Unlock()
```
ale-linux:
```go
ids := slices.Collect(maps.Keys(api.filters))
```

gateway/api/filters/feed.go
```diff
@@ -0,0 +1,127 @@
...
+
+// blockSink receives committed blocks off the synchronizer hot path.
+type blockSink interface {
+	onBlock(b blocks.Block)
+}
+
+// BlockFeed implements blocks.BlockHandler. Handle never blocks on slow RPC
+// consumers: it does a non-blocking send into an internal buffer and returns.
+// A dedicated drain goroutine fans blocks out to the registered sink.
+type BlockFeed struct {
```
ale-linux, line 21: This is a solid approach: decouple the synchroniser goroutines that produce data from the worker that consumes it and updates the data structures. However if you look at how we do things right now you'll see that processing is always synchronous. I would like to stick with this approach for now, because it gives us the guarantee that if we register handlers in a specific order and the previous handler has returned, its data structure is guaranteed to have been updated. We already use this property at least once (endorser is registered first so that when the gw thinks a tx has been committed, its rwset has by definition already been committed in the KVS.

Let's please stick with it for now: `Handle` takes a lock and updates all filters synchronously.

I do agree with you that it's bad practice to do work from an I/O goroutine though. I would however solve the problem differently, by decoupling the I/O goroutine that reads out delivery/notification data from the network from the one calling the handlers. I would concretely then propose to

1. switch to synchronous processing here in this PR
2. open another PR to decouple I/O and processing at the place when we read data (e.g. in the SDK where we go

```go
// StreamAllTransactions opens the sidecar's StreamAllTransactions server-stream and
// delivers each TxEventBatch to processor until the context is cancelled or an error occurs.
func (p *Peer) StreamAllTransactions(ctx context.Context, req *notification.StreamAllRequest, processor notification.AllTxProcessor) error {
	stream, err := committerpb.NewNotifierClient(p.Connection()).StreamAllTransactions(ctx, toProtoStreamAllRequest(req))
	if err != nil {
		return fmt.Errorf("open stream-all-transactions: %w", err)
	}

	for {
		batch, err := stream.Recv()
		if err != nil {
			if err == io.EOF || ctx.Err() != nil {
				return nil
			}
			return fmt.Errorf("recv: %w", err)
		}
		sdkBatch := convertTxEventBatch(batch)
		if err := processor.ProcessBatch(ctx, sdkBatch); err != nil {
			return fmt.Errorf("process batch: %w", err)
		}
	}
}
```

_there_ we could add a channel and split

gateway/api/filters/match.go
```diff
@@ -0,0 +1,111 @@
...
+
+// logsFromBlock extracts Ethereum logs from a Fabric SDK block, attaching
+// block/tx context. Non-EVM and revert txs are skipped.
+func logsFromBlock(b blocks.Block) []*types.Log {
+	blockHash := common.BytesToHash(b.Hash)
+	out := make([]*types.Log, 0)
+	logIndex := uint(0)
+
+	for _, tx := range b.Transactions {
+		if len(tx.InputArgs) < 2 || len(tx.InputArgs[0]) != 1 || tx.InputArgs[0][0] != byte(fc.ProposalTypeEVMTx) {
```
ale-linux, line 28: that's the right approach, but it's somehow different from

https://github.com/hyperledger/fabric-x-evm/blob/main/gateway/core/chain.go#L201

@arner : it sounds like we're collecting logs even if the tx is not valid?

gateway/api/filters/criteria.go
```diff
@@ -0,0 +1,78 @@
+/*
```
ale-linux, line 1: This file appears to duplicate some of the code inside gateway/api/eth.go. Can we refactor the code so that we have a single implementation?

gateway/api/filters/api.go
```diff
@@ -0,0 +1,294 @@
...
+
+	blockLogs := logsFromBlock(b)
```
ale-linux, line 131: can we return early if len(ids) is zero?

ale-linux (changes requested):
Thanks @Siddhartha-singh01, great stuff! A couple of changes please..

@arner can you also please have another look?

Siddhartha-singh01:
@ale-linux thanks for the thorough review, that sync vs async point was the important one.

I pushed a follow up that covers what you asked for:

1. `Handle` is synchronous again. `BlockFeed` just calls into `FilterAPI` under the API lock and returns, same style as the other synchronizer handlers. No drain goroutine, no drop mutex, no `SetSink`.
2. `FilterAPI` is created up front and passed into `NewBlockFeed` / `NewServer` / `NewTestServer`.
3. Shared `CriteriaToLogFilter` (and log conversion) so `eth_getLogs` and the filter package use one path.
4. `NewBlockFilter` / `NewFilter` share an `install` helper.
5. Empty filter map returns early in `onBlock`.
6. Integration harness wires the same `BlockFeed` into the handler chain and registers that `FilterAPI` on the InProc RPC client, so filters are exercised there too.

On the log matching vs `chain.go` note: live filters still skip invalid txs. Happy to align `ConvertToDomain` in a separate change if you and @arner want that cleaned up.

Would appreciate another look when you have a moment.

gateway/api/filters/feed.go
arner: After a long discussion with @ale-linux I realised my mistake in the issue description. Indeed we don't need all this async stuff; the filter API is synchronous (install, get filtered logs, uninstall), and all we do with the handler is add a few items to an in-memory struct. I was under the impression that we would already push events through a websocket but that's not yet the case.

So yes, I agree that we can remove this complication. Actually @Siddhartha-singh01 , there's no need for the BlockFeed anymore. We can implement Handle on the FilterAPI directly.

gateway/app/app.go
```diff
@@ -179,10 +181,15 @@ func buildApp(ctx context.Context, cfg config.Config, gwSigner sdk.Signer, logge
 		return nil, err
 	}
 
+	filterAPI := filters.NewFilterAPI(gateway)
+	blockFeed := filters.NewBlockFeed(filterAPI)
+
 	// Chain must be called before gateway, to persist blocks before marking transactions complete.
-	handlers := append(extraHandlers, chain, gateway)
+	// blockFeed updates filter state synchronously in Handle, like the other handlers.
```
arner, line 188: Not necessary to mention; future readers don't care about the history.

Siddhartha-singh01:
@arner thanks, that clears it up.

Working through it now: dropping `BlockFeed` and putting `Handle` straight on `FilterAPI`, then wiring that into the synchronizer / app / integration paths. Keeping it local for the moment and will push once I have given it another pass.

gateway/api/filters/match.go
arner: Oops good catch, that's a bug! Posted a fix here: https://github.com/hyperledger/fabric-x-evm/pull/362

Siddhartha-singh01:
@ale-linux @arner thanks for the back and forth on this. I went through every point from the review and from Arne's note about dropping BlockFeed.

What changed:
1. BlockFeed is gone. FilterAPI implements Handle itself and updates the in memory filters under the API lock before returning.
2. FilterAPI is created up front and passed into the synchronizer handler chain, NewServer, and NewTestServer. No SetSink, no drain goroutine, no drop mutex.
3. NewBlockFilter and NewFilter share one install helper.
4. CriteriaToLogFilter is shared with eth_getLogs.
5. Handle returns early when there are no filters.
6. Integration wires the same FilterAPI into the handler chain and the InProc RPC client.

Local short tests for gateway and integration are green, plus race on the filters package. Would appreciate another look when you have a moment.

gateway/api/filters/criteria.go
ale-linux: We still need to do this please

ale-linux (changes requested):
I think we're good! One last thing remaining..

gateway/api/filters/criteria.go
Siddhartha-singh01: Done. eth_getLogs now calls into the shared CriteriaToLogFilter and DomainLogToTypes in the filters package, so that mapping lives in one place. Thanks for catching it.

Siddhartha-singh01:
@ale-linux thanks for the review. I cleaned up the last bit: eth_getLogs uses the shared criteria helpers from the filters package now, no duplicate mapping left in eth.go. Should be good for another look.

ale-linux (approved):
LGTM - this needs rebasing though, and pls on top of https://github.com/hyperledger/fabric-x-evm/pull/227 soon as we merge it

Siddhartha-singh01:
@ale-linux merged latest main into this branch and resolved the conflicts. Kept FilterAPI on the synchronizer and RPC paths, and picked up the hardhat state primer rewrite from #368 so the setBalance/setCode/setStorageAt commit wait failures should be gone. Also tightened filter expiry (absolute deadlines, faster sweep, no revive after expiry, GetFilterLogs refreshes the deadline). Ready for another look once CI finishes.

Siddhartha-singh01:
@ale-linux @arner hey, just a quick heads up on the OZ baseline tweak in this PR.

CI complained about one ERC20Votes test failing in a before each hook with: transaction . was dropped before it committed

I looked into it and it does not look related to the filters work. We already treat this kind of flake for Governor hooks. Same idea here: sometimes a setup tx gets dropped before it commits, and it is not consistent which Votes test hits it.

So I added a small flaky entry for the Votes family in testdata/oz_known_failures.json, matching what we already do for Governor. Happy to adjust it if you want it more specific.

Thanks

arner (approved):
Thanks for all the effort @Siddhartha-singh01 , this is a good feature!

Issue #323, opened 2026-08-17
Real hardhat_set* support (setBalance, setCode, setStorageAt) via endorsed system directives


arner:
## Problem

Hardhat's direct state-mutation RPCs — `hardhat_setBalance`, `hardhat_setCode`, `hardhat_setStorageAt` (and possibly `hardhat_setNonce`, not yet confirmed needed — see Non-goals) — all want the same thing: mutate an account's state directly, with no real EVM execution behind it, in a way that then behaves normally afterward (pay gas, get overwritten, etc.). Current state:

- `hardhat_setBalance`: not registered anywhere. 29 baseline entries currently carry this cause (`testdata/oz_known_failures.json`) — discovered indirectly, via `hardhat_impersonateAccount`'s stub (#252/PR #280) getting those same 29 tests past impersonation and onto this blocker instead.
- `hardhat_setCode`: registered, but a pure no-op stub (`HardhatAPI.SetCode`, `gateway/testimpl/hardhat_helpers.go:42-48`) — "return success without actually modifying state."
- `hardhat_setStorageAt`: not registered. 3 baseline entries, all `ERC1967Utils` proxy-slot fixture setup (`ADMIN_SLOT`/`BEACON_SLOT`/`IMPLEMENTATION_SLOT`) — makes sense, EIP-1967 proxy tests need to poke storage slots directly to set up test state.

None of these are small in isolation, but they're the *same* problem three times over. Worth building the shared mechanism once.

## Why not direct KVS injection (the obvious-looking shortcut)

There's an existing, working precedent for "commit a state change with no real transaction behind it": `gateway/testimpl/fund_accounts.go`'s `FundTestAccounts` (merged, used by #254) builds a `StateDB` directly, extracts the RWS, wraps it in a synthetic `blocks.Block`, and calls `kvs.Handle(ctx, block)` — the same ingestion entrypoint real committed blocks use, skipping signing/endorsement/ordering entirely.

[...]

PR #341, opened 2026-08-24
feat(gateway): implement hardhat_setBalance via endorsed system directive

arunabha003:
### Addressed issue

Part of #323 — PR 1 of 3 (shared infrastructure + `hardhat_setBalance`).
`hardhat_setCode` and `hardhat_setStorageAt` follow, reusing this infrastructure.

### Summary

`hardhat_setBalance` wasn't registered, so 29 OZ tests failed on it. Applying a
state change with no transaction behind it can't go through a direct KVS write:
the orderer never learns about it, so the endorser's and gateway's block heights
desynchronise. Instead a balance change is submitted as a non-EVM "system
directive" over the normal endorse → order → commit path, so the orderer assigns
the block number and both stores stay consistent.

- **common**: `ProposalTypeSetBalance` + an address/wei payload codec.
- **gateway**: `SubmitDirective` submits a directive and blocks until it commits,
  waiting on block height. It bypasses `ValidateTx`/`TxQueue`, which would reject
  an unsigned directive and never complete it (a directive leaves no eth receipt).
  This makes `hardhat_setBalance` synchronous, like Hardhat's own.
- **endorser**: a test-only `DirectiveEndorser` applies the balance and endorses
  the resulting RWS; wired only into the self-contained testnode.

Safety is build-time rather than a flag, per the issue discussion: the handler
lives only in `endorser/testimpl`, which production never imports, and
`core.Endorser.Execute` refuses non-EVM proposal types outright.

### Validation done

- `make checks` and `make unit-tests` pass.
- `TestTestNode_HardhatSetBalance` against a real in-process testnode: set +
  immediate read, delta-down, block-height progression, and spending from a
  funded account via a normal transfer. Passes under `-race`.
- OZ baseline re-run and updated: 26 entries blocked on `hardhat_setBalance` now
  pass.
  
 ### Screenshots 
  
  
image

### DIFF
--- a/common/proposal.go
+++ b/common/proposal.go
@@ -10,8 +10,9 @@ type ProposalType byte
 
 const (
 	ProposalTypeEVMTx ProposalType = 0xfb
-	ProposalTypeCall
-	ProposalTypeState
+	// ProposalTypeSetBalance marks a setBalance invocation's first arg so
+	// ConvertToDomain can filter it out; it is not a discriminator.
+	ProposalTypeSetBalance ProposalType = 0xfa
 )
 
 const (
--- a/endorser/testimpl/directive_endorser_test.go
+++ b/endorser/testimpl/directive_endorser_test.go
@@ -1,0 +1,123 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+*/
+
+package testimpl
+
+import (
+	"context"
+	"math/big"
+	"testing"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/hyperledger/fabric-protos-go-apiv2/peer"
+	estorage "github.com/hyperledger/fabric-x-evm/endorser/storage"
+	"github.com/hyperledger/fabric-x-sdk/blocks"
+	"github.com/hyperledger/fabric-x-sdk/endorsement"
+)
+
+const testNS = "basic"
+
+var directiveTestAddr = ethcommon.HexToAddress("0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266")
+
+// noopBuilder is a fake endorsement.Builder that records the read-write set it
+// was asked to endorse instead of signing anything.
+type noopBuilder struct {
+	lastRWS blocks.ReadWriteSet
+}
+
+func (b *noopBuilder) Endorse(_ endorsement.Invocation, res endorsement.ExecutionResult) (*peer.ProposalResponse, error) {
+	b.lastRWS = res.RWS
+	return &peer.ProposalResponse{}, nil
+}
+
+// balFromRWS returns the balance a read-write set writes for addr, or nil when
+// the set contains no balance write.
+func balFromRWS(rws blocks.ReadWriteSet, addr ethcommon.Address) *big.Int {
+	key := "acc:" + addr.Hex() + ":bal"
+	for _, w := range rws.Writes {
+		if w.Key == key {
+			if w.IsDelete {
+				return big.NewInt(0)
+			}
+			return new(big.Int).SetBytes(w.Value)
+		}
+	}
+	return nil
+}
+
+// commitRWS seeds kvs with rws through the same Handle path real commits use.
+func commitRWS(t *testing.T, kvs *estorage.LightKVS, ns string, rws blocks.ReadWriteSet) {
+	t.Helper()
+	blockNum, err := kvs.BlockNumber(context.Background())
+	if err != nil {
+		t.Fatalf("BlockNumber: %v", err)
+	}
+	block := blocks.Block{
+		Number: blockNum,
+		Transactions: []blocks.Transaction{{
+			ID:     "seed",
+			Number: 0,
+			Valid:  true,
+			NsRWS:  []blocks.NsReadWriteSet{{Namespace: ns, RWS: rws}},
+		}},
+	}
+	if err := kvs.Handle(context.Background(), block); err != nil {
+		t.Fatalf("Handle: %v", err)
+	}
+}
+
+// testInvocation builds a minimal invocation good enough for a nil builder's
+// Endorse to sign; the endorser under test doesn't inspect its contents.
+func testInvocation() endorsement.Invocation {
+	return endorsement.Invocation{}
+}
+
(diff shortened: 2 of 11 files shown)


common/proposal.go
```diff
@@ -12,6 +12,10 @@ const (
 	ProposalTypeEVMTx ProposalType = 0xfb
 	ProposalTypeCall
 	ProposalTypeState
+	// ProposalTypeSetBalance is a test-only system directive (hardhat_setBalance).
+	// Value is explicit: this block has no iota, so an omitted value would repeat
+	// 0xfb and collide with ProposalTypeEVMTx.
+	ProposalTypeSetBalance ProposalType = 0xfa
 )
```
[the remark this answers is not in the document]
arner: Actually, it turns out we're not even using ProposalTypeCall and ProposalTypeState, can you remove them please? The ProposalTypeEVMTx is scattered throughout the codebase but not doing anything useful. I would like to remove the ProposalTypeEVMTx in another PR (too much to include in this one).

arner (changes requested):
Good start, but I'd like to propose a slightly different approach.

We're relying on this ProposalTypeSetBalance byte as a discriminator to decide which function to call. Partly as an artifact from an older approach that actually lost its value by now.

Instead, the DirectiveEndorser could just have the functions "SetBalance" and later the others directly instead of using a switch on the type. The interface in the gateway would define that function directly rather than SubmitDirective.

arunabha003:
> Good start, but I'd like to propose a slightly different approach.
> 
> We're relying on this ProposalTypeSetBalance byte as a discriminator to decide which function to call. Partly as an artifact from an older approach that actually lost its value by now.
> 
> Instead, the DirectiveEndorser could just have the functions "SetBalance" and later the others directly instead of using a switch on the type. The interface in the gateway would define that function directly rather than SubmitDirective.

got it, working on this approach .

arunabha004:
@arner , i have updated the code, please review the latest pushes thanks !

arunabha003:
@arner , i have updated the code, please review the latest pushes thanks !

endorser/testimpl/directive_endorser_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/hyperledger/fabric-x-sdk/blocks"
+	"github.com/hyperledger/fabric-x-sdk/endorsement"
+)
+
+const testNS = "basic"
+
+var directiveTestAddr = ethcommon.HexToAddress("0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266")
+
+// lastRWS captures the read-write set from the most recent noopBuilder.Endorse call.
+var lastRWS blocks.ReadWriteSet
```
arner: nit: if we make this a field on noopBuilder, the tests can safely be run with `-parallel n`.

arunabha003:
hi @arner , does it look good? Upon your confirmation I can move towards the PR 2 . Whenever u get time please review. Thanks

gateway/app/testnode.go
arner, line 1: nice!

gateway/core/endorse.go
```diff
@@ -62,6 +62,37 @@ func (e EndorsementClient) ExecuteTransaction(ctx context.Context, tx *types.Tra
...
+}
+
+// balanceSetter is implemented only by the test-only directive endorser.
+type balanceSetter interface {
+	SetBalance(ctx context.Context, inv endorsement.Invocation, addr ethcommon.Address, amount *big.Int) (*peer.ProposalResponse, error)
+}
+
+// SetBalance forwards a setBalance directive to every endorser; tx is a synthetic
+// carrier for invocation plumbing, not the payload.
+func (e EndorsementClient) SetBalance(ctx context.Context, addr ethcommon.Address, amount *big.Int, tx *types.Transaction) (sdk.Endorsement, error) {
```
arner: looks like tx is unused, let's remove it

gateway/core/api.go
```diff
@@ -214,6 +224,87 @@ func (g *Gateway) SubmitFabricTx(ctx context.Context, hash common.Hash, end sdk.
...
+// observable. It deliberately bypasses ValidateTx/TxQueue (a directive is not a
+// user-signed EVM tx and would be rejected there), and since a directive leaves no
+// eth receipt, commit is observed by polling the target balance directly.
+func (g *Gateway) SetBalance(ctx context.Context, addr common.Address, amount *big.Int) error {
+	if got, err := g.BalanceAt(ctx, addr, nil); err == nil && got.Cmp(amount) == 0 {
+		return nil
+	}
+
+	// Synthetic tx: satisfies the endorser Execute signature and gives the envelope a hash.
+	tx, err := g.fakeDirectiveTx()
```
arner: Seems like the endorser does not need it anymore, now that we moved the feature out of the Execute function.

That means (at least looks like) we can get rid of this function completely; all we need is a hash. We can generate a new hash any way we like (either deterministic or random) - as long as it's only used once.

gateway/core/api.go
```diff
@@ -214,6 +224,87 @@ func (g *Gateway) SubmitFabricTx(ctx context.Context, hash common.Hash, end sdk.
...
+// only the random payload below needs to vary to keep hashes distinct.
+var (
+	directiveTxKeyOnce sync.Once
+	directiveTxKey     *ecdsa.PrivateKey
+	directiveTxKeyErr  error
+)
+
+// fakeDirectiveTx builds an unexecuted carrier tx; the shared key plus a random
+// payload give each directive a distinct hash so they do not collide downstream.
+func (g *Gateway) fakeDirectiveTx() (*types.Transaction, error) {
```
arner: see comments above, we might not need this.

arner (changes requested):

gateway/core/api.go
arunabha003: exactly, this makes sense...thanks . Now addressed

gateway/core/endorse.go
arunabha003: addressed

gateway/core/api.go
arunabha003: right, missed that one, now addressed

arunabha003:
@arner made the required changes. Let me know if all looks correct now. thanks

arner (approved):
Great @arunabha003 , thanks! I like the structure :)

PR #343, opened 2026-08-25
fix: off-by-one in delivery to notifier switch

ale-linux:
The synchronizer was waiting for delivery to reach the current block before switching to the notifier. Delivery can never actually outpace notification and so in practice that meant the switch never actually happened. This commit changes the condition so the switch happens once everything before that point has been processed, while handling the overlap between the two inputs so that no block is processed twice. The approach is lock-free.


arner (commented):
This will panic on switch, because hash and previous hash may not be null in the database.

That was fixed (workaround) in the other PR at https://github.com/hyperledger/fabric-x-evm/pull/339:

```
	// TODO: StreamAllTransactions carries no block header, so there's no real hash to
	// use here. Replace with the real hash once
	// https://github.com/hyperledger/fabric-x-committer/issues/773 is implemented.
	var parentNum uint64
	if batch.BlockNumber > 0 {
		parentNum = batch.BlockNumber - 1
	}
        b := blocks.Block{
		Number:       batch.BlockNumber,
		Hash:         blockNumberHash(batch.BlockNumber),
		ParentHash:   blockNumberHash(parentNum),
		Transactions: txs,
	}

// ...

// blockNumberHash encodes n as a 32-byte big-endian hash-shaped placeholder.
func blockNumberHash(n uint64) []byte {
	h := make([]byte, 32)
	binary.BigEndian.PutUint64(h[24:], n)
	return h
}
```

gateway/app/hybridx/hybrid.go
```diff
@@ -274,23 +272,84 @@ func (a *hybridAdapter) Handle(ctx context.Context, b blocks.Block) error {
...
+	return d == upTo
 }
 
 // dispatchBatch converts an AllTxBatch to a blocks.Block via AllTxBatchDispatcher
 // and dispatches it through the hybrid's handler chain.
 // The dispatcher is allocated once per gate in Start and stored here.
 func (g *notifGate) dispatchBatch(ctx context.Context, batch notification.AllTxBatch) error {
-	return g.dispatcher.HandleBatch(ctx, batch)
+	if err := g.dispatcher.HandleBatch(ctx, batch); err != nil {
+		// Fatal by design, and unrecoverable in place: the notification stream has
```
arner, line 383: It could be that the handlers return a context closed error (which is expected during graceful shutdown -- the context is wired all the way from the cmd process). That would be caught with a check:

```go
    if ctx.Err() != nil {
        return err
    }
```

arner: Actually, we need it in `common/notification.go` too... Could you add it there too above the panic?

gateway/app/hybridx/seeder.go
```diff
@@ -0,0 +1,120 @@
...
+		_ = throwaway.Start(peekCtx)
+	}()
+
+	select {
+	case first := <-got:
+		switch {
+		case first == expected:
+			// Normal case: peer delivered exactly the block we expected.
+		case expected == 0 && first == 1:
+			// fabrictest quirk: the in-process test network numbers its first
```
arner: Good catch, but let me fix this in fabrictest...

gateway/app/hybridx/seeder.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return fmt.Errorf("hybridx: seed claim: create throwaway synchronizer: %w", err)
+	}
+	go func() {
+		// Error is expected: peekCtx is cancelled as soon as the first block arrives.
+		_ = throwaway.Start(peekCtx)
+	}()
+
+	select {
+	case first := <-got:
```
arner: This blocks the thread until a new block comes in, so if we restart against an up to date chain (where new blocks are not added actively), the delivery and notifier don't start until someone creates a new transaction.

If there are issues connecting the synchronizer, it hangs indefinitely without logs. At the moment, now that we still use a different synchronizer for the endorser (to be changed soon), it also means that WaitUntilSynced report that the service is ready while it's actually unhealthy.

gateway/app/hybridx/seeder.go
arner, line 1: Is the seeder really needed? We already know the expected block from Synchronizer.Start ((lastBlock+1 if lastBlock>0, else 0). It looks like this might only be there to cover the fabrictest quirk of not starting with the expected block? If we start both counters with a sentinel value (math.MinInt64), deliveryShim.Handle can set the right values for dispatched and claimed. We could keep the missing-blocks check there too, by comparing the first block against expected, so we don't lose that.

It seems that we might not need the seeder at all, and that would make it simpler and more robust (only couple of lines of code, no second synchronizer and grpc connection needed, and Start doesn't block on the network before anything runs). Wdyt?

arner (changes requested):
Looks like a sound solution, but I think it might be able to do without the seeder part.

Can you also please copy the integration test from https://github.com/arner/fabric-x-evm/blob/6f4e4db4744a5bc9f4686863642f2dd5f18689ff/integration/hybridx_test.go?

gateway/app/hybridx/hybrid.go
ale-linux: done

gateway/app/hybridx/seeder.go
ale-linux: done - I got rid of the seeder

gateway/app/hybridx/seeder.go
ale-linux: gone

arner (approved):
Great, this is a cleaner solution than my previous one at github.com/hyperledger/fabric-x-evm/pull/339. Should we write an issue for the TODO? Thanks @ale-linux

ale-linux:
Yup - done (https://github.com/hyperledger/fabric-x-evm/issues/346). Thanks @arner, merging

PR #355, opened 2026-08-27
Idempotency for world state stores, with parity tests

arner:
- SDK upgrade, including stricter MVCC checks in fabrictest and an idempotent VersionedDB
- Idempotent block replay across LightKVS, PebbleKVS, and RevertibleLightKVS
- Error if content differs
- Fixes a Get embedding bug and de-duplicates version-assignment logic with LightKVS.
- PebbleKVS.commitBlock now also does per-write verify-or-error (was whole-block skip) to match the other stores
- ValidateDatabaseProtocol removed — PebbleKVS supports both protocols; found a DBMemory+fabric-x bug that will be tracked separately
- Expanded parity test suite (kvs_parity_test.go) running the same contract tests across all KVS backends


Storm1289:
Hi @arner, two things I wanted to check:

1. LightKVS / RevertibleLightKVS replay verification only covers the current tip - anything older than the tip is skipped unverified (since LightKVS keeps only final state), whereas PebbleKVS verifies every (key, block, tx) on replay. So the two backends give different idempotency guarantees: a stale older-than-tip redelivery with different content would be caught by Pebble but silently skipped by LightKVS. Is that gap acceptable for where LightKVS/memory is used, or worth tracking as a follow-up?

2. On the DBMemory + fabric-x gap you flagged (the per-key MVCC version being wrong across a delete/rewrite, with config.go allowing DBMemory for fabric-x uncaveated) - I searched the open issues and didn't find one tracking it yet. Are you planning to open a separate issue for it so it doesn't get lost?

arner:
1. I'm ok with this for now. It's just extra validation of something that can officially never happen. Wdyt? @ale-linux , should we do the hard verification on earlier blocks too?
2. Issue created at https://github.com/hyperledger/fabric-x-evm/issues/359

endorser/storage/pebblekvs.go
```diff
@@ -199,46 +202,55 @@ func (p *PebbleKVS) Update(updates []KeyValueVersion) error {
...
+					return fmt.Errorf("failed to decode existing write for %q: %w", u.Key, err)
+				}
+				if !bytes.Equal(rec.Value, u.Value) || rec.IsDelete != u.IsDelete || rec.TxID != u.TxID {
+					return fmt.Errorf(
+						"conflicting write for %q at block=%d tx=%d: existing (is_delete=%v tx_id=%s) differs from replayed (is_delete=%v tx_id=%s)",
+						u.Key, u.BlockNum, u.TxNum, rec.IsDelete, rec.TxID, u.IsDelete, u.TxID)
+				}
+				continue // identical replay: already stored, don't re-version it
+			}
+		}
```
kushnireyal, line 253: what if err is `pebble.ErrNotFound`? it seems that this case is not covered here but just falls out as no error. Is this intentional?

endorser/storage/revertible_lightkvs.go
```diff
@@ -116,47 +111,70 @@ func (kvs *RevertibleLightKVS) NewSnapshot(blockNumber *uint64) (execution.ReadS
...
+	for _, tx := range b.Transactions {
+		collectWrites(&updates, tx.NsRWS, b.Number, uint64(tx.Number), tx.ID, tx.Valid)
+	}
+
+	if kvs.hasCheckpoint.Load() && b.Number == current.BlockNumber {
+		return verifyReplay(current, updates)
+	}
+
+	return kvs.applyBlockSequential(b.Number, updates)
+}
```
kushnireyal, line 163: So we dropped the wraparound, is that intentional? can we solve the nextindex double-meaning issue but keep the wraparound? or do you think the wraparound is actually not so crucial?

kushnireyal (approved):
see comments

arner:
Thanks @kushnireyal, both good points. In the newest commit I error on empty blocks in Pebble, and reinstated the wraparound for the revertible light KVS.

common/notification.go
```diff
@@ -104,6 +104,7 @@ func (d *AllTxBatchDispatcher) HandleBatch(ctx context.Context, batch notificati
 		Number:       batch.BlockNumber,
 		Hash:         blockNumberHash(batch.BlockNumber),
 		ParentHash:   blockNumberHash(parentNum),
+		Timestamp:    time.Now().Unix(),
```
ale-linux, line 107: different nodes will get different timestamps - I guess that's not a problem because

- clients will get slightly different times when querying (so what)
- currently this information doesn't flow back to the endorser

right?

endorser/storage/lightkvs.go
```diff
@@ -232,30 +192,9 @@ func (r *Reader) Get(namespace, key string) (*blocks.WriteRecord, error) {
...
-				"version":   vv.Version,
-				"value":     truncateValue(vv.Value, 32),
-				"is_delete": vv.IsDelete,
-				"tx_id":     vv.TxID,
-			}
-			if jsonData, err := json.MarshalIndent(logRec, "", "  "); err == nil {
-				fmt.Printf("[LightKVS Get] %s\n", string(jsonData))
-			}
-		}
-
```
ale-linux, line 251: 🫠 thanks

endorser/storage/lightkvs.go
```diff
@@ -276,13 +215,32 @@ func (r *Reader) Close() error {
...
+// explicitly instead, so an empty block still advances the checkpoint.
+// Test only: unexported since nothing in production calls it (blocks only
+// ever arrive as blocks.Block, via Handle).
+func (kvs *LightKVS) update(updates []KeyValueVersion) error {
+	blockNum := uint64(0)
+	if len(updates) > 0 {
+		blockNum = updates[0].BlockNum
+	}
+	return kvs.applyBlock(blockNum, updates)
+}
```
ale-linux: can we then maybe move it to the test file?

endorser/storage/lightkvs.go
```diff
@@ -46,6 +46,10 @@ type LightKVS struct {
 
 	// Next index in the ring buffer to write to
 	NextIndex atomic.Uint32
+
+	// hasCheckpoint distinguishes an applied block 0 from a fresh store, which
+	// Current.BlockNumber alone cannot.
+	hasCheckpoint atomic.Bool
```
ale-linux: Let's move it away from here and make it a field of the `Snapshot` struct. We can call it `Empty` or `Placeholder` - defaults to false for all but the `Snapshot` created by the constructor. Then by atomically loading `Current` we also load the status of the snapshot and know whether it was populated

endorser/storage/lightkvs.go
```diff
@@ -379,22 +326,60 @@ func collectWrites(updates *[]KeyValueVersion, nsrwsList []blocks.NsReadWriteSet
...
+	// Applied even with no writes, so the checkpoint tracks ledger height.
+	return kvs.applyBlock(b.Number, allUpdates)
+}
+
+// verifyReplay checks a redelivery of the current tip against what's already
+// stored, erroring if the net effect of updates (last write per key wins,
+// matching applyBlock) differs from current. LightKVS stores only the final
+// per-key value after a block's writes, not each write individually, so
+// comparison happens at that same granularity.
+func verifyReplay(current *Snapshot, updates []KeyValueVersion) error {
```
ale-linux, line 344: I guess there's no way of verifying versions?

endorser/storage/lightkvs.go
```diff
@@ -350,6 +296,7 @@ func (kvs *LightKVS) Update(updates []KeyValueVersion) error {
 	// Atomically swap in the new snapshot
 	// New readers will see this snapshot; existing readers keep their old snapshot
 	kvs.Current.Store(newSnapshot)
+	kvs.hasCheckpoint.Store(true)
```
ale-linux: https://github.com/hyperledger/fabric-x-evm/pull/355/changes#r3960325996

this way we can remove this

endorser/storage/revertible_lightkvs.go
```diff
@@ -96,82 +101,94 @@ func (kvs *RevertibleLightKVS) NewSnapshot(blockNumber *uint64) (execution.ReadS
...
+func (kvs *RevertibleLightKVS) update(updates []KeyValueVersion) error {
 	blockNum := uint64(0)
 	if len(updates) > 0 {
 		blockNum = updates[0].BlockNum
 	}
-	newSnapshot := &Snapshot{
-		BlockNumber: blockNum,
-		Data:        newData,
+	return kvs.applyBlock(blockNum, updates)
+}
```
ale-linux: let's also move this one to the test file

endorser/storage/revertible_lightkvs.go
```diff
@@ -96,82 +101,94 @@ func (kvs *RevertibleLightKVS) NewSnapshot(blockNumber *uint64) (execution.ReadS
...
+
+// Handle overrides the promoted LightKVS.Handle so synchronizer-delivered
+// blocks share this type's own applyBlock — same ring wraparound as plain
+// LightKVS, but also maintaining the eviction floor that this type's
+// NewSnapshot/RevertToBlock need and LightKVS's don't (LightKVS's NewSnapshot
+// is exact-match only, so it never risks a stale nearest-older fallback).
+//
+// The replay guard itself mirrors LightKVS.Handle exactly: older blocks are
+// skipped unverified, and a tip redelivery goes through verifyReplay.
+func (kvs *RevertibleLightKVS) Handle(ctx context.Context, b blocks.Block) error {
```
ale-linux, line 134: this implementation of `Handle` is identical to the base kvs - why duplicate it?

endorser/storage/revertible_lightkvs.go
ale-linux: +1 - I'd like to keep it the previous way - no ring buffer and a hard error if we don't have enough history space (we can bump up).

ale-linux: and iiuc we can get rid of the oldest field - we basically either keep all the history we need to keep or fail the test (and need to rerun with a higher history count).

endorser/storage/revertible_lightkvs.go
```diff
@@ -319,15 +342,33 @@ func (kvs *RevertibleLightKVS) RevertToBlock(blockNumber uint64) error {
...
+	// OldestAvailable is untouched: this only discards a now-abandoned future branch.
+	kvs.clearFromRingIndex(targetIndex)
 
 	revertLogger.Debugf("RevertibleLightKVS.RevertToBlock() successfully reverted to block %d with merged state and trimmed future history", blockNumber)
 	return nil
 }
+
+// clearFromRingIndex discards every History entry at or after targetIdx's
+// snapshot, so the next commit resumes writing from that point.
+func (kvs *RevertibleLightKVS) clearFromRingIndex(targetIdx int) {
```
ale-linux: ...and we can get rid of this

endorser/storage/versioned_db_wrapper.go
```diff
@@ -75,9 +75,18 @@ func (s *VersionedDBSnapshot) Close() error {
...
 
-// Get retrieves the value for a key as of the given block number.
+// Get retrieves the value for a key as of the given block number, honoring the
+// KVS convention that lastBlock 0 means "latest" (see blockRefFromLastBlock) —
+// the same route LightKVS.Get and PebbleKVS.Get take. Passing 0 straight
+// through to VersionedDB.Get would instead match `version_block <= 0`, so
+// every latest-read would come back empty.
 func (w *VersionedDBWrapper) Get(namespace, key string, lastBlock uint64) (*blocks.WriteRecord, error) {
-	return w.db.Get(namespace, key, lastBlock)
+	r, err := w.NewSnapshot(blockRefFromLastBlock(lastBlock))
```
ale-linux, line 84: Ok but why can't we just call

```go
return w.db.Get(namespace, key, blockRefFromLastBlock(lastBlock))
```

ale-linux (changes requested):
Thanks Arne, a few comments from my side

common/notification.go
arner: Yeah... It's how we've been doing it so far in the Synchronizer, so this is just making sure it's doing the same. To be honest I'm not thrilled about it. Remember when we discussed to use the average timestamp of all the transactions in the block? We don't have that information after the roundtrip yet, but when we have all the metadata encoded in the transaction, we can do that.

Propose to keep it like this for now for consistency, and fix later.

endorser/storage/lightkvs.go
arner: No, because the committer doesn't give us versions. We increment on our own side...

endorser/storage/revertible_lightkvs.go
arner: That's a subtle one. kvs.applyBlock here refers to RevertibleKVS. In LightKVS it refers to LightKVS... So we have to override Handle, despite that it syntactically looks identical. Otherwise the RevertibleKVS never overrides the Handle.

endorser/storage/revertible_lightkvs.go
arner: I can't reply to your comment above "no ring buffer and a hard error", but I'm not a fan of that. 

`fxevm testnode` is meant for people to use for their local development workflow. You can work for hours and keep running your integration tests, so it's going to happen a lot and I don't want to have to explain it. We shouldn't panic after an arbitrary number of blocks, IMO this is not good enough of a reason for such drastic measures. "This" being the complexity cost of the solution here -- I have to say that that's not good either, and this function is terrible.

So we need a simpler solution. Would you like me to push a new commit to this PR? My preference would be to merge this first (once you're happy with the other parts) and do it as a followup, to keep it manageable. We would be merging working code, and I could start the cleanup right away.

ale-linux: > You can work for hours and keep running your integration tests, so it's going to happen a lot and I don't want to have to explain it

but if it happens with the fixed buffer, it'll happen with the ring buffer as well - there's no magic size that guarantees that things will always work with the ring buffer either.

We can document it, make the history size practically infinite for all intents and purposes, so imho the extra complexity isn't warranted.

endorser/storage/versioned_db_wrapper.go
arner: Good idea. I looked into it more deeply and this is only used by tests. I'm making a small change in the SDK (mvcc validator signature) and that will allow us to completely remove this whole function. That also makes it clearer when we're reading from history vs the current blockheight!

endorser/storage/revertible_lightkvs.go
ale-linux: just capturing here our offline discussion: we can just get rid of the problem by doubling the history array size every time we would reach the max. Acceptable for test code + helps keep things simple

arner: I reverted to the panic first, as it was before this PR and per your previous suggestion. Let's get this PR merged first, and then create a nice solution on a clean slate.

ale-linux (approved):
LGTM - thanks @arner

Issue #356, opened 2026-08-28
gateway: unreliable Fabric commit/validity signal in the domain model


Storm1289:
## Summary

The domain model's Fabric validity signal is unreliable. `domain.Transaction.FabricTxStatus` is set in `ConvertToDomain` from the SDK's `Transaction.Status`, which is never populated - so it is always 0. The nonce gate's `Observe` uses `FabricTxStatus != 0` to skip invalidated transactions (so an MVCC-invalidated tx does not wrongly advance a sender's nonce), but since the value is always 0, that check is currently a **no-op**.

The Ethereum receipt status can't be used instead, because a **revert still consumes the nonce** - EVM status 0 covers both a revert (nonce consumed) and an invalidated tx (nonce not consumed), so it can't tell them apart.

## Why it matters

The nonce gate must advance a sender's nonce only for transactions that actually consumed it (Fabric-valid commits, reverts included) and skip MVCC-invalidated ones. With the current signal it can't distinguish them, so once transactions genuinely get invalidated it could advance a nonce that was never consumed and strand that sender's later transactions.

## Proposed direction (fabric-x-evm side)

`ConvertToDomain` already uses `tx.Valid` (the SDK block-transaction's validity boolean) to derive the EVM receipt status, and it separates exactly the two cases the nonce cares about:
- a revert has `tx.Valid == true`  -> nonce consumed -> advance
- an MVCC-invalid tx has `tx.Valid == false` -> nonce not consumed -> skip

So: surface `tx.Valid` into the domain model, switch `Observe` to skip on `!valid` instead of `FabricTxStatus != 0`, and drop the dead `tx.Status` / `FabricTxStatus` plumbing.

## Depends on / open

- Confirm `tx.Valid` is reliably set on the committer path.
- A separate SDK-side fix for the unset `Transaction.Status` (maintainer-owned).

## Related

- #312 (commit EVM execution failures instead of dropping them)
- #257 (report accurate gasUsed in receipts)
- Surfaced in the #227 (nonce parking) review.

## Done when

The nonce gate's validity check uses a reliable signal (`tx.Valid`, surfaced into the domain model), and the dead `FabricTxStatus` / `tx.Status` plumbing is removed.

arner:
If I understand correctly, this does not depend on the SDK fix? Of course we should do something about it there, but this solution is separate from that?

Storm1289:
[...]

PR #358, opened 2026-08-31
feat(gateway): surface tx.Valid as FabricValid in the domain model

Storm1289:
## Summary

Part of #356.

Surface the reliable Fabric-validity signal into the domain model: `ConvertToDomain` now carries the SDK block transaction's `tx.Valid` as a new `domain.Transaction.FabricValid`. This is the signal the nonce gate needs to tell a Fabric-valid commit (nonce consumed, reverts included) from an MVCC-invalidated one - unlike `FabricTxStatus`, which is derived from the never-set `tx.Status` and is always 0.

## Changes

- `gateway/domain/models.go` - add `FabricValid bool`.
- `gateway/core/chain.go` - set `FabricValid` from `tx.Valid` in `convertTransaction`.
- `gateway/core/chain_test.go` - cover the new field.

## Scope / follow-ups

Producer side only. Two steps land separately:
- switching the nonce gate's `Observe` from `FabricTxStatus` to `FabricValid` - lives in #227 (nonce parking), and lands once the gate and this field are on the same branch.
- removing the dead `FabricTxStatus` plumbing (a DB column + sqlc regen) - a small separate follow-up.

## Tests

- [x] `go test ./gateway/core/...`
- [x] `make checks`

### DIFF
--- a/gateway/core/chain.go
+++ b/gateway/core/chain.go
@@ -159,7 +159,7 @@ func ConvertToDomain(b blocks.Block) domain.Block {
 			status = 1
 		}
 
-		etx, err := convertTransaction(tx.InputArgs[1], b.Hash, b.Number, tx.Number, tx.ID, status, tx.Status, tx.Events, &logIndex)
+		etx, err := convertTransaction(tx.InputArgs[1], b.Hash, b.Number, tx.Number, tx.ID, status, tx.Status, tx.Valid, tx.Events, &logIndex)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


Storm1289:
Hi @arner, please review whenever you have sometime

arner (approved):

arner:
This fixes the gap for now, let's revisit when https://github.com/hyperledger/fabric-x-sdk/pull/58 is merged.

PR #361, opened 2026-09-01
feat: add eth_subscribe newHeads over WebSocket

Siddhartha-singh01:
## Summary
Depends on #340 (Part II filters). This is Part III of #329.

Adds `eth_subscribe` / `eth_unsubscribe` for `newHeads` only, wired through `FilterAPI.NewHeads` so go-ethereum's rpc server picks up the subscribe methods by reflection.

What this does:
1. `BlockFeed.Subscribe` for live consumers with non blocking fan out and `SubscriberCount` for tests
2. `FilterAPI.NewHeads` uses the geth notifier pattern and selects on `rpcSub.Err()` so unsubscribe and WS close both release the goroutine
3. Header notifications carry the Fabric block hash (same identity as `eth_getBlockBy*`)
4. `docs/COMPATIBILITY.md` updated for `newHeads`; `logs` / `newPendingTransactions` subscriptions stay out of scope

@arner ready for a look after #340 lands (or stacked review is fine).

## Test plan
- [x] `go test -race ./gateway/api/filters/` (subscribe, backpressure, WS round trip, connection close cleanup)
- [x] `go test -race ./gateway/api/`
- [x] `go test -short ./...` for packages that run cleanly locally


gateway/api/filters/feed.go
```diff
@@ -0,0 +1,215 @@
...
+// blockSink receives committed blocks off the synchronizer hot path.
+type blockSink interface {
+	onBlock(b blocks.Block)
+}
+
+// BlockFeed implements blocks.BlockHandler. Handle never blocks on slow RPC
+// consumers: it does a non-blocking send into an internal buffer and returns.
+// A dedicated drain goroutine fans blocks out to the registered sink and to
+// any Subscribe callers (newHeads).
+type BlockFeed struct {
```
Storm1289: Before going further on this one: @ale-linux asked on #340 to make the handler path synchronous ("`Handle` takes a lock and updates all filters synchronously"), and @arner followed up with "there's no need for the BlockFeed anymore, we can implement Handle on the FilterAPI directly".

If that lands, this whole file goes away and `NewHeads` loses `feed.Subscribe`, which is what it is built on. Is this PR waiting on #340 settling first? It might save you a rework.

gateway/api/filters/head.go
```diff
@@ -0,0 +1,55 @@
...
+func headFromBlock(b blocks.Block) *rpcHead {
+	zero := (*hexutil.Big)(new(big.Int))
+	return &rpcHead{
+		Number:           (*hexutil.Big)(new(big.Int).SetUint64(b.Number)),
+		Hash:             common.BytesToHash(b.Hash),
+		ParentHash:       common.BytesToHash(b.ParentHash),
+		Sha3Uncles:       types.EmptyUncleHash,
+		LogsBloom:        make(hexutil.Bytes, 256),
+		TransactionsRoot: types.EmptyRootHash,
+		StateRoot:        types.EmptyRootHash,
```
Storm1289: Why is `stateRoot` hardcoded to `EmptyRootHash` here? `RPCBlock` in gateway/api/models.go sets it from `b.StateRoot`, so a client subscribed to `newHeads` and a client calling `eth_getBlockByNumber` get a different
`stateRoot` for the same block.

Same question for `transactionsRoot` on the line above: models.go uses `common.Hash{}` once the block has transactions.

gateway/api/filters/head.go
```diff
@@ -0,0 +1,55 @@
...
+	StateRoot        common.Hash      `json:"stateRoot"`
+	ReceiptsRoot     common.Hash      `json:"receiptsRoot"`
+	Miner            common.Address   `json:"miner"`
+	Difficulty       *hexutil.Big     `json:"difficulty"`
+	ExtraData        hexutil.Bytes    `json:"extraData"`
+	GasLimit         hexutil.Uint64   `json:"gasLimit"`
+	GasUsed          hexutil.Uint64   `json:"gasUsed"`
+	Timestamp        hexutil.Uint64   `json:"timestamp"`
+	MixHash          common.Hash      `json:"mixHash"`
+	Nonce            types.BlockNonce `json:"nonce"`
```
Storm1289: `GasLimit`, `MixHash` and `Nonce` are never assigned in `headFromBlock`. Deliberate?

gateway/api/filters/head.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/hyperledger/fabric-x-sdk/blocks"
+)
+
+// rpcHead is the newHeads notification payload. Hash is the Fabric block hash
+// (not keccak of the stubbed header fields), matching eth_getBlockBy* identity.
+type rpcHead struct {
```
Storm1289: `rpcHead` is now a second definition of the block header alongside `RPCBlock` in gateway/api/models.go, and it is already missing `totalDifficulty`, `size` and `baseFeePerGas`. `baseFeePerGas` in particular is one ethers.js reads off the head for fee estimation.

Can we derive the head from the existing type rather than declare a second one? Two structs describing the same block will drift

gateway/api/filters/criteria.go
```diff
@@ -0,0 +1,78 @@
...
+	"math/big"
+
+	gethfilters "github.com/ethereum/go-ethereum/eth/filters"
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/hyperledger/fabric-x-evm/gateway/domain"
+)
+
+// CriteriaToLogFilter maps geth FilterCriteria onto the store's LogFilter.
+// head is the current chain tip, used when fromBlock is latest/pending/omitted.
+// Same rules as gateway/api.EthAPI.filterCriteriaToLogFilter.
```
Storm1289: If the rules are the same, can we call that function instead of restating it? `CriteriaToLogFilter`, `resolveFromBlock` and `resolveToBlock` are near-verbatim copies of eth.go, and `domainLogToTypes` (api.go:309) is an exact copy of `domainLogToTypesLog`. This is the refactor @ale-linux asked for on #340

gateway/api/filters/head.go
Storm1289, line 20: This is a second block-header type alongside `RPCBlock` in models.go, already missing `baseFeePerGas`, `totalDifficulty` and `size`. Can we derive it from the existing one? Two structs for the same block will drift.

gateway/api/filters/head.go
Storm1289, line 48: `stateRoot` is hardcoded empty here, but models.go sets it from `b.StateRoot` - so `newHeads` and `eth_getBlockByNumber` disagree on the same block. Same for `transactionsRoot` above.

Also `GasLimit`, `MixHash` and `Nonce` are never assigned, and the miner address on line 50 is copied from models.go unexplained in both.

gateway/api/filters/api.go
```diff
@@ -0,0 +1,325 @@
...
+// NewHeads creates a subscription that pushes a header each time a block commits.
+// go-ethereum's rpc.Server reflects this method and exposes eth_subscribe/eth_unsubscribe.
+func (api *FilterAPI) NewHeads(ctx context.Context) (*rpc.Subscription, error) {
+	notifier, supported := rpc.NotifierFromContext(ctx)
+	if !supported {
+		return &rpc.Subscription{}, rpc.ErrNotificationsUnsupported
+	}
+
+	rpcSub := notifier.CreateSubscription()
+	feedSub := api.feed.Subscribe(16)
```
Storm1289, line 181: Why 16? It sets how far a slow client falls behind before we drop its notifications.

Separately, `Subscribe` after `Close` returns a subscription whose channel never closes, so that goroutine blocks forever.

gateway/api/filters/api.go
```diff
@@ -0,0 +1,325 @@
...
+	if !ok {
+		return false
+	}
+	delete(api.filters, id)
+	f.deadline.Stop()
+	return true
+}
+
+// GetFilterChanges drains buffered hashes or logs since the last poll.
+func (api *FilterAPI) GetFilterChanges(id rpc.ID) (interface{}, error) {
```
Storm1289, line 244: `interface{}` -> `any`, here and at 249 and 275.

gateway/app/app.go
```diff
@@ -174,10 +176,14 @@ func buildApp(ctx context.Context, cfg config.Config, gwSigner sdk.Signer, logge
...
 
+	blockFeed := filters.NewBlockFeed()
+
 	// Chain must be called before gateway, to persist blocks before marking transactions complete.
-	handlers := append(extraHandlers, chain, gateway)
+	// blockFeed observes commits without blocking the hot path (see filters.BlockFeed.Handle).
+	handlers := append(extraHandlers, blockFeed, chain, gateway)
 	gwSync, err := NewGatewaySynchronizer(cfg.Network.Protocol, chain, cfg.Network.Channel, cfg.Network.Namespace, cfg.Gateway.Committer.ToPeerConf(), gwSigner, logger, handlers...)
 	if err != nil {
+		blockFeed.Close()
```
Storm1289, line 186: `blockFeed.Close()` is repeated at six error returns . One defer with a success flag?

gateway/testimpl/server.go
Storm1289, line 26: Nil to disable a whole API fails quietly forget the feed and you get a server with no filter methods and no error. Can the tests pass a real `BlockFeed`?

testdata/oz_known_failures.json
Storm1289, line 1: These are all `hardhat_setBalance` entries that main already covers via #341 looks like rebase fallout. Own PR?

Storm1289 (commented):
Hi @Siddhartha-singh01, Most of the filter-side points here trace back to #340 and #362 rather than to this PR the synchronous handler rework, the eth.go duplication, and the tx.Valid divergence in log extraction. Once those settle it would be worth rebasing this on top of them, since the BlockFeed that `NewHeads` subscribes to may not survive #340.

The one thing that looks specific to this PR is the header: `newHeads` and `eth_getBlockByNumber` currently report different values for the same block.

Siddhartha-singh01:
@Storm1289 thanks for the careful review, that header mismatch was the real bug.

I rebased this on the latest #340 work (FilterAPI owns Handle now, no BlockFeed) and worked through your notes:

1. newHeads now encodes via RPCBlockFromDomain, same shape as eth_getBlockByNumber with full=false, including stateRoot from the stored block
2. FilterAPI sits after chain in the handler list so the block is already persisted when we build the payload
3. SubscribeHeads after Close returns an already closed channel
4. HeadsBuffer is a named constant for the lag window
5. GetFilterChanges uses any
6. buildApp closes FilterAPI with one defer on the failure path
7. WS and filter tests pass a real FilterAPI (and close in proc clients)
8. Dropped the oz_known_failures noise from this branch

Would love another look when you have a moment.

Siddhartha-singh01:
Merged the latest filters branch (#340) into this one and resolved the conflicts. Kept newHeads fan-out and FilterAPI-after-chain ordering, and picked up the hardhat primer / filter expiry fixes from main. Local short tests for api, filters, app, and integration are green.

gateway/api/filters/api.go
```diff
@@ -124,29 +135,34 @@ func (api *FilterAPI) timeoutLoop() {
...
 	}
-	var blockLogs []*types.Log
-	if needLogs {
-		blockLogs = logsFromBlock(b)
-	}
-	hash := common.BytesToHash(b.Hash)
-	for id := range api.filters {
-		api.deliverOneLocked(id, b, hash, blockLogs)
+	if len(api.headSubs) > 0 {
+		api.fanOutHeads(api.domainBlockFor(ctx, b))
```
arner: This is an expensive I/O operation (queries the db). We can release the lock while doing this so we don't block concurrent calls:

```go
	needHeads := len(api.headSubs) > 0
	api.mu.Unlock()

	if needHeads {
		block := api.domainBlockFor(ctx, b)
		api.mu.Lock()
		api.fanOutHeads(block)
		api.mu.Unlock() // or rely on the defer
	}
```

gateway/app/app.go
```diff
@@ -226,12 +226,20 @@ func buildApp(ctx context.Context, cfg config.Config, gwSigner sdk.Signer, logge
...
 	filterAPI := filters.NewFilterAPI(gateway)
+	ok := false
+	defer func() {
+		if !ok {
+			filterAPI.Close()
+		}
+	}()
 
 	// Chain must be called before gateway, to persist blocks before marking transactions complete.
+	// Keep FilterAPI before chain (same order as main) so commit-path behavior stays
```
arner: I'm not sure about this. Shouldn't we only notify listeners _after_ we stored the block? That would remove the need for the awkward retry in the filters package. Can we try to pot it after chain? If specific OZ tests fail after this, maybe we can see why they do. In this case it seems that data integrity is more important than speed.

gateway/api/filters/heads.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil
+	}
+	if len(b.StateRoot) > 0 {
+		return b
+	}
+	bq, ok := api.logs.(BlockQuerier)
+	if !ok {
+		return b
+	}
+	for range 20 {
```
arner: See comment on app.go... Hopefully we can smoothen this out somehow :)

arner (changes requested):
Shaping up nicely. Two points; maybe the first will help make the second possible...

gateway/api/filters/api.go
Siddhartha-singh01: Done. Handle updates filters under the lock, then unlocks before domainBlockFor, and only locks again for fanOutHeads.

gateway/app/app.go
Siddhartha-singh01: Agreed. FilterAPI is after chain again so newHeads only fires once the block is stored. Dropped the retry path.

gateway/api/filters/heads.go
Siddhartha-singh01: Removed BlockForHead. With FilterAPI after chain, domainBlockFor is enough.

Siddhartha-singh01:
@arner thanks for the notes. Addressed both points in ad34ad2: FilterAPI sits after chain again, and the DB read for newHeads no longer holds the API lock. Also removed the retry helper. Local unit tests and a focused MerkleTree hardhat run look good.

arner (approved):
Simpler this way, I like it! Thanks @Siddhartha-singh01

PR #362, opened 2026-09-03
fix: invalid eth or fabric tx cannot have logs

arner:
As discovered in https://github.com/hyperledger/fabric-x-evm/pull/340#discussion_r3921983455


gateway/core/chain.go
```diff
@@ -197,7 +197,7 @@ func convertTransaction(ethTxBytes []byte, blockHash []byte, blockNumber uint64,
 	hash := ethTx.Hash().Bytes()
 
 	var logs []domain.Log
-	if len(events) > 0 && !fc.IsRevertEvent(events) {
+	if ethStatus == 1 && len(events) > 0 {
```
ale-linux, line 201: why? Can we add some comment?

ale-linux (changes requested):

gateway/core/chain.go
arner: We used to only check if it's a Revert. But there is at least one more reason why a transaction could have status 0 (invalid): the execution failure. For instance for out-of-gas, invalid-opcode, stack-overflow, write-protection, depth-limit, or max-codesize. We have the status code in common/proposal.go: 

```
StatusExecFailure int32 = 460 // valid tx whose EVM execution failed (out of gas, invalid opcode, ...); should be mined (not yet)
```

But as the comment says we don't currently include it in a block (because the SDK filters out non-2xx statuses before we even submit it). This is a gap in compatibility that we can include in the overhaul of the InputArgs vs Payload update. So to be fair I don't think this bug/omission can actually happen today - though it's still safer to fix now.

I made this after you found it in https://github.com/hyperledger/fabric-x-evm/pull/340/changes#r3921983455 . I guess we have to update that place too with the same solution.

arner:
Closes https://github.com/hyperledger/fabric-x-evm/issues/312

ale-linux (approved):

PR #364, opened 2026-09-04
remove tx type filter from AllTxBatchDispatcher, move notification to hybridx

arner:
The type filter excluded the new hardhat_set* transaction types. This only happens after the hybridx synchronizer switches to the AllTxBatchDispatcher, which was visible only on constrained environments like the CI (intermittently). This commit removes that filter. This has very little effect on performance, even on namespaces with many mixed transaction types (which is not realistic anyway) and the ConvertToDomain function later filters the transactions before storing.

We move the AllTxBatchDispatcher from common to hybridx, because it's only for Fabric-X and exclusively used by the hybrid synchronizer.


ale-linux (changes requested):
I would not do this and go instead with https://github.com/hyperledger/fabric-x-evm/pull/368

Issue #359, opened 2026-09-01
LightKVS handles deletes differently from Fabric-X committer


arner:
LightKVS (and RevertibleLightKVS, which shares its version scheme) resets its version counter after a delete instead of staying monotonic, so a real fabric-x committer's worldstate would disagree with this backend's read-set across any delete/rewrite.

The test TestParityProtocolVersionCompatibility in the open PR https://github.com/hyperledger/fabric-x-evm/pull/355/changes#diff-4452150e0c807ea465e0495205fe8ebd9952432ca9e574e38053a40e24db624bR810 confirms this (currently skipped).

arunabha003:
can i get assigned to this issue ? @arner

arunabha003:
Hi @arner , took a look at this issue. The counting fix seems simple enough, LightKVS erases the key on delete, so the version counter goes with it and the next write starts at 0 again. Fix which i considered is to keep a tombstone instead, like PebbleKVS and the VersionedDB already do. implemented that and all four backends line up after that and the parity test passes with the skip removed.

but because of this change TestLocal started failing. deploys are fine, but the transaction after a deploy silently doesn't commit. Confirmed the fails were tombstones specific, not the counting. when a txn reads a deleted key the endorser records no version for it
but a store with tombstones still returns a record for it, and the MVCC validator doesnt check IsDelete so any non-nil record counts as "exists". So it thinks there's a conflict and rejects a tx that was fine. Zero writes are "deletes" in the EVM so this comes up a lot.

Then I ran the same validator against all four backends on plain main, no changes: LightKVS and RevertibleLightKVS pass, PebbleKVS and VersionedDB both get rejected. So it's already there for the tombstone backends . LightKVS just looked fine because erasing the key meant there was nothing to confuse the validator. Fixing #359 makes it consistent and the existing problem stops being hidden. Caveat: the validator I used is the one in the SDK's fabrictest, so this is the test harness rather than a real committer, though the comment on that check says it matches the committer's SQL.

So this issue probably needs a second part  either the validator ignores IsDelete records, or the store hides tombstones from it and keeps them internally for the counter. One's an SDK change, one stays here. 
So what do u think let me know .

arner:
[...]

PR #367, opened 2026-09-09
Bound PebbleKVS growth by latest-value keys plus a history window

kushnireyal:
PebbleKVS stored every write under a versioned key and never pruned so it grows forever and reads are slowed down as it grows. We remove the version from the key.
- result: 100k keys rewritten 50 times: reads from 43.5K/s before the change to 157.1K/s now. RAM 206 MiB before to 116 MiB now.
- Tradeoff: no more arbitrary historical reads


endorser/storage/pebblekvs.go
```diff
@@ -244,44 +309,186 @@ func (p *PebbleKVS) commitBlock(blockNum uint64, updates []KeyValueVersion) erro
 			Value:    u.Value,
 		})
 
-		if err := batch.Put(key, value); err != nil {
+		// Retain the record this write supersedes. Only the first write to a key in
+		// a block does so: later ones supersede a value created in this block, and
+		// time travel is block-granular.
+		if !seen {
```
arner: There's a subtle (but important) issue here for idempotent replay. If there are two transactions in the original block each writing to the same key, the code won't be able to find the first transaction (as it's overwritten by the second) and will hard abort with "block is already checkpointed but this write was never committed".

endorser/storage/pebblekvs.go
```diff
@@ -60,10 +75,9 @@ import (
...
 //     with the highest version whose block <= N (mirroring the sqlite-backed
-//     VersionedDB, see fabric-x-sdk state.VersionedDB.Get). LightKVS instead
-//     keeps only a small ring buffer of recent snapshots and errors for evicted
-//     blocks; PebbleKVS can serve any historical block. This is a strict
-//     superset of LightKVS's read behavior.
+//     VersionedDB, see fabric-x-sdk state.VersionedDB.Get), for any N inside the
+//     retained window. Outside it, NewSnapshot errors, as LightKVS does for blocks
+//     evicted from its ring. Arbitrary historical blocks are not supported.
 //   - Deletes are stored as tombstone records (IsDelete=true), matching the
 //     VersionedDB, rather than removing the key as LightKVS does. The execution
```
arner: Let's remove "rather than removing the key as LightKVS does.": this will be fixed as per https://github.com/hyperledger/fabric-x-evm/issues/359.

endorser/storage/pebblekvs.go
```diff
@@ -290,14 +497,28 @@ func (p *PebbleKVS) latestVersion(fullKey string) (int64, error) {
 // height (including 0 for genesis). Any historical block is serviceable; unlike
 // LightKVS this never errors for an "evicted" block.
```
arner: It looks like it does: `fmt.Errorf("snapshot not found for block number %d (retained history starts at %d, head is %d)",`

endorser/storage/pebblekvs_history_test.go
```diff
@@ -0,0 +1,459 @@
...
+	if err := kvs.Handle(ctx, mkBlock(1, 0, "tx1", true, "ns1",
+		blocks.KVWrite{Key: "k", Value: []byte("v1")})); err != nil {
+		t.Fatalf("Handle block 1: %v", err)
+	}
+
+	first, err := kvs.NewSnapshot(nil)
+	if err != nil {
+		t.Fatalf("NewSnapshot first: %v", err)
+	}
+	survivor, err := kvs.NewSnapshot(nil) // same height
```
arner, line 454: This tests two snapshots at the same height, but we don't cover lowestLivePinLocked for two concurrent snapshots at different heights.

Even better if we can also have something that runs concurrent NewSnapshot/Get/Close goroutines against a concurrently-committing writer... That could make sure we don't have any races (with `-race`).

arner (changes requested):
Looks good! A couple of comments. Regarding the parity tests:

- Maybe we can add a kvs_parity_test for the fact that both LightKVS and Pebble must error on evicted blocks?
- In TestParityReplayIsNoOp, lets also ensure that two writes to the same block survive a replay.
- Not really from your PR, but can we move TestPebbleUpdateRejectsMultiBlockBatch to pebblekvs_test.go?

endorser/storage/pebblekvs.go
kushnireyal: fixed. I changed it to compare the final value of each key like we do in lightkvs

endorser/storage/pebblekvs.go
kushnireyal: removed

endorser/storage/pebblekvs.go
kushnireyal: Done. I updated the comment to say the window is the last historySize blocks and older ones error

endorser/storage/pebblekvs_history_test.go
kushnireyal: Done, added both. The concurrent one found a real bug where a new reader could get its history pruned away, fixed it

kushnireyal:
> Looks good! A couple of comments. Regarding the parity tests:
> 
> * Maybe we can add a kvs_parity_test for the fact that both LightKVS and Pebble must error on evicted blocks?
> * In TestParityReplayIsNoOp, lets also ensure that two writes to the same block survive a replay.
> * Not really from your PR, but can we move TestPebbleUpdateRejectsMultiBlockBatch to pebblekvs_test.go?

All three done, added the evicted-blocks parity test (runs on lightkvs and pebble only, the other two keep more history so they don't error), added two writes to the same block in TestParityReplayIsNoOp, and moved TestPebbleUpdateRejectsMultiBlockBatch to pebblekvs_test.go. Also, simplified the code a bit

arner (approved):
Very nice, thanks @kushnireyal !

PR #368, opened 2026-09-09
Hardhat handlers simplified

ale-linux:


common/proposal.go
```diff
@@ -10,15 +10,8 @@ type ProposalType byte
...
-	// ProposalTypeSetBalance marks a setBalance invocation's first arg so
-	// ConvertToDomain can filter it out; it is not a discriminator.
-	ProposalTypeSetBalance ProposalType = 0xfa
-	// ProposalTypeSetCode marks a setCode invocation's first arg so
-	// ConvertToDomain can filter it out; it is not a discriminator.
-	ProposalTypeSetCode ProposalType = 0xf9
-	// ProposalTypeSetStorageAt marks a setStorageAt invocation's first arg so
-	// ConvertToDomain can filter it out; it is not a discriminator.
-	ProposalTypeSetStorageAt ProposalType = 0xf8
+	ProposalTypeCall
```
arner, line 13: We really shouldn't create fake ethereum "deployment" transactions for priming state. They are now processed by the full pipeline but they don't make sense. That's confusing for people and tools and can cause unwanted effects.

Let's just have one non-evm proposaltype for now, so we don't have to store them as fake transactions.

ProposalTypeStatePrimer ProposalType = 0xfa.

See below; we don't need the fake transaction to "waitForCommit", each operation has it's own way of validating that the effects have taken place.

PS we don't use call and state, let's not put them back here.

testutil/priming/state_primer.go
```diff
@@ -345,8 +345,25 @@ func (sp *StatePrimer) commitAndWait(end sdk.Endorsement, tx *types.Transaction,
...
 		// https://github.com/hyperledger/fabric-x-evm/issues/221
 		// is complete we will add the return again
-		waitForCommit(context.Background(), ec, tx)
+		WaitForCommit(context.Background(), ec, tx)
 	}
 
 	return nil
 }
+
+// ForceSetBalance drives addr's balance to exactly amount, unlike SetBalance which
```
arner: Then the name of SetBalance has to change to AddBalance. Or rather, we can probably just replace the old SetBalance with the new implementation (I'm going to make a wild guess and say that it is so far only used on addresses with zero balance...).

testutil/priming/state_primer.go
```diff
@@ -4,7 +4,7 @@ Copyright IBM Corp. All Rights Reserved.
 SPDX-License-Identifier: LGPL-3.0-or-later
 */
 
-package integration
+package priming
```
arner: We now have `endorser/testimpl`, `gateway/testimp`, and `testutil`. Next to `testdata` and `integration`... It's getting out of hand. I'd rather we park this under `gateway/testimpl`, even though it's cross disciplinary.

There is a difference between test utils and testimpl. The "hardhat" testnode is a real thing that is also used for development, not only for integration tests. We should probably revisit and clean this up a bit anyway, but for now let's not add yet another way to it.

gateway/testimpl/hardhat_helpers.go
```diff
@@ -10,67 +10,156 @@ DO NOT use in production environments.
...
+const primePollInterval = 50 * time.Millisecond
 
 // HardhatAPI provides Hardhat-specific RPC methods for testing.
-// Most methods are stubs that let Hardhat tests run; SetBalance, SetCode and
-// SetStorageAt are backed by real system directives via the gateway.
+// Most methods are stubs that let Hardhat tests run. The three state-changing ones
+// (setBalance, setCode, setStorageAt) are real: each commits a priming transaction
+// via StatePrimer, which endorses locally with every endorser's key and submits
+// through the normal commit flow. The endorsers themselves are never asked to
+// execute anything, so this works unchanged against remote/served endorsers.
```
arner, line 43: There are no remote endorsers in this flow.

gateway/testimpl/hardhat_helpers.go
```diff
@@ -10,67 +10,156 @@ DO NOT use in production environments.
...
+	what string,
+	apply func(*priming.StatePrimer),
+	confirm func(context.Context) (bool, error),
+) error {
+	if api.primer == nil {
+		return fmt.Errorf("%s: test RPC server was built without a state primer", what)
+	}
+
+	if ok, err := confirm(ctx); err == nil && ok {
+		return nil
```
arner: But also return err if it's not nil?

gateway/testimpl/hardhat_helpers.go
```diff
@@ -10,67 +10,156 @@ DO NOT use in production environments.
...
+	// Commit returns once the priming tx is no longer pending, but the gateway's
+	// state reads can still trail it by a block; poll until they agree.
+	waitCtx, cancel := context.WithTimeout(ctx, primeCommitTimeout)
+	defer cancel()
+	ticker := time.NewTicker(primePollInterval)
+	defer ticker.Stop()
+
+	for {
+		if ok, err := confirm(waitCtx); err == nil && ok {
+			return nil
```
arner, line 118: Or return error?

testutil/priming/state_primer.go
```diff
@@ -345,8 +345,25 @@ func (sp *StatePrimer) commitAndWait(end sdk.Endorsement, tx *types.Transaction,
 		// for commit loop unworkable. As soon as
 		// https://github.com/hyperledger/fabric-x-evm/issues/221
 		// is complete we will add the return again
-		waitForCommit(context.Background(), ec, tx)
+		WaitForCommit(context.Background(), ec, tx)
```
arner: Maybe we can inject in the caller context here, so that the RPC call can also be canceled by the client.

gateway/testimpl/server.go
```diff
@@ -30,7 +30,27 @@ import (
 // SECURITY WARNING: This server performs server-side transaction signing,
 // which is inherently insecure. Use ONLY for development and testing.
 // NEVER use in production environments.
-func NewTestServer(b api.Backend, testAccounts []common.Address, testAccountKeys map[common.Address]*ecdsa.PrivateKey, lightKVS estorage.Revertible, store storage.Revertible, pool TxPool) (*rpc.Server, error) {
+// TestServerOption configures optional test-server capabilities.
+type TestServerOption func(*testServerOpts)
```
arner: I'm not a fan of introducing this new pattern just for the state primer. Can't it just be a nillable constructor parameter?

Or maybe buildApp can have a cleaner constructor anyway, with all the test stuff grouped.

something like:
```
type testRPCDeps struct {
	kvs               estorage.Revertible
	builders          []endorsement.Builder // one per endorser
	accountsPath      string
	monotonicVersions bool
}
```

gateway/testimpl/hardhat_helpers.go
```diff
@@ -10,67 +10,156 @@ DO NOT use in production environments.
...
+	primeMu sync.Mutex
+}
+
+// NewHardhatAPI creates a new Hardhat API instance whose state-changing methods are
+// stubs. Use NewHardhatStateAPI to back them with a real state primer.
+func NewHardhatAPI() *HardhatAPI {
+	return &HardhatAPI{}
+}
+
+// NewHardhatStateAPI creates a Hardhat API whose setBalance/setCode/setStorageAt
```
arner: Hardhat is hardhat. By giving choice to the user whether that includes this particular feature or not, we make it more difficult for ourselves. If we just make the priming part mandatory, we can probably get rid of a lot of wiring and possible mutations.

integration/priming_alias.go
```diff
@@ -0,0 +1,23 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: LGPL-3.0-or-later
+*/
+
+package integration
+
+import "github.com/hyperledger/fabric-x-evm/testutil/priming"
```
arner: We now have `endorser/testimpl`, `gateway/testimpl`, and `testutil`. Next to `testdata` and `integration`... It's getting out of hand. I'd rather we park this under `gateway/testimpl/primer`, even though it's more or less cross disciplinary.

There is a difference between test utils and testimpl. The "hardhat" testnode is a real thing that is also used for development, not only for integration tests. We should probably revisit and clean this up a bit anyway, but for now let's not add yet another way to it.

Actually can't we get rid of this alias? Integration and gateway just import testimpl/primer?

gateway/testimpl/hardhat_helpers.go
```diff
@@ -10,67 +10,156 @@ DO NOT use in production environments.
...
+	defer api.primeMu.Unlock()
+
+	// Reset discards anything a previous call left queued, so this transaction
+	// carries only the write we are about to make.
+	if _, err := api.primer.Reset(); err != nil {
+		return fmt.Errorf("%s: reset primer: %w", what, err)
+	}
+	apply(api.primer)
+
+	if err := api.primer.Commit(ctx, true); err != nil {
```
arner: This should be "false"; below we already validate that the effects have actually taken place.

(This might mean that the integration test implementation can be simplified too but let's keep that for later)

One caveat is that the primer will need its own submitter, not submitters[0] due to the pooling. But that should be easy.

gateway/testimpl/hardhat_helpers.go
```diff
@@ -10,67 +10,156 @@ DO NOT use in production environments.
...
+	ctx context.Context,
+	what string,
+	apply func(*priming.StatePrimer),
+	confirm func(context.Context) (bool, error),
+) error {
+	if api.primer == nil {
+		return fmt.Errorf("%s: test RPC server was built without a state primer", what)
+	}
+
+	if ok, err := confirm(ctx); err == nil && ok {
```
arner: Or return error?

arner (changes requested):
The direction is good, this does make it simpler indeed. 

My biggest pet peeve is the fake ethereum transaction, but it looks like it can be resolved without too much hassle.

common/proposal.go
ale-linux: why? That mixes production with test code. The current approach just makes the priming tx "look" like an ethereum tx and no tool gets confused - and if they do we can change the fake ethereum tx to make it look like something tools don't understand and silently skip (which I'm sure is what most tools do).

gateway/testimpl/hardhat_helpers.go
ale-linux: ah but the point is that this is designed precisely with remote endorsers in mind - this setup works with them as well if ever we want to run the hardhat tests with multiple endorsers.

testutil/priming/state_primer.go
ale-linux: yep, good point!

testutil/priming/state_primer.go
ale-linux: yep, good point!

gateway/testimpl/server.go
ale-linux: yep, good point!

gateway/testimpl/hardhat_helpers.go
ale-linux: yep, good point!

integration/priming_alias.go
ale-linux: I think there's an import loop - but yeah, I hear ya... let me try to fix it

gateway/testimpl/hardhat_helpers.go
ale-linux: yep, good point!

common/proposal.go
arner: As discussed offline, I begrudgingly accept the fake ethereum tx 😅

arner: Can you still remove the ProposalTypeCall and ProposalTypeState please?

gateway/testimpl/hardhat_helpers.go
ale-linux: yep, good idea

gateway/testimpl/hardhat_helpers.go
arner: It's hard to reconcile remote endorsers with local keys, these are two different things. I'm not sure what this comment is trying to say, the code can endorse state changes, unrelated to remote or served endorsement.

arner (approved):
LGTM!

PR #372, opened 2026-09-11
chore: gate make checks on go fix

Storm1289:
Adds a go fix -diff gate to the checks target so modernizations are caught in CI instead of in review, and applies the single hit it reports today. The analyzer set grows with each Go release, so the gate gets stronger for free.


arner:
Good addition! Alas, it conflicts on the one fix 😅. If you rebase and fix the conflict, I'll approve! Thanks

Storm1289:
okay, wait doing it

Storm1289:
Hi @arner, fixed

arner (approved):
Thanks @Storm1289

PR #373, opened 2026-09-14
Add standalone endorser process

arner:
`fxevm start` can now run a bare endorser process (no gateway).

- Config.Gateway is now optional; standalone endorsers require endorser.server so they're actually reachable
- healthcheck works for either process shape via one plain TCP check
- Endorser and synchronizer construction unified across both process shapes, removing a couple of latent inconsistencies along the way (protocol-string handling, duplicated defaults, split logging).

Next PR will make sure that the gateway calls its internal endorser in go code rather than talking to it over GRPC.


gateway/app/app.go
```diff
@@ -318,7 +311,7 @@ func (a *App) Run(ctx context.Context) error {
 	g.Go(func() error { return a.synchronizer.Start(gctx) })
 
 	// Wait for initial sync before serving traffic
-	if err := WaitUntilSynced(gctx, a.synchronizer, a.cfg.Synchronizer.SyncTimeout()); err != nil {
+	if err := synchronizer.WaitUntilSynced(gctx, a.synchronizer, a.cfg.Synchronizer.SyncTimeout()); err != nil {
```
ale-linux, line 314: so the endorser app and the gateway app have two different synchronisers... but didn't we say the gw would have its own endorser, and both would be fed by the same syncrhoniser?

ale-linux (approved):
LGTM - thanks Arne. Just a Q but it shouldn't gate merging

gateway/app/app.go
arner: This function is used when we run a single process with an endorser and the gateway. The single endorser is built here. The handlers we pass (via buildApp) are: `kvs, filterAPI, chain, gateway`. In this case, we call NewEndorserCore which doesn't build and start it's own synchronizer.

If there's no gateway, we build and start a synchronizer (standalone.go).

PR #374, opened 2026-09-14
gateway: bound parked-transaction memory for abandoned gaps

### LINKED ISSUE #347: nonce gate: bound parked-transaction memory for abandoned gaps
The nonce gate parks future-nonce transactions per sender until the gap fills. `evictLRU` only evicts senders that have **no** parked transactions, so a sender that parks a transaction whose gap never fills is never evicted - a potential memory-exhaustion / DoS vector.

Today there are two partial guards:
- a per-sender cap (`maxPerSender`), and
- a parked-TTL sweep (`evictExpiredParked`) that runs only when that sender parks *again*.

So a sender that parks once and never returns keeps those entries indefinitely.

## What we should do
Remove parked transactions after a while independently of whether the sender parks again - e.g. a periodic age-based sweep, and/or a global bound on total parked transactions across all senders - so parked memory stays bounded even for abandoned gaps.

## Scope / done when
- Parked transactions past an age threshold are dropped regardless of sender activity.
- Total parked memory is bounded across senders.
- Existing nonce-gate tests stay green.

Storm1289:
## Summary
Closes #347.

The nonce gate parked future-nonce transactions until the gap filled, but the TTL sweep only ran when **that same sender** parked again, and LRU never dropped a sender that still held parked txs. A sender that parked once and never came back kept those entries forever.

This bounds parked memory without waiting for that sender to return.

## Changes
- Sweep expired parked transactions on any `Admit` and any `Observe` (including an empty block), across all senders not only when the abandoned sender parks again.
- Cap total parked transactions globally (4096). A new park drops the oldest across senders.
- Keep the existing per-sender cap (64).
- `Observe` still promotes a gap that just filled before sweeping TTL, so a late commit still releases the next nonce instead of throwing it away.
- Skip gRPC `Canceled` (not only `context.Canceled`) when aggregating 2-of-2 endorsement errors, so `nonce too low` is not hidden by a sibling RPC abort. Pre-existing flake from #306; same symptom as `TestFablo/nonce_validation`.
- Apply `go fix` on the nonce-gate tests (`for i := range n`, `atomic.Int32`) so this PR's files pass the `make checks` gate from #372.

## Files
| File | Change |
|---|---|
| `gateway/core/nonce_parking.go` | Cross-sender TTL sweep + global parked cap |
| `gateway/core/nonce_parking_test.go` | Abandoned-gap, global-cap, LRU-after-TTL, concurrent-cap tests; `go fix` on this file |
| `gateway/core/api_nonce_gate_test.go` | Empty `Handle` expires parked txs; client can resubmit |
| `gateway/core/endorse.go` | Treat gRPC `codes.Canceled` as canceled when aggregating 2-of-2 errors |
| `gateway/core/endorse_test.go` | Rejection still wins when the other endorser returns gRPC `Canceled` |

Out of scope: `go fix` hits in `#340` filter code and other packages. Those are not this PR; they can be a follow-up.

## Tests
- `go test ./gateway/core -race`
- `go test ./gateway/api/filters ./endorser/execution ./gateway/api ./gateway/api/rpcerr`
- `go test ./... -short`
- `TestFabricX` (including `two_of_two_endorsement_policy`) on a fresh Fabric-X network
- `TestLocal` / `TestLocalX`
- `make perf-smoke` (10k USDC, 0 failed)
- `TestFablo` via `make integration-tests` (nonce_validation flake addressed by the gRPC `Canceled` skip)

## Test plan
- [x] `go test ./gateway/core -race`
- [x] `make test-x` (fresh `start-x`)
- [x] `make integration-tests`
- [ ] `make checks` on CI (this PR's files are `go fix` clean; filter files from `#340` on `main` may still trip the gate)


Storm1289:
Hi @ale-linux @arner, take a look whenever you have sometime

ale-linux:
Thanks @Storm1289 - I went for this simpler approach instead - wdyt?

Storm1289:
Looks good, thanks.

One thing: with the global parked cap gone, total parked is bounded by arrival rate over the TTL rather than by a fixed number  20k senders parking one each holds 20k. Fine by you, or worth keeping a hard cap?

arner (approved):
LGTM

PR #375, opened 2026-09-15
Bump to latest SDK

ale-linux:


endorser/storage/lightkvs.go
```diff
@@ -25,6 +25,13 @@ var (
 
 // KVS is implemented by both LightKVS and VersionedDBWrapper.
 // It combines snapshot reads, block handling, and lifecycle management.
+// There is one read API per question a caller can ask. Get (from
+// blocks.RecordGetter) answers "what is the current value", and NewSnapshot (from
+// execution.KVSSnapshotter) answers "what was the value at block N" — note that its
+// nil means latest while its 0 means genesis, so the two cannot be collapsed into one
+// signature. Get used to take a lastBlock too, with the opposite convention that 0
```
arner: We don't have to explain what used to be.

endorser/storage/versioned_db_wrapper.go
```diff
@@ -75,13 +75,13 @@ func (s *VersionedDBSnapshot) Close() error {
...
-// through to VersionedDB.Get would instead match `version_block <= 0`, so
-// every latest-read would come back empty.
-func (w *VersionedDBWrapper) Get(namespace, key string, lastBlock uint64) (*blocks.WriteRecord, error) {
-	r, err := w.NewSnapshot(blockRefFromLastBlock(lastBlock))
+// Get implements blocks.RecordGetter, reading the latest committed state — the same
+// route LightKVS.Get and PebbleKVS.Get take. It resolves the height through
+// NewSnapshot rather than calling VersionedDB.Get directly, because that method takes
+// a lastBlock and matches `version_block <= lastBlock`, so there is no value to pass
+// for "latest": 0 would come back empty for every key.
+func (w *VersionedDBWrapper) Get(namespace, key string) (*blocks.WriteRecord, error) {
```
arner, line 79: This function body can be just 

```go
return w.db.GetCurrent(namespace, key)
```

To save us a DB call and the snapshot creation. The comment would suffice with the first half sentence.

gateway/core/chain_test.go
```diff
@@ -222,38 +221,44 @@ func TestConvertTransaction_InvalidSignature(t *testing.T) {
...
 
 	tests := []struct {
-		name           string
-		ethStatus      uint8
-		validationCode int
-		fabricValid    bool
+		name         string
+		ethStatus    uint8
+		fabricStatus blocks.Status
+		fabricValid  bool
```
arner: In principle there's only one "valid" status: blocks.StatusCommitted. So we can drop the fabricValid field (which was there because the SDK wasn't propagating statuses properly before)..

gateway/domain/models.go
```diff
@@ -30,9 +30,13 @@ type Transaction struct {
 	ContractAddress []byte
 	Status          uint8 // EVM receipt status: 1 is success, 0 is a revert or an invalid tx
 	FabricTxID      string
-	FabricTxStatus  int   // from the SDK Transaction.Status, which is currently never set
-	FabricValid     bool  // Fabric-valid commit: the nonce was consumed, reverts included
-	Logs            []Log // populated for receipt queries
+	// FabricTxStatus is the SDK's protocol-neutral blocks.Status for the commit
+	// (1 committed, 3 MVCC conflict, and so on — see blocks.Status). Both the
+	// delivery and notification paths now populate it; it used to be left at 0 on
```
arner: We don't have to say what it used to be like.

gateway/domain/models.go
```diff
@@ -30,9 +30,13 @@ type Transaction struct {
...
+	// the delivery path, so older rows carry 0 regardless of outcome.
+	FabricTxStatus int
+	FabricValid    bool  // Fabric-valid commit: the nonce was consumed, reverts included
```
arner: See above, I'd say we might as well remove it?

endorser/storage/versioned_db_wrapper.go
ale-linux: good point - fixed

endorser/storage/lightkvs.go
ale-linux: yep - gone

gateway/domain/models.go
ale-linux: done

arner (approved):
That cleans up nicely, thanks @ale-linux !

PR #376, opened 2026-09-16
storage: remove unused blocknum from StateDB

arner:
We don't need blocknum in the StateDB, because it's already operating on a snapshot. This simplifies the mental model and several call sites.

+ go fix to modernize some patterns.


ale-linux (approved):
I love less code!! ❤️

PR #377, opened 2026-09-16
fix(endorser): keep LightKVS versions monotonic across deletes

### LINKED ISSUE #359: LightKVS handles deletes differently from Fabric-X committer

arunabha003:
## Addressed issue

  Closes #359.

  ## Summary

  - LightKVS now stores a **tombstone** instead of removing the key, and assigns
    versions from a **per-batch counter** — the same scheme
    `PebbleKVS.commitBlock` already uses. The count survives a delete, and
    repeated writes to one key within a block get consecutive versions rather
    than sharing one. `RevertibleLightKVS` picks this up through the shared
    write path.
  - `RevertToBlock`'s merge collapses as a result: with tombstones, a key present
    in the target snapshot is always still present in the current one, so the two
    set-difference branches become a single pass over the current snapshot.
  in the target snapshot is always still present in the current one, so the two
  set-difference branches become a single pass over the current snapshot.
- `TestParityProtocolVersionCompatibility` already covered this and was skipped
  for both LightKVS backends. The skip is removed and they now assert the same
  `MAX(version)+1` scheme as PebbleKVS and VersionedDB.

Note :

- This depends on the MVCC validator fix in hyperledger/fabric-x-sdk#59, which
  arrived here via #375. Without it the validator treats any non-nil record as
  "key exists", so a transaction reading a tombstoned key is rejected as a
  conflict.
- Tombstones are not reclaimed, so deletes no longer free the map entry. Happy
  to follow up with a bound if that matters for the in-memory backend.

## Validation done

- `make checks` and `make unit-tests` pass.
- `TestParityProtocolVersionCompatibility` passes unskipped for `LightKVS`,
  `RevertibleLightKVS`, `PebbleKVS` and `VersionedDB`, in both `fabric` and
  `fabric-x` modes.
- `TestLocal` and `TestLocalX` pass.
- All four backends now produce the same version sequence over
  write/delete/rewrite/delete/rewrite:

## Screenshots
image


endorser/storage/kvs_parity_test.go
```diff
@@ -46,7 +46,7 @@ func kvsBackends() []kvsBackend {
 				return NewLightKVS(historySize)
 			},
 			persistent:   false,
-			mvccVersions: false,
+			mvccVersions: true,
```
arner: Now that all mvccVersions are true, let's remove this entire field and the if statements.

endorser/storage/lightkvs.go
```diff
@@ -205,9 +205,12 @@ func (r *Reader) Close() error {
...
 
 // applyUpdates computes new snapshot data by applying updates on top of
-// oldData, assigning each write the existing version + 1 (or 0 for a new
-// key). Shared by LightKVS.applyBlock and RevertibleLightKVS.applyBlock,
-// which only differ in how they track the eviction floor when wrapping.
+// oldData. Mirrors PebbleKVS.commitBlock's version scheme: a delete is stored
+// as a tombstone rather than removing the key, and each write's version is
+// MAX(version)+1 for its key, so multiple writes to one key within a batch
+// get consecutive versions and the counter never resets across a tombstone.
+// Shared by LightKVS.applyBlock and RevertibleLightKVS.applyBlock, which only
```
arner: Let's refer less to other parts of the code in our comments. Keep it short and to the point, about what this function does, not whether it's analogous to some other kvs.

arner (dismissed):
Awesome @arunabha003 , good to see the parity tests give us confidence about how these stores work. Just a couple of small notes.

endorser/storage/lightkvs.go
arunabha003: addressed

endorser/storage/kvs_parity_test.go
arunabha003: addressed

endorser/storage/lightkvs.go
```diff
@@ -345,13 +348,10 @@ func verifyReplay(current *Snapshot, updates []KeyValueVersion) error {
 	}
 	for key, u := range final {
 		existing, ok := current.Data[key]
-		if u.IsDelete {
-			if ok {
-				return fmt.Errorf("conflicting write for %q at block=%d: existing value present, replayed is a delete", key, u.BlockNum)
-			}
-			continue
+		if !ok || existing.IsDelete != u.IsDelete || existing.TxID != u.TxID {
+			return fmt.Errorf("conflicting write for %q at block=%d: replayed content differs from existing", key, u.BlockNum)
```
ale-linux: this message is less informative than the old one right?

ale-linux (commented):
@arner could you also pls have a look?

endorser/storage/lightkvs.go
arner: Hmm yes let's go back to a more specific message, also if that involves some more nesting. If this ever happens, it's really bad so users have to know exactly what broke.

arner (commented):
As far as I'm concerned this is ready to merge after the specific message is back, wdyt @ale-linux ?

endorser/storage/lightkvs.go
arunabha003: restored the specific messages and split the cases out. Does it look good now?

arner (approved):
Thanks for your contribution @arunabha003 !

PR #380, opened 2026-09-18
feat: add resource caps for filters and newHeads subscriptions

### LINKED ISSUE #333: Per-client resource caps on filters and subscriptions
Part IV of https://github.com/hyperledger/fabric-x-evm/issues/329.

**Problem:** creating a filter or subscription is one cheap RPC call that then holds server-side
memory (map entry, timer, growing slice, or a goroutine + channel) for up to the 5-minute expiry
(filters) or the life of the connection (subscriptions) — with no limit today on how many a single
client can accumulate. That's unbounded-allocation exposure (CWE-770): a client can create them
faster than they can expire and sustain memory growth indefinitely. Not solved by this epic's
backpressure decisions, which only protect the block-commit hot path from a slow *existing*
subscriber — they say nothing about how many can exist.

**Depends on:**
- https://github.com/hyperledger/fabric-x-evm/issues/331
- https://github.com/hyperledger/fabric-x-evm/issues/332

(caps both the filter registry and the subscription registry those issues create).

**Scope:**

- **Filters: a global cap on total concurrently-open filters** (block + log combined), not a
  per-client one. fxevm's JSON-RPC surface has no auth/session concept today, and filters are
  explicitly connection-independent by design (https://github.com/hyperledger/fabric-x-evm/issues/331,
  decision 9) — there's no real client identity to attribute a per-client cap to without adding an auth layer,
  which is out of scope here. A global cap is the achievable granularity: `eth_newBlockFilter`/`eth_newFilter`
  return a standard JSON-RPC error once the cap is hit, existing filters are unaffected, and a freed slot
  (via uninstall or expiry) becomes available again. Configurable, with a conservative default —
  the exact number matters less than the cap existing and being tunable.
- **Subscriptions: a per-connection cap**, which *is* naturally attributable — each WS connection
  already has its own `notifier`/context, so counting `NewHeads()` calls per connection is direct,
  no new identity concept needed. A legitimate client only ever needs one `newHeads` subscription
  per connection, so the default can be small. A global cap across all connections is worth adding
  too, as defense in depth against many-connections-each-with-one-subscription, but the
  per-connection cap is the one that actually stops a single misbehaving client.
- Explicitly not attempting per-IP or per-identity attribution beyond "this WS connection" — that
  needs real auth/session infrastructure this epic doesn't have and shouldn't invent as a
  side-effect of a resource-cap fix.

**Acceptance criteria / testing:**

- Create the configured max number of filters, assert the next one errors cleanly; uninstall one,
  assert a new one can now be created (cap is on concurrent count, not lifetime total).
- Same shape for subscriptions, scoped to one connection: open the per-connection max, assert the
  next `eth_subscribe("newHeads")` on that same connection errors; a *different* connection can
  still open its own subscriptions independently.
- Existing, legitimate single-filter/single-subscription usage (i.e. everything the other issues in this epic
  already test) is completely unaffected by a reasonable default cap — this is a regression risk
  worth its own explicit test, not just an assumption.

Siddhartha-singh01:
## Summary
Stacked on #361 (Part III newHeads). This is Part IV of #329 / closes #333.

Adds configurable caps so filter and subscription creation cannot grow unbounded server memory:

1. Global concurrent filter cap (block + log combined). `eth_newBlockFilter` / `eth_newFilter` return JSON-RPC `-32005` LimitExceeded when full; uninstall or expiry frees a slot.
2. Per-WS-connection newHeads cap (default 1) plus a global newHeads cap across connections. Same `-32005` when either is hit. Connection key comes from `rpc.PeerInfo.RemoteAddr`.
3. Defaults: 1000 filters, 1 newHeads per connection, 1000 newHeads global. Tunable via `gateway.filters` in config.
4. Wiring uses `NewFilterAPIWithLimits` from app startup so the caps are live in production, not only tests.

Legitimate single-filter / single-subscription use is unchanged under the defaults.

## Test plan
- [x] `go test ./gateway/api/filters/` (cap hit, uninstall frees slot, expiry frees slot, per-conn and global newHeads caps)
- [x] `go test ./gateway/api/` (subscribe path returns LimitExceeded)
- [x] `go test ./gateway/app/`


arner:
Now that we merged https://github.com/hyperledger/fabric-x-evm/pull/361, could you rebase this branch please @Siddhartha-singh01 ? Looking forward to closing this epic!

Siddhartha-singh01:
Merged latest main (includes #361 and #381) and resolved the conflicts.

Kept the filter/subscription caps on this branch: global filter cap, per-connection + global newHeads caps, gateway.filters config, and the LimitExceeded (-32005) path. Also kept endorser validation from #381 alongside the filters validation.

Local smoke tests on gateway/api/filters, gateway/api, gateway/app, and gateway/config are green.

gateway/api/filters/api.go
Storm1289: This caps how many filters exist, not how much memory they hold. newHeads already drops when HeadsBuffer is full; log filters append with no bound. One un-polled eth_newFilter on a busy chain still grows without limit, and the new global cap lets you have 1000 of those. Can we cap (or drop) the per-filter buffer here? Otherwise the PR body should not say this bounds server memory.

gateway/api/subscribe.go
```diff
@@ -60,3 +63,14 @@ func headPayload(b *domain.Block) any {
...
+
+// subscriptionConnKey attributes a newHeads subscription to a WS connection.
+// PeerInfo.RemoteAddr is stable across eth_subscribe calls on the same socket;
+// the Notifier pointer is not (a new one is created per subscribe call).
+func subscriptionConnKey(ctx context.Context, notifier *rpc.Notifier) any {
+	info := rpc.PeerInfoFromContext(ctx)
+	if info.RemoteAddr != "" {
+		return info.Transport + "|" + info.RemoteAddr
+	}
+	return notifier
```
Storm1289: Falling back to *rpc.Notifier bypasses the per-connection cap: you already note that a new notifier is created per eth_subscribe. If PeerInfo.RemoteAddr is empty we should not pretend we have a connection key either count against the global cap only, or fail closed. RemoteAddr for a live WS socket is fine.

gateway/api/filters/limits.go
```diff
@@ -0,0 +1,45 @@
...
+// than the caps existing and being tunable via config.
+var DefaultLimits = Limits{
+	MaxFilters:              1000,
+	MaxSubscriptionsPerConn: 1,
+	MaxSubscriptionsGlobal:  1000,
+}
+
+func (l Limits) withDefaults() Limits {
+	out := DefaultLimits
+	if l.MaxFilters > 0 {
```
Storm1289: 0 meaning "use 1000" will surprise anyone who sets max-filters: 0 in yaml. Please treat omitted as default and make 0 either invalid or actually zero.

Storm1289 (commented):
Looks good overall @Siddhartha-singh01 . I think there are three issues though, left them inline.

gateway/api/filters/api.go
Siddhartha-singh01: Good catch. Added MaxFilterBuffer (default 1024): when a block/log filter is full we drop the oldest entries on append, same idea as newHeads backpressure. Covered in TestFilterBuffer_DropsOldest.

gateway/api/subscribe.go
Siddhartha-singh01: Agreed. Dropped the Notifier fallback. If RemoteAddr is empty we pass a nil connection key now, so only the global subscription cap applies.

gateway/api/filters/limits.go
Siddhartha-singh01: Fixed. gateway.filters fields are *int now: omitted (nil) keeps the default, explicit 0 means zero. The old "0 means 1000" withDefaults path is gone.

Siddhartha-singh01:
Addressed Storm's review in 5ce9284:

1. Per-filter buffer cap (default 1024), drop oldest on overflow for hashes and logs
2. No *rpc.Notifier fallback for the connection key; empty RemoteAddr → global cap only
3. Config filters fields are pointers so omitted vs explicit 0 are distinct

Local tests on gateway/api/filters, gateway/api, gateway/config, gateway/app are green.

Siddhartha-singh01:
Quick Tests was failing on a missing license header in gateway/config/filters_limits_test.go. Added it and pushed.

Siddhartha-singh01:
OZ failed with one regression: RateLimiter SlidingWindow tryConsume accumulate (expected [300,9700] got [350,9650]). That is the same hardhat-time-rpc flake class as the other SlidingWindow RateLimiter entries already in oz_known_failures.json — not related to filter caps.

Quarantined it as flaky hardhat-time-rpc and pushed. Integration was skipped only because OZ failed (workflow needs: oz-hardhat-compat).

PR #381, opened 2026-09-18
require local endorser

arner:
This completes the split between the local and remote endorsers.
The local one is mandatory now and always used for state queries.
Remotes are called over GRPC. The endorsement client brings them
together (no real change there, except that "local" got its own field
for clarity. The test uses a gateway + endorser for org1, and just an
endorser for org2.


gateway/app/app.go
```diff
@@ -320,6 +311,15 @@ func (a *App) Run(ctx context.Context) error {
 	appLogger.Debugf("starting gateway with %d workers", a.cfg.Gateway.WorkerCount)
 	a.gateway.Start(gctx)
 
+	// Serve this org's own endorser over gRPC when configured, so other orgs
+	// can reach it. cfg.Endorser itself is nil for the self-contained testnode
+	// which builds its endorser directly.
+	if a.cfg.Endorser != nil && a.cfg.Endorser.Server != nil {
+		g.Go(func() error {
+			return eserver.ServeEndorser(gctx, a.localEndorser, a.cfg.Endorser.Server)
```
ale-linux, line 319: we close this via gctx?

integration/perf/replay_json_dataset_test.go
```diff
@@ -123,7 +123,9 @@ func perfHandlerChain(completionTracker *TxCompletionTracker) integration.Handle
 		// Perf replay submits through NonceBypassGateway, which enqueues straight
 		// past the sequencer, so a real nonce gate would only add per-sender
 		// bookkeeping on every commit without ever gating a submission.
-		gw, err := app.BuildGateway(ctx, ends, gwSigner, cfg.Network, tracker, submitters, cfg.Gateway.SubmitterCount, cfg.Gateway.WorkerCount, txQueue, gwtestimpl.NewPassthroughGate(txQueue), cfg.Gateway.EndorsementChanSize, txPerSec)
+		// ends is caller-ordered [local, remotes...]; see test_helpers.go's
+		// defaultHandlerChain for the same convention.
+		gw, err := app.BuildGateway(ctx, ends[0], ends[1:], gwSigner, cfg.Network, tracker, submitters, cfg.Gateway.SubmitterCount, cfg.Gateway.WorkerCount, txQueue, gwtestimpl.NewPassthroughGate(txQueue), cfg.Gateway.EndorsementChanSize, txPerSec)
```
ale-linux, line 128: hmm, we just have 1 here no?

ale-linux (approved):
Thanks Arne! LGTM - modulo two qs which I think are ok

integration/perf/replay_json_dataset_test.go
arner: ah yes, only the local one. Oh well, this way the function keeps working if we ever change that. It's just a copy of how we do it in the other integration tests.

gateway/app/app.go
arner: Yes, the server handles it:
```
	// Handle server shutdown.
	<-gCtx.Done()
        s.Stop()
```

gctx is the main context, because also in app.go:

```
	// Shutdown trigger: fires when any goroutine fails or context is canceled
	g.Go(func() error {
		<-gctx.Done()
		return a.Shutdown()
	})
```

gctx itself is a derivative of a `signal.NotifyContext` all the way in the cmd main function, so it responds to sigterm.

Issue #379, opened 2026-09-17
gateway: `TxQueueInterface.Enqueue` has no way to report a failure


Storm1289:
`Enqueue` takes a transaction and returns nothing:

    Enqueue(tx *types.Transaction)

That is fine today, because the existing queues only append to a slice and cannot fail. It stops being fine with the dependency-manager-backed queue in #371: that implementation has to endorse inside `Enqueue` to get a read-write set before it can feed `IncomingTxs`, and endorsement can fail. With this signature there is nowhere for that failure to go.

There are two call sites, and they are not the same case:

- `nonceGate.Admit` (`gateway/core/nonce_parking.go:214`) already returns an error and sits on the synchronous `eth_sendRawTransaction` path, so a failure here could reach the client. Only the signature is in the way.
- `nonceGate.Observe` (`gateway/core/nonce_parking.go:274`) releases a parked transaction from the block handler. `Observe` returns nothing and the client already has its hash, so a failure here has no caller to report to regardless of what `Enqueue` returns.

Options:

- Add an error return to `Enqueue` and have all three implementations satisfy it. Straightforward for the `Admit` path; the `Observe` path still needs a decision about what to do with the error.
- Endorse outside the queue and pass the read-write set in, so `Enqueue` cannot fail.

Not blocking #371 - see [#371](https://github.com/hyperledger/fabric-x-evm/issues/371) for the discussion that surfaced it. Opened to track it separately.

Issue #371, opened 2026-09-11
[EPIC] Dependency-aware scheduling with the committer dependency manager


Storm1289:
## Summary

Two transactions that touch the same key cannot be endorsed at the same time. Both endorsers read the same version, both build a read-write set against it, and both get submitted. The committer's dependency graph then puts them in order at validation, so the second one's read version is out of date, it fails MVCC validation, and the client loses the transaction.

The gateway cannot see this coming, because it does not track which keys a transaction touches.

This epic brings the committer's dependency manager into the gateway so that no two clashing transactions are ever in flight at once. The work is split into the sub-issues listed below.

The current pipeline is described in the existing transaction pipeline diagrams and stays as it is - a snapshot of today. The diagrams here are the future state.

Multi-gateway scheduling is deliberately out of scope here and moves to its own story.

## The rule everything rests on

> The network never starts work on a transaction until every transaction it depends on has committed.

## A gateway and its own endorser have to stay in step

The rule only works if one more thing is true:

> If the gateway decides a transaction can go ahead, all the writes it depends on must already be applied **by that gateway's own endorser**.

The single-process test harness keeps them in step, because it registers `[endorser KVS, chain, gateway]` on one synchronizer.

## Two holds, one release signal

A transaction can be held twice on its way through, for unrelated reasons.

**The nonce gate** holds by sender ordering: a transaction runs only when it is that sender's next one. **The dependency manager** holds by key conflict, read off the read-write set, across senders.

The nonce gate sits in front of the ingest queue, the dependency manager behind endorsement, and the same committed block releases both - but not to the same place. The nonce gate releases into the ingest queue, because a parked transaction is new work. The dependency manager releases straight back to endorsement, because its transaction was admitted long ago and needs only a fresh endorsement.

Because the nonce gate already serialises a single sender, the dependency manager is left with the cross-sender case: different senders touching the same key. It does not have to reason about nonces at all.

## Why held transactions must be re-endorsed

A held transaction **has** to be re-endorsed. It was held because something else in flight touches one of its keys. When that blocker commits it bumps the version of that key, so the read-write set the held transaction is carrying describes state that no longer exists. Submitting it would fail validation for the same reason it was held.

It also cannot be endorsed and then submitted straight away. While it was being endorsed another clashing transaction may have arrived, so the fresh read-write set has to go back through the dependency manager for a new conflict check. The path is a cycle, not a straight line.

Only held transactions need this. One that passed straight through still holds a current read-write set.

[...]

Issue #349, opened 2026-08-27
Two-mempool split: separate ingest from the endorsed/dependency queue


Storm1289:
## Summary

Today a single `txqueue` both buffers incoming transactions and tracks dependencies, so endorsement latency at the front blocks accepting new work. Split it into two mempools so accepting work never waits on endorsement:

- an **ingest FIFO queue** where transactions land after admission, and
- the existing `txqueue` (now the **endorsed queue**), which holds `(transaction, endorsement, read-write set)` instead of a bare transaction and feeds dependency-aware scheduling.

New pipeline: `SendTransaction -> ingest FIFO -> workers drain, endorse (get the rwset), enqueue the (tx, rwset) pair into the endorsed queue -> re-endorse if needed -> submit`.

This is the umbrella; the work is tracked in the two sub-issues below.

## Out of scope

The real dependency manager / full conflict scheduling - a separate later chapter that needs the committer-team discussion.

## Done when

Both sub-issues land, main stays green, and the split is in place with today's re-endorse-then-submit behavior preserved.

Issue #51, opened 2026-03-26
Design Failed Transaction Recording


ale-linux:
Design and document the approach for recording and reporting failed transactions in fabric-x-evm, considering the differences between Ethereum's failure model and Fabric's MVCC validation. In Ethereum, transactions can fail due to out-of-gas, contract revert, or invalid nonce, but all failures are recorded on-chain with a failed status. In Fabric, transactions can additionally fail MVCC validation, which is not an Ethereum concept. Define how to handle MVCC failures: whether to retry automatically, how many retries to attempt, and when to report a transaction as permanently failed. Document how failed transactions are stored in the database, what status codes are returned to clients, and how transaction receipts reflect different failure types. Ensure the approach maintains Ethereum API compatibility while providing visibility into Fabric-specific failure modes for debugging. We also must decide what to do for transactions that cause the EVM execution to fail/revert: typically ethereum would still store the transaction on the ledger (as a matter of fact, the transaction already _is_ on the ledger, given that ordering has already taken place. In fabric however, we can still decide what to do with it, given that ordering happens after execution. This task tracks also the design aspect of this.

---

**Quarter:** Q2

PR #383, opened 2026-09-18
gateway: schedule transactions through the dependency manager

Storm1289:
## Summary

First two Phase 1 items of #371: a read-write set adapter, and a `TxQueueInterface` backed by the committer's dependency manager, so two transactions touching the same key are never in flight at once.

This is the prototype @ale-linux described in #371 - something to run and look at, not production code. It is off by default: nothing selects it unless a caller passes `core.NewDepGraphQueue()` as the queue, the same way the perf tests already select `TxQueueV2`.

## How it maps

`TxQueueInterface` lines up with `dependencygraph` one for one:

| `TxQueueInterface` | `dependencygraph` |
|---|---|
| `Enqueue` | endorse for a read-write set, send a `TransactionBatch` on `IncomingTxs` |
| `Dequeue` | served from `OutgoingDepFreeTxsNode` |
| `Handle` | send the block's transactions on `IncomingValidatedTxsNode` |

Workers downstream are unchanged: they dequeue, endorse and submit.

## The adapter is smaller than the sub-issue assumed

The sub-issue describes converting `blocks.ReadWriteSet` into `[]*applicationpb.TxNamespace`. That conversion is not needed. `TxWithRef.Content` is an `applicationpb.Tx`, and a fabric-x `ProposalResponse.Payload` already is one - the endorsement builder puts it there, and since #306 it survives the gRPC hop. So `txContent` unmarshals and validates rather than converting.

## Two bugs this prototype found, both mine

**Concurrent endorse workers deadlocked the manager.** Batch IDs must reach the local constructor in order - it spins on `CompareAndSwap(id-1, id)` and reads with one worker. With 8 endorse workers each numbering and sending independently, batch 6 could arrive before batch 5, and the constructor would wait forever for a predecessor stuck behind it in the same channel. Numbering and sending are now one atomic step. Reproduced reliably: 1 worker passed instantly, 8 wedged.

**A nil `PrometheusMetricsProvider` panics.** It is not optional - `newPerformanceMetrics` calls straight through and segfaults on the nil receiver.

## Behaviour worth knowing

`Enqueue` cannot block: the nonce gate calls it while holding its own write lock, so endorsement happens on a worker pool rather than inline.

The queue does not preserve FIFO across senders - batch order follows endorsement completion, not `Enqueue`. The nonce gate already serialises a single sender, so only cross-sender transactions arrive together and their order is free.

`Bind` rejects classic Fabric at wiring time. The manager schedules on `applicationpb` namespaces, which a classic endorsement does not carry, so every transaction would otherwise be dropped one at a time at runtime.

`conflictEnq` now reports contention, by comparing the committed-block count at enqueue against release. Approximate, and the only signal available: the package clears a node's dependencies before handing it over.

## Question for review: the dependency tail

Importing `dependencygraph` pulls the Docker client into the gateway binary:

```
dependencygraph -> utils/monitoring -> utils/test -> fsouza/go-dockerclient
```

The cause is `utils/monitoring/test_exports.go` in fabric-x-committer. It is a plain `.go` file rather than `_test.go`, so it compiles into the production package and brings `utils/test` - which imports `testing`, testify and the Docker client - with it.

Measured on `cmd/fxevm`:

| | size | docker symbols |
|---|---|---|
| main | 52.7 MB | 0 |
| this branch | 58.3 MB | 10 |

`go.mod` gains 15 indirect modules (`moby/moby/client`, `opencontainers/image-spec`, `containerd/log`, `sirupsen/logrus` and others).

Renaming that file to `test_exports_test.go` upstream would remove the whole tail. Happy to carry it for the prototype if you would rather not block on that - wanted to raise it rather than land it quietly. @ale-linux @arner, what would you prefer?

## Out of scope

Mempool split (#349/#350/#351), `Enqueue` error handling (#379), multi-gateway, retry (#50), failed-transaction reporting (#51), release gated on the endorser's applied height (#373), lagging-endorser wait, metrics and `WaitingTxsLimit` tuning, the loopback test harness.

## Follow-up before this is more than a prototype

Comments in the new files still point at phases and issue numbers from #371's plan (`#379`, "Phase 2 metrics item"). Those are accurate today but will go stale as the plan moves - they need a pass once the shape settles.

## Tests

30 tests: 6 for the adapter, 24 for the queue.

- `go test ./gateway/core -race -count=3`
- `go test ./... -short`
- `go vet ./...`, `gofmt`
- Full `TestFabricX` case matrix run through `DepGraphQueue` against a live fabric-x committer on a fresh network: 10/10, alongside the default queue's 13/13

Covered: clash-free pass-through, clashing held until commit, chains of contention released one at a time, rejected blockers still releasing dependents, `Complete` releasing on submission failure, conflict counting, drop accounting, concurrent producers and consumers, idempotent `Close` and `Complete`, protocol rejection, and a classic Fabric payload rejected rather than read as an empty read-write set.


gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,453 @@
...
+		depGraphLogger.Errorf("read-write set for tx %s: %v", hash.Hex(), err)
+		q.drop(hash)
+		return
+	}
+
+	// IDs must start at 1 and rise by one: the constructor spins on
+	// CompareAndSwap(id-1, id) against a zero-valued atomic, so ID 0 never
+	// lands. Numbering and sending are one step, or endorse workers racing here
+	// deliver them out of order and the constructor deadlocks.
+	q.sendMu.Lock()
```
ale-linux: iiuc this mutex is only protecting `batchID` and `incoming`, both of which should already be safely usable from multiple goroutines - do we really need it? what am I missing?

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,453 @@
...
+	batch := &dependencygraph.TransactionBatch{
+		ID: q.batchID.Add(1),
+		Txs: []*servicepb.TxWithRef{{
+			Ref:     &committerpb.TxRef{TxId: hash.Hex()},
+			Content: content,
+		}},
+	}
```
ale-linux: we should check with Senthil what the guidelines are here... should we try to batch them with some timeout? Or is it safe to send them in as we get them?

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,453 @@
...
+
+// submitToManager endorses for a read-write set and sends a one-transaction
+// batch. A failure here drops the transaction: Enqueue has no error return, so
+// there is nowhere to report it (#379).
+func (q *DepGraphQueue) submitToManager(tx *types.Transaction) {
+	hash := tx.Hash()
+
+	q.mu.RLock()
+	endorser := q.endorser
+	q.mu.RUnlock()
```
ale-linux: What is happening here?

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,453 @@
...
+	if endorser == nil {
+		depGraphLogger.Errorf("Bind was never called, dropping tx %s", hash.Hex())
+		q.drop(hash)
+		return
+	}
```
ale-linux: I don't think we want to cater for this case - it would be a programming error and we couldn't survive it anyway

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,453 @@
...
+	if q.done {
+		q.mu.Unlock()
+		return
+	}
+	if _, dup := q.tracked[hash]; dup {
+		q.mu.Unlock()
+		return
+	}
+	q.tracked[hash] = &trackedTx{tx: tx, enqueuedAtBlock: q.blocksHandled.Load()}
+	q.byTxID[hash.Hex()] = hash
```
ale-linux: what are we doing here? 

- the name appears wrong - the txid isn't used anywhere
- there are ways of converting string hash to binary and vice versa w/o using a map

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,453 @@
...
+	q.mu.RLock()
+	endorser := q.endorser
+	q.mu.RUnlock()
+	if endorser == nil {
+		depGraphLogger.Errorf("Bind was never called, dropping tx %s", hash.Hex())
+		q.drop(hash)
+		return
+	}
+
+	end, err := endorser.ExecuteTransaction(q.ctx, tx)
```
ale-linux: with reverts and all we must be careful here - @arner since you last touched this code could you pls weigh in?

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,453 @@
...
+	if err != nil {
+		depGraphLogger.Errorf("endorse tx %s: %v", hash.Hex(), err)
+		q.drop(hash)
+		return
+	}
+	content, err := txContent(end)
+	if err != nil {
```
ale-linux, line 249: I think we want to panic here for now so that these irrecoverable errors surface. We also need to create an issue in the code hygiene epic to remove it in production

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,453 @@
...
+		}
+	}
+}
+
+func (q *DepGraphQueue) markReady(nodes dependencygraph.TxNodeBatch) {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	for _, node := range nodes {
+		hash, known := q.byTxID[node.VerifierTx.GetRef().GetTxId()]
```
ale-linux: how can this work? We add to this map

```go
	q.byTxID[hash.Hex()] = hash
```

so that's an ethereum tx hash, but here we try to get it out by txid

ale-linux (commented):
Thanks @Storm1289 - that's precisely the way forward! 👍 

Now two things:

1) have a look at my comments and maybe fix what is worth fixing; but, more importantly...
2) we need to actually plug it in some test to see what breaks. I would recommend plugging it into the `TestReplayJSONDataset` inside integration/perf/replay_json_dataset_test.go to see an end to end test. You can refer to .github/workflows/perf-smoke-comment.yml to see how to drive that test

Storm1289:
Thanks @ale-linux, Pushed a6a40be and f95ad63.

- `byTxID` gone. It did work - `TxId` is set to `hash.Hex()` in `submitToManager` - but you were right that it read as broken, and the name was wrong. `markReady` now derives the hash with `common.HexToHash`.
- Endorse workers start in `Bind` rather than the constructor, so the endorser is written before anything that reads it exists. That removes both the read lock and the nil check.
- Unreadable read-write set now panics, as does `Enqueue` before `Bind` - with the workers starting in `Bind`, an unbound queue would otherwise accumulate in silence. Both prototype-only, tracked in #386, which notes it is tangled with #379 since `Enqueue` has no error return to use instead.
- Expanded the comment on the mutex, and added tests for the two new panics.

On the mutex: each is safe alone, but the manager needs batches to *arrive* in order and neither gives that. A worker can be descheduled between taking an ID and sending, so batch 6 reaches the constructor first and it spins on `CompareAndSwap(5, 6)` while batch 5 sits behind it in the same channel. 8 endorse workers wedged on the concurrency test, 1 passed instantly.

On batching: agreed, one for Senthil. One transaction per batch today. Worth asking him about `NumOfLocalDepConstructors` at the same time, since single-transaction batches leave it nothing to parallelise.

On reverts: a revert does not reach that path - `StatusEVMRevert` is in the endorsable set in `endorse.go:97`, so `ExecuteTransaction` returns no error for it. What errors there is a rejected transaction, nonce too low and the like, which I currently drop. @arner over to you.

**Plugged into `TestReplayJSONDataset`** behind a `-depgraph` flag, alongside `-oldqueue`. 10k USDC transfers, 500 outstanding, fresh network for each run:

| queue | throughput | failed | conflict rate |
|---|---|---|---|
| `TxQueue` | 156.70 tx/s | 0 | 0.000 |
| `TxQueueV2` | 152.25 tx/s | 0 | 0.000 |
| `DepGraphQueue` | 155.51 tx/s | 0 | **0.047** |

10000/10000 committed, nothing dropped, throughput within noise of v1. The conflict rate is the interesting part: 470 transactions were held and released, which neither of the others sees - v1 does not look, and v2's participant heuristic never fired on this workload. Same throughput while actually serialising those 470 suggests the holds were short rather than over-serialising.

gateway/core/depgraph_queue.go
ale-linux: > On the mutex: each is safe alone, but the manager needs batches to arrive in order and neither gives that. A worker can be descheduled between taking an ID and sending, so batch 6 reaches the constructor first and it spins on CompareAndSwap(5, 6) while batch 5 sits behind it in the same channel. 8 endorse workers wedged on the concurrency test, 1 passed instantly.

but then the right form here is to have all endorsers enqueue into a buffered channel, and have another goroutine drain that channel and package batches - something like

```go
func consume[T any](
	ch <-chan T,
	timeout time.Duration,
	threshold int,
	submit func([]T),
) {
	ticker := time.NewTicker(timeout)
	defer ticker.Stop()

	var cache []T

	for {
		select {
		case item := <-ch:
			cache = append(cache, item)

			if len(cache) >= threshold {
				submit(cache)
				cache = nil
			}

		case <-ticker.C:
			// Drain everything currently available.
			for {
				select {
				case item := <-ch:
					cache = append(cache, item)
				default:
					// Nothing left to drain.
					submit(cache)
					cache = nil
					goto next
				}
			}

		next:
		}
	}
}
```

so basically we go:

1. gw calls `Enqueue`
2. `Enqueue` sends into `admitted`
3. endorsement workers drain `admitted`, call `submitToManager`
4. `submitToManager` gets rwset, sends into a new channel
5. ONE single goroutine drains this channel with a logic similar to the one I pasted above. This one doesn't need a mutex, not even an atomic integer: it is the one and only that packages a batch and sends it into `incoming`.

Wdyt?

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,461 @@
...
+// namespaces to schedule on, so every transaction would be dropped one at a
+// time at runtime instead of failing once at startup.
+func (q *DepGraphQueue) Bind(e txEndorser, protocol string) error {
+	resolved, err := cmn.NormalizeProtocol(protocol)
+	if err != nil {
+		return err
+	}
+	if resolved != cmn.ProtocolFabricX {
+		return fmt.Errorf("dependency manager queue requires %q, got %q", cmn.ProtocolFabricX, resolved)
+	}
```
ale-linux: let's get rid of the protocol check - do we really need it?

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,461 @@
...
+
+	q.mu.Lock()
+	if q.bound {
+		q.mu.Unlock()
+		return fmt.Errorf("dependency manager queue is already bound")
+	}
```
ale-linux: I think we can get rid of this check - we `Bind` only once and at the right time anyway

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,461 @@
...
+	q.endorser = e
+	q.bound = true
```
ale-linux: we can get rid of this and assume we're always bound

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,461 @@
...
+
+// Enqueue accepts a transaction for scheduling. It must not block: the nonce
+// gate calls it while holding its own write lock, so the endorsement happens on
+// an endorse worker rather than here.
+func (q *DepGraphQueue) Enqueue(tx *types.Transaction) {
+	hash := tx.Hash()
+
+	q.mu.Lock()
+	if !q.bound {
+		q.mu.Unlock()
```
ale-linux: can we please `defer` the unlock?

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,461 @@
...
+	if _, dup := q.tracked[hash]; dup {
+		q.mu.Unlock()
+		return
+	}
+	q.tracked[hash] = &trackedTx{tx: tx, enqueuedAtBlock: q.blocksHandled.Load()}
+	q.totalEnq++
+	q.mu.Unlock()
+
+	select {
+	case q.admitted <- tx:
```
ale-linux, line 205: shouldn't we also keep the lock while enqueueing?

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,461 @@
...
+	default:
+		// Nothing downstream can report this, and blocking here would stall the
+		// nonce gate for every sender. Dropping is the behaviour #379 covers.
+		depGraphLogger.Warnf("admitted channel full, dropping tx %s", hash.Hex())
+		q.drop(hash)
```
ale-linux: let's also panic so that these errors surface for now

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,461 @@
...
+}
+
+func (q *DepGraphQueue) markReady(nodes dependencygraph.TxNodeBatch) {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	for _, node := range nodes {
+		// TxId is the transaction hash we set in submitToManager.
+		hash := common.HexToHash(node.VerifierTx.GetRef().GetTxId())
+		t, tracked := q.tracked[hash]
```
ale-linux, line 330: this keeps being wrong: on one side we add the ethereum hash of the transaction and in the other (here) we set the fabric(-x) txid

gateway/core/depgraph_queue.go
```diff
@@ -0,0 +1,461 @@
...
+		if !tracked {
+			continue
+		}
+		t.node = node
+		// Approximate, and the only signal available: a transaction released in
+		// the same block window it arrived in cannot have waited on anything.
+		if q.blocksHandled.Load() > t.enqueuedAtBlock {
+			q.conflictEnq++
+		}
```
ale-linux: I don't think that's the right way.. can you maybe check with Senthil?

ale-linux (changes requested):
Excellent progress! Some more comments

Storm1289:
Thanks @ale-linux

We do not need to ask Senthil anything further about the dependency manager. He has already shared the dependency graph documentation on Slack and it covers everything we needed - I have been through it, and the package source confirms it. Worth a read if you want the detail, but nothing here is blocked on his input.

**Contention metric.** You were right that the conflict figure was wrong. Once batching landed, the block-window approximation was timing how long a transaction sat in the batcher, not whether the graph held it. That is removed. The package already exports the real signal, `coordinator_dependency_graph_dependent_transactions_queue_size` (`metrics.go:157`), so we now keep the `monitoring.Provider` we were already passing and sample it once per block from `Handle`, reporting the peak via `PeakDependentTxs()`. It is a separate accessor rather than the fourth value of `Stats()`, because that slot gets divided by `totalEnq` to print a rate and a peak is not a rate. `TxQueue` and `TxQueueV2` are untouched, `go.mod` unchanged.

**Batch ordering.** The contract is in `localDependencyConstructor.construct` (`local_dependency_constructor.go:130`): IDs must be gapless from 1, and arrival may be out of order by at most N-1 batches where N is `NumOfLocalDepConstructors`. We run one constructor, so any reordering deadlocks. The single-goroutine batcher in `60370ce` is what satisfies that.

**Batch sizing.** The threshold of 64 and the 10ms timeout are still empirical, left for Phase 2 tuning.

**OZ Hardhat** looks like a flake.

gateway/core/depgraph_queue.go
Storm1289: Done, the check is gone

gateway/core/depgraph_queue.go
Storm1289: Done in `60370ce`

gateway/core/depgraph_queue.go
Storm1289: Done

gateway/core/depgraph_queue.go
Storm1289: Done, it is a single `defer q.mu.Unlock()` now, no manual unlocks on the early returns.

gateway/core/depgraph_queue.go
Storm1289: Done, the lock is now held across the send into `admitted` and the `tracked` write.

gateway/core/depgraph_queue.go
Storm1289: Done

gateway/core/depgraph_queue.go
Storm1289: You were right that something was off here, though not quite the part you flagged.

The real bug was the reverse conversion. `common.HexToHash` never fails: given an empty or malformed string it pads or truncates and still returns a hash. So an unexpected id would silently become the zero hash, miss in `tracked`, and the transaction would be dropped without a trace.

`hashFromTxRefID` now returns `(common.Hash, bool)` and validates length, prefix and hex decoding, and `markReady` panics on anything it does not recognise, consistent with the other prototype panics in #386. The round-trip test now also checks a set of malformed ids that previously all produced a valid-looking hash.

PR #384, opened 2026-09-21
feat: use the real nonce gate in testnode and reset it on evm_revert

arner:
The test RPC replaced the nonce gate with a passthrough, because evm_revert rewinds ledger nonces and would leave a cached next-nonce stale.

Clients that pipeline sequential nonces from one sender then had every future-nonce transaction fail with "nonce too high" in the worker and get dropped.

Add a wrapper around the NonceGate that replaces the instance with a fresh one whenever a revert is called. The passthrough gate stays for the integration and perf harnesses, which replay signed transactions.


gateway/testimpl/resettable_gate.go
```diff
@@ -0,0 +1,61 @@
...
+// NewResettableGate returns a gate that must be Bound before its first transaction.
+func NewResettableGate() *ResettableGate { return &ResettableGate{} }
+
+// Bind sets how to build the real gate and installs one. It runs once the gateway exists, since the gate reads nonces from it.
+func (g *ResettableGate) Bind(fresh func() core.NonceSequencer) {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	g.fresh = fresh
+	g.inner = fresh()
+}
```
ale-linux, line 36: I know agents love to do this - but do we really need this? The binding is quite static right?

gateway/testimpl/resettable_gate.go
```diff
@@ -0,0 +1,61 @@
...
+	g.mu.RLock()
+	defer g.mu.RUnlock()
+	g.inner.Observe(committed)
+}
+
+func (g *ResettableGate) IsPending(hash common.Hash) *types.Transaction {
+	g.mu.RLock()
+	defer g.mu.RUnlock()
+	return g.inner.IsPending(hash)
+}
```
ale-linux, line 61: do we really need serialisation? I'm guessing those tests will synchronously call a revert and then restart no?

ale-linux (commented):
Love this approach! Very clean. A couple of dumb Qs..

gateway/testimpl/resettable_gate.go
arner: Yeah I had the same question. The reason is because the factory needs the gateway. And the gateway needs the noncegate.... So we have to first build the gateway with the resettable noncegate, and only then bind it to the constructor function. I couldn't find a cleaner way.

gateway/testimpl/resettable_gate.go
arner: Observe gets blocks from the network asynchronously, so technically we could call the Reset at the same time as when a new block is still coming in and reads the pointer. For the IsPending, in practice we wait for transactions to be committed before returning. But IsPending can still be called in parallel, Gateway.TransactionByHash is called via awaitCommit and that one does `g.nonceGate.IsPending(hash)`. There could be a race there, or a check for a future nonce. So to be sure I'd prefer to have a mutex here.

ale-linux (approved):

PR #385, opened 2026-09-21
feat: add eth_getBlockReceipts and null contractAddress for non-creation receipts

arner:
- Serve eth_getBlockReceipts for a block number or hash: null for an unknown block, [] for an empty one, otherwise the same receipts eth_getTransactionReceipt returns.
- Return contractAddress: null unless the transaction created a contract.
- Return the real blockTimestamp on the logs of eth_getTransactionReceipt.


gateway/api/eth.go
```diff
@@ -347,6 +348,56 @@ func (api *EthAPI) GetTransactionReceipt(ctx context.Context, hash common.Hash)
...
+		return nil, nil
+	}
+
+	receipts := make([]*rpcReceipt, 0, len(b.Transactions))
+	if len(b.Transactions) == 0 {
+		return receipts, nil
+	}
+
+	// The block query loads the transactions but not their logs; fetch those for the whole block at once.
+	logs, err := api.b.GetLogs(ctx, domain.LogFilter{FromBlock: &b.BlockNumber, ToBlock: &b.BlockNumber})
```
ale-linux, line 377: it's a bit annoying though to have to query the DB 3x no? Any way we can have a single query to do it all?

arner: Not really. It can't be one, because we can't get a Block struct with a field with []Logs in one go. Or to merge Logs and Transactions. I mean, technically possible but would require a whole new mapping with nullable fields and quite some code to re-map it to a block with transactions and logs.

To go to two, perhaps we could not get the block, but only rely on the block_hash and block_num from the transaction. But we need to know when we get zero transactions whether a block is null or just empty, so that requires logic (and the third call if it's necessary).

Tbh I don't find it too annoying. It's microseconds on an embedded SQLite instance. Maybe at some point we identify this as a bottleneck and optimize, but I would prefer the simpler approach of 3 calls here.

ale-linux (approved):

